PR #1, opened 2021-09-01
build image with `ko`, add e2e tests

### LINKED ISSUE #13: setup automated e2e acceptance
given that we have not just access to a container when running these actions, but instead, VMs, we might as well run a local cluster (guess what, we already have scripts for that!) and then do a full run of the example pointing at the local registry.

it might or might not be an overkill to run this on every commit to every PR - I believe we could automate that notion of "heavy duty tests only for certain PRs" by filtering the events via labels (like how some projects have "ok-for-tests" etc) we could have a label "ok-for-acceptance", which would be nice as it'd not block those small PRs that have not much to do with the whole thing.

proposed steps:

```
0. capture a common ip that anyone can reach:

        - host (so `kbld` can push the controller image to it)
        - kubelet within `kind-control-plane` container (so it can pull the
          image to run the controller)
        - kpack controller within a container inside `kind-control-plane` (so
          it can push the images it builds: builder and actual app image)


1. bring the registry up

        - plain `registry:2` using `docker run` but publishing to the host the
          ports we care about (5000)


2. bring kubernetes up

        - having that insecure registry trusted ($ip:5000)


3. generate a local release pointing at that registry

        KO_DOCKER_REPO=$ip:5000 make release


4. install all the dependencies

        ./hack/install-examples-dependencies.sh


5. run the examples

        ytt --ignore-unknown-comments -f ./examples \
          --data-value registry.server=$ip:5000
          --data-value registry.username="" \
          --data-value registry.password="" \
          --data-value image_prefix="$ip:5000/foo-" | \
          kapp deploy --yes -a example -f-


6. wait until we have all good

        is_ksvc_ready () {
          kubectl wait services.serving dev --for condition=ready
        }
```

cirocosta:
todo:

- [x] ~overlay to remove knative's default resource reservation~


cirocosta:
alright, now the final allocated cpu (+ k8s-system ones) should be really low

```
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource           Requests    Limits
  --------           --------    ------
  cpu                950m (5%)   100m (0%)
  memory             290Mi (1%)  390Mi (1%)
  ephemeral-storage  100Mi (0%)  0 (0%)
  hugepages-1Gi      0 (0%)      0 (0%)
  hugepages-2Mi      0 (0%)      0 (0%)
```

cirocosta:
the e2e is currently failing, and I think is a regression, actually 🤔 @idoru @waciumawanjohi, I _think_ it can be related to the caching work, but I'm not 100% sure

despite `image` having produced the status we want

```
Status:
  Build Cache Name:  dev-cache
  Build Counter:     1
  Conditions:
    Last Transition Time:         2021-09-01T18:30:15Z
    Status:                       True
    Type:                         Ready
    Last Transition Time:         2021-09-01T18:30:15Z
    Status:                       True
    Type:                         BuilderReady
  Latest Build Image Generation:  1
  Latest Build Reason:            CONFIG
  Latest Build Ref:               dev-build-1-wq85k
  Latest Image:                   10.88.111.6:5000/example-dev@sha256:1627ec064ce8463177b61d6999329250766a84efe4642d37bbd848702e4f4a99
  Latest Stack:                   io.buildpacks.stacks.bionic
  Observed Generation:            1
```

we see that workload doesn't make progress

```
Status:
  Conditions:
    Last Transition Time:  2021-09-01T18:29:45Z
    Message:
    Reason:                Ready
    Status:                True
    Type:                  SupplyChainReady
    Last Transition Time:  2021-09-01T18:29:50Z
    Message:               Component 'image-builder' is waiting to read value '.status.latestImage'
    Reason:                MissingValueAtPath
    Status:                Unknown
    Type:                  ComponentsSubmitted
    Last Transition Time:  2021-09-01T18:29:45Z
    Message:
    Reason:                ConditionInUnknownState
    Status:                Unknown
    Type:                  Ready
  Observed Generation:     1
  Supply Chain Ref:
    Kind:  ClusterSupplyChain
    Name:  supply-chain
```

although I see from the logs that the reconciler is reconciling every 5s (i.e., it's not stuck):

```
{"level":"info","ts":1630522063.6280022,"logger":"controller.supply-chain","msg":"started","name":"supply-chain","namespace":"","name":"supply-chain","namespace":""}
{"level":"info","ts":1630522063.6281989,"logger":"controller.supply-chain","msg":"finished","name":"supply-chain","namespace":"","name":"supply-chain","namespace":""}
{"level":"info","ts":1630522065.65732,"logger":"controller.workload","msg":"started","name":"dev","namespace":"default","name":"dev","namespace":"default"}
{"level":"info","ts":1630522065.6742113,"logger":"controller.workload","msg":"finished","name":"dev","namespace":"default","name":"dev","namespace":"default"}
```

if I then proceed with restarting the controller by getting rid of the pod and letting the replicaset controller do its thing, I can see progress:

```
k get apps
NAME   DESCRIPTION           SINCE-DEPLOY   AGE
dev    Reconcile succeeded   30s            30s
```

any thoughts?

waciumawanjohi:
I see that this PR creates an `e2e.sh` script in the `hack` directory. One of the responsibilities of this is to bring up a kind cluster with a registry. We already have a script to do this: `local-dev/create-local-cluster-and-registry.sh`. It leverages the open source project [ctlptl](https://github.com/tilt-dev/ctlptl) and is very declarative.

waciumawanjohi:
Pain point: Ran e2e script. It failed because I did not have `ko` installed. After installing `ko` I reran the script. Script failed because I had to manually stop and remove docker containers `cartographer-registry` and `cartographer-control-plane`.

cirocosta:
Hey @waciumawanjohi , thanks for checking out!

> **waciumawanjohi**: _[...] We already have a script to do this: local-dev/create-local-cluster-and-registry.sh. It leverages the open source project ctlptl and is very declarative._

I had used `ctlptl` for it in a first iteration, but I think this is a better approach as it doesn't need to do any "localhost:*" mirror setup or  rely on names that would only resolve within the network that the containers (registry and kind) are in - with the approach here (where we use the same ip address of the host for everything) we get all of our tools tagging the images with the same `$ADDR:$PORT` name so one can push/pull those images however they want wherever they are 

when trying `ctlptl` for the e2e, I wrote the following in an issue in the previous repository we had:

> I think we have a neat setup already for local dev, but while that works with
> using plain "localhost:65432" in the setup we have for building and running the
> image from the pod (yes, `localhost:65432` is reachable from our host and
> `localhost:65432` is """reachable"""" - there's just a mirror setup for
> containerd), it's not from `kpack`'s pod perspective (as it wouldn't resolve
> 'localhost' properly, so we'd need to have the image pointing at something like
> the machines `eth0` or something like that
> 
> 
> ```console
> $ cat /etc/containerd/config.toml
> 
> ... (some unrelated stuff stripped out by me)
> 
>     [plugins."io.containerd.grpc.v1.cri".registry]
>       [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
>         [plugins."io.containerd.grpc.v1.cri".registry.mirrors."kontinue-registry:5000"]
>           endpoint = ["http://kontinue-registry:5000"]
>         [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:65432"]
>           endpoint = ["http://kontinue-registry:5000"]
> ```
> 
> i.e., we'd need to make it so:
> 
> 1. `ko`, from our host network ns can reach it
> 2. `kubelet`'s container runtime can reach it (i.e., from `kind-control-plane`
>    container network ns)
> 3. `kpack`, from the container inside the `kind-control-plane` container  can
>    reach it
> 
> ```
> HOST-NS
>         ko
> 
>         KIND-CONTROL-PLANE-NS
>                 containerd
> 
>                 KPACK-NS
>                         kpack
> 
>         REGISTRY-NS
>                 registry
> ```
> 
> given that they all can reach what's under `host-ns`, providing eth0's as the
> ip of the registry would do the trick without having to do any name resolution
> trickery.
> 
> to get that IP out there without hardcoding anywhere, we could make use of
> `ytt` 🤔  
> 

(+, tbh, I don't see much value in using `ctlptl` just for bringing a registry - as showcased here, we can achieve the same without relying on an extra tool for it)

cirocosta:
> **waciumawanjohi** _Pain point: Ran e2e script. It failed because I did not have ko installed. After installing ko I reran the script._

I'd argue that's fine - being under `hack/ci`, it assumes that it's meant to be run in CI where the tools are already installed and that it would be run once -  no idempotency properties at all.

I think it's a useful property of it that it works regardless of whether it is in a CI environment or not, in which case you can tweak that `main` function as much as you want (just like one could tweak a line in a declarative `.yaml`, one could very well comment which functions `main` is running).

I started adding a check and splitting the `run` into two functions

```diff
diff --git a/hack/ci/e2e.sh b/hack/ci/e2e.sh
index 1d33c71..da064fb 100755
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.

-
 set -o errexit
 set -o nounset
 set -o pipefail
@@ -29,24 +28,28 @@ readonly KUBERNETES_CONTAINER_NAME=cartographer-control-plane
 readonly COMMAND=${1:-run}

 main() {
+        assert_tools_installed
+
         case $COMMAND in
-        run)
+        setup)
                 display_vars
                 start_registry
                 start_local_cluster

                 generate_cartographer_release
                 install_dependencies
-
+                ;;
+        test)
                 setup_example
                 test_example
                 ;;
+
         teardown)
                 delete_containers
                 ;;
         *)
                 echo "error: unknown command '$COMMAND'."
-                echo "usage: $0 (run|teardown)."
+                echo "usage: $0 (setup|run|teardown)."
                 exit 1
                 ;;
         esac
@@ -64,6 +67,24 @@ display_vars() {
        "
 }

+assert_tools_installed() {
+        assert_installed "ko"
+        assert_installed "ytt"
+        assert_installed "kapp"
+        assert_installed "docker"
+        assert_installed "kubectl"
+}
+
+assert_installed() {
+        local cmd=$1
+
+        if [[ -z $(command -v $cmd) ]]; then
+                echo "ERROR: program '$cmd' not found in $PATH."
+                echo "Aborting."
+                exit 1
+        fi
+}
+
 setup_example() {
         ytt --ignore-unknown-comments \
                 -f $DIR/../../examples/source-to-knative-service \
```

but, tbh, I don't think we should have those in place.

wdyt?

thx!

cirocosta:
(force-push due to rebase)

cirocosta:
e2e tests passed, but the previous way of generating a release left .. leftovers

Screen Shot 2021-09-03 at 9 05 40 AM

should be good now 🤞

squeedee:
Todd and I wrote a tool that creates portable, easy to edit and reason about, Jobs/Tasks as docker images. It may help with portability in cases like this. Or at the very least, it might be worth considering how to make tasks like these into docker images.

squeedee:
@cirocosta With docker logged in to gcr and to docker hub, and `ko` logged into docker hub, i am seeing:

```
2021/09/03 11:23:18 Using base gcr.io/distroless/static:nonroot for github.com/vmware-tanzu/cartographer/cmd
2021/09/03 11:23:20 Building github.com/vmware-tanzu/cartographer/cmd for linux/amd64
2021/09/03 11:23:23 Publishing projectcartographer/cartographer-controller/cmd-72b2fdd4563b5da9ab700e504fe28dcd:latest
Error: error processing import paths in "-": error resolving image references: HEAD https://index.docker.io/v2/projectcartographer/cartographer-controller/cmd-72b2fdd4563b5da9ab700e504fe28dcd/blobs/sha256:b49b96595fd4bd6de7cb7253fe5e89d242d0eb4f993b2b8280c0581c3a62ddc2: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details)
```

I simply cannot understand what other authorization we need to provide?

cirocosta:
@squeedee 🤔  that's odd - there should not be the need of _any_ auth whatsoever as it uses a local registry with no auth.

e.g., looking at `.github/workflows/validation.yaml` you'll see that there are no credentials anywhere (just a `./hack/ci/e2e.sh` call)

are you setting env vars / modifying the script? e.g., 

> Publishing projectcartographer/cartographer-controller/cmd-72b2fd

makes me think `REGISTRY` has been set?

cirocosta:
Oh, I guess you're referring to `make release`? looking at the error message


```
						KO_DOCKER_REPO	                      automatically generated by ko
                                      .------------------------------------------. .---------------------------------.
[...] HEAD https://index.docker.io/v2/projectcartographer/cartographer-controller/cmd-72b2fdd4563b5da9ab700e504fe28dcd/blobs/sha256:b49b96595fd4bd6de7cb7253fe5e89d242d0eb4f993b2b8280c0581c3a62ddc2[...]
```

 closely it seems like `make release` has been run with `KO_DOCKER_REPO=projectcartographer/cartographer-controller`, but it should actually be just `KO_DOCKER_REPO=projectcartographer` instead.

---

(force-pushed to rebase)

waciumawanjohi:
> but, tbh, I don't think we should have those in place.
> 
> wdyt?

Teardown script allays concerns about being able to rerun tests. There may be call in the future to have a script that automatically calls teardown after run.

waciumawanjohi:
> i.e., we'd need to make it so:
> 
> ko, from our host network ns can reach it
> kubelet's container runtime can reach it (i.e., from kind-control-plane
> container network ns)
> kpack, from the container inside the kind-control-plane container can
> reach it

Technical challenges in getting `ctlptl` to integrate with the tools in question are good reasons to maintain our own script instead.

> (+, tbh, I don't see much value in using ctlptl just for bringing a registry - as showcased here, we can achieve the same without relying on an extra tool for it)

I do see value in reducing the number of bash scripts we need to maintain. Offloading that complexity to an external dependency is good.

To reiterate, I think your points about the technical difficulty of using ctlptl with our tools is the overriding concern at the moment.

waciumawanjohi (approved):

Issue #6, opened 2021-08-25
Introduce Pipeline


cirocosta:
```
As a supply chain author
I want to stamp immutable objects for changing inputs 
So that I can run immutable resources based on constrained resources (EG SCM commits)
```

* Pipeline CRD
* Template PipelineRun CRD
* When Pipeline changes - stamp new.

## Notes
Assume there are no completion rules and wait for condition ready True
Use a new PipelineController.

## Todo
- [x] Create Pipeline with params that can be used in RunTemplate
- [x] Create Runtemplate that lets us specify a template to stamp and selects for the Pipeline
- [x] Create Controller that stamps for pipelines
- [x] Ensure unit tests exist for static schema validation
- [x] Q: Do we need validating webhooks? not yet
- [x] Ensure each stamp creates a new object
- [x] change runTemplateName to runTemplateRef (to include namespacing)?
- [x] New story Q: Deal with templated naming versus `generateName`? reject templated naming, make new issue
- [x] ~react to template changes~ see issue #5
- [x] used 'Stub' instead of 'Returns' on fakes... clean that up
- [x] New story for Params
- [x] New story for adding pipeline to supply chain e2e
- [x] New story for changing the template causes a new reconcile/stamp - func mapping watcher thing
- [x] New story Patch less on status update

PR #16, opened 2021-09-01
Decouple stamper from workload

squeedee:
Makes the Stamper usable by Pipeline and eventually Delivery.

[#6]

Co-authored-by: Ciro da Silva da Costa <ciroscosta@vmware.com>


pkg/utils/matchers/meaningful_error.go
```diff
@@ -0,0 +1,48 @@
+package matchers
+
+import (
+	"strings"
+
+	"github.com/onsi/gomega/types"
+
+	"fmt"
```
cirocosta: group the core ones together

```

import (
	"fmt"
	"strings"

	"github.com/onsi/gomega/types"
)
```

pkg/apis/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
cirocosta: `golangci-lint` running with `fix: true` will end up auto-adding this back up when auto formatting

> 
> `//go:build lines`
> The go command now understands //go:build lines and prefers them over // +build lines. The new syntax uses boolean expressions, just like Go, and should be less error-prone. As of this release, the new syntax is fully supported, and all Go files should be updated to have both forms with the same meaning. To aid in migration, gofmt now automatically synchronizes the two forms. For more details on the syntax and migration plan, see https://golang.org/design/draft-gobuild.

https://golang.org/doc/go1.17

pkg/utils/matchers/meaningful_error.go
```diff
@@ -0,0 +1,47 @@
...
+	} else {
+		return fmt.Sprintf("Expected\n\t%#v\nto contain \n\t%#v", actual.(error).Error(), matcher.expected)
+	}
+}
+
+func (matcher *beMeaningfulMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	if actual == nil {
+		return "Expected error not to have occurred"
+	} else {
+		return fmt.Sprintf("Expected\n\t%#v\n not to contain \n\t%#v", actual.(error).Error(), matcher.expected)
```
cirocosta, line 59: I think casting to `error` is not necessary:

```go
		// If a string is acceptable according to the format, see if
		// the value satisfies one of the string-valued interfaces.
		// Println etc. set verb to %v, which is "stringable".
		switch verb {
		case 'v', 's', 'x', 'X', 'q':
			// Is it an error or Stringer?
			// The duplication in the bodies is necessary:
			// setting handled and deferring catchPanic
			// must happen before calling the method.
			switch v := p.arg.(type) {
			case error:
				handled = true
				defer p.catchPanic(p.arg, verb, "Error")
				p.fmtString(v.Error(), verb)
				return
```

https://github.com/golang/go/blob/ab7c904bf2bb883b83001d8bb188d1809d74eee6/src/fmt/print.go#L612-L621

cirocosta: (unless we want the property of getting a `panic` in case a non-error type is passed through that `interface{}`)

cirocosta:
good stuff! really happy to see stamper now being generic.

I added a few commits to solve some things that would fail during lint, I hope you don't mind (if  you do, feel free to force push, no worries!)

just left a nitpicking comment, setting approved as overall lgtm 👍

cirocosta (approved):

pkg/templates/stamper_test.go
```diff
@@ -18,14 +18,92 @@ import (
...
+			It("sets the owner reference in the stamped output", func() {
+				template := `{ "kind": "Silly", "apiVersion": "silly.io/v1"}`
+				stamped, err := stamper.Stamp([]byte(template))
+
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(stamped.GetOwnerReferences()).To(HaveLen(1))
+				owner := stamped.GetOwnerReferences()[0]
+
+				Expect(owner).To(MatchFields(IgnoreExtras, Fields{
```
cirocosta, line 64: neat! i wasn't aware of this (https://pkg.go.dev/github.com/onsi/gomega/gstruct)

pkg/utils/matchers/meaningful_error.go
squeedee: Correct. If you use a non error, you don't understand the matcher. It panics and you're made to review what you thought the matcher was for.

PR #17, opened 2021-09-01
introduce pipeline

### LINKED ISSUE #6: Introduce Pipeline

squeedee:
Introduce pipeline.

closes #6

~Currently blocked on: https://github.com/vmware-tanzu/cartographer/pull/16~


cirocosta:
@emmjohnson I just added copyright so we can get the tests moving, will go through it 👍

experiments/pipeline-service/pkg/apis/v1alpha1/run_template.go
```diff
@@ -51,7 +51,7 @@ type RunTemplateSpec struct {
 	//
 	// To this template are made available:
 	//
-	// 	- the pipeline object that is referencing this RunTemplate
+	// 	- the pipeline object that is referencing this RunTemplateRef
```
cirocosta: whops? I think this came from a "replace all"

```console
$ git diff main..6-introduce-pipeline  --stat
 experiments/pipeline-service/pkg/apis/v1alpha1/run_template.go                   |   4 +-
 experiments/pipeline-service/pkg/apis/v1alpha1/zz_generated.deepcopy.go          |   2 +-
```

i think there should be no changes here

cirocosta:
have you been able to consistently run the integration tests? I quite often hit this snag

```
Summarizing 1 Failure:

[Fail] Stamping a resource on Pipeline Creation a RunTemplate that produces a Resource a Pipeline that matches the RunTemplateRef [It] Stamps a new Resource
/home/cirocosta/dev/vmware-tanzu/cartographer/tests/integration/pipeline_service/pipeline_service_test.go:107

Ran 2 of 2 Specs in 10.880 seconds
FAIL! -- 1 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestWebhookIntegration (10.88s)
FAIL
```

seems like there's some big flakiness there

pkg/apis/v1alpha1/pipeline.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type PipelineSpec struct {
+	// +kubebuilder:validation:Required
+	RunTemplateRef TemplateReference `json:"runTemplateRef"`
+}
+
+type TemplateReference struct {
+	// +kubebuilder:validation:Enum=RunTemplate
+	Kind string `json:"kind"`
```
cirocosta: currently, this will end up generating a field in the CustomResourceDefinition that marks this as required, but I think we should make it so it's not really given that, like mentioned in the `validation`, this has only one possibility and must always be filled

currently:

```yaml
# schema.openapiv3schema.properties.spec.runtemplateref.properties ...
                required:
                - kind
                - name
                - namespace
```

so my suggestion would be

```suggestion
	Kind string `json:"kind,omitempty"`
```

and then internally deal with it, or (better, imo), get rid of this completely (especially as we're not considering `apiVersion`, thus, not fully being an "open" reference)

pkg/apis/v1alpha1/pipeline.go
```diff
@@ -0,0 +1,79 @@
...
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// +kubebuilder:validation:MinLength=1
+	Namespace string `json:"namespace"`
```
cirocosta: in the same line of thinking of the comment about `kind`, I think we could make the ux better here by making the namespace not required, but rather optional, defaulting to  the same namespace as the pipeline

```suggestion
	Namespace string `json:"namespace,omitempty"`
```

pkg/apis/v1alpha1/pipeline.go
```diff
@@ -0,0 +1,79 @@
...
+
+const (
+	ReadyRunTemplateReason                            = "Ready"
+	NotFoundRunTemplateReason                         = "RunTemplateNotFound"
+	StampedObjectRejectedByAPIServerRunTemplateReason = "StampedObjectRejectedByAPIServer"
+	TemplateStampFailureRunTemplateReason             = "TemplateStampFailure"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
```
cirocosta, line 38: I usually expect to see objects to share information about their age and current highest-level status, so I usually add

```go
// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type==\"Ready\")].status`
// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```

this is something we currently _don't_ do for Workload and ClusterSupplyChain, but I think we should - makes the `kubectl` experience much better imo

pkg/controller/pipeline/reconciler.go
```diff
@@ -0,0 +1,79 @@
...
+	repository repository.Repository
+	realizer   realizerpipeline.Realizer
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	logger := logr.FromContext(ctx).
+		WithValues("name", request.Name, "namespace", request.Namespace)
+	logger.Info("started")
+
+	pipeline, getPipelineError := r.repository.GetPipeline(request.Name, request.Namespace)
```
cirocosta: any reason for not using `err` here?

usually errors are not named as `err` when there's a very good reason for it (e.g., to not replace an error that was captured before that we want to return at a later point ...), but this doesn't seem to be the case

```suggestion
	pipeline, err := r.repository.GetPipeline(request.Name, request.Namespace)
```

pkg/controller/pipeline/reconciler.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if kerrors.IsNotFound(getPipelineError) {
+		logger.Info("pipeline no longer exists")
+		logger.Info("finished")
```
cirocosta: rather than adding `finished` every time we're going to return, it might be worth instead going for `defer` right after `"started"` at the beginning

e.g. 

```diff
 func (r *reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
         logger := logr.FromContext(ctx).
                 WithValues("name", request.Name, "namespace", request.Namespace)
         logger.Info("started")
+        defer logger.Info("finished")
```

then removing all the other `finished` calls

---

_(yep, `defer`s in hot-paths were frowned upon, but now they're quite optimized https://go-review.googlesource.com/c/go/+/202340/)_

pkg/controller/pipeline/reconciler_test.go
```diff
@@ -0,0 +1,220 @@
...
+	. "github.com/onsi/gomega/gstruct"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
+	. "github.com/vmware-tanzu/cartographer/pkg/controller/pipeline"
```
cirocosta: imo we should restrict dot imports to only two ocasions:

- dsls (like, ginkgo's / heredoc - actually, a bit controversial out there; imo too much dsl looks bad, but that's just me I guess)
- import cycles (see https://github.com/golang/go/wiki/CodeReviewComments#import-dot)

regardless, currently we don't do dot imports very often for the package we're testing:

```
$ find . -name "*_test.go" | echo | xargs -n1 ag -Q '. "github.com/vmware-tanzu/cartographer'
conditions/condition_manager_test.go
23:     . "github.com/vmware-tanzu/cartographer/pkg/conditions"

apis/v1alpha1/common_test.go
24:     . "github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"

templates/interpolator_test.go
28:     . "github.com/vmware-tanzu/cartographer/pkg/utils/matchers"   //  ok, DSL, so .. a pass?

controller/pipeline/reconciler_test.go
34:     . "github.com/vmware-tanzu/cartographer/pkg/controller/pipeline"    // this one
```

out of 43 go test files

```go
$ find . -name "*_test.go" | wc -l
43
```

so I'd suggest that we don't do a dot import here as it looks like it's not necessary

pkg/controller/workload/reconciler.go
```diff
@@ -25,7 +25,7 @@ import (
 
 	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
 	"github.com/vmware-tanzu/cartographer/pkg/conditions"
-	"github.com/vmware-tanzu/cartographer/pkg/realizer"
+	realizerworkload "github.com/vmware-tanzu/cartographer/pkg/realizer/workload"
```
cirocosta: I noticed that under each package we're naming the import after the name of the controller

I wonder if that's a smell that the way we broke the modules down are not after the domain they're working on, but instead, the "thing they are"

regardless, I'd suggest keeping

```suggestion
	realizer "github.com/vmware-tanzu/cartographer/pkg/realizer/workload"
```

but no strong opinions really

pkg/realizer/pipeline/realizer_test.go
```diff
@@ -0,0 +1,215 @@
...
+	Context("with a valid RunTemplate", func() {
+		BeforeEach(func() {
+			templateAPI := &v1alpha1.RunTemplate{
+				Spec: v1alpha1.RunTemplateSpec{
+					Template: runtime.RawExtension{
+						Raw: []byte(D(`{
+								"apiVersion": "v1",
+								"kind": "ConfigMap",
+								"metadata": { "generateName": "my-stamped-resource-" },
+								"data": { "has": "data" }
```
cirocosta, line 71: _very very_ minor nit

```diff
diff --git a/pkg/realizer/pipeline/realizer_test.go b/pkg/realizer/pipeline/realizer_test.go
index 17c591a..21d2162 100644
--- a/pkg/realizer/pipeline/realizer_test.go
+++ b/pkg/realizer/pipeline/realizer_test.go
@@ -69,7 +69,8 @@ var _ = Describe("Reconcile", func() {
                                                                "kind": "ConfigMap",
                                                                "metadata": { "generateName": "my-stamped-resource-" },
                                                                "data": { "has": "data" }
-                                                       }`)),
+                                                       }`,
+                                               )),
                                        },
                                },
                        }
```

pkg/registrar/registrar.go
```diff
@@ -115,6 +121,28 @@ func registerSupplyChainController(mgr manager.Manager) error {
...
+	})
+	if err != nil {
+		return fmt.Errorf("controller new [pipeline-service]: %w", err)
+	}
+
+	if err := ctrl.Watch(
+		&source.Kind{Type: &v1alpha1.Pipeline{}},
+		&handler.EnqueueRequestForObject{},
+	); err != nil {
+		return fmt.Errorf("watch [pipeline-service]: %w", err)
```
cirocosta, line 138: I don't think we use `[__]` anywhere

```console
$ ag -Q ' [' | grep fmt | grep \\[
registrar/registrar.go:131:             return fmt.Errorf("controller new [pipeline-service]: %w", err)
registrar/registrar.go:138:             return fmt.Errorf("watch [pipeline-service]: %w", err)
```

```suggestion
		return fmt.Errorf("watch pipeline-service: %w", err)
```

same for the error above on line 131

pkg/repository/repository.go
```diff
@@ -118,6 +126,29 @@ func (r *repository) GetTemplate(ref v1alpha1.TemplateReference) (templates.Temp
...
 }
 
+// FIXME: Probably should not return a templates.Template, not sure it adheres to interface
+func (r *repository) GetTemplate(ref v1alpha1.TemplateReference) (templates.Template, error) {
+	apiTemplate, err := v1alpha1.GetAPITemplate(ref.Kind)
+	if err != nil {
+		return nil, fmt.Errorf("get api template: %w", err)
+	}
+
+	err = r.cl.Get(context.TODO(), client.ObjectKey{
```
cirocosta, line 136: for me the rule of thumb is that for anything that  might be blocking and you / some caller might want to cancel, a context should be passed from the arguments down to that function[^1], which nowadays is pretty much what you'd expect for any go library touching the network (yeah i know, it _might_ not touch the network given the cache, but, interface-wise, it should take it anyway imo)

I think there's a bigger story here for passing `ctx` to all methods of this interface, but, leaving the comment so we don't forget about it

[^1]: sure, there are exceptions like  filesystem syscalls that yeah .. can block .. might block ... but standard go apis don't give you that granularity and .. it's very unlikely that there'll be problems

pkg/repository/repository.go
```diff
@@ -118,6 +126,29 @@ func (r *repository) GetTemplate(ref v1alpha1.TemplateReference) (templates.Temp
...
+		Name:      ref.Name,
+		Namespace: ref.Namespace,
+	}, apiTemplate)
+	if err != nil {
+		return nil, fmt.Errorf("get: %w", err)
+	}
+
+	template, err := templates.NewModelFromAPI(apiTemplate)
+	if err != nil {
+		return nil, fmt.Errorf("NewModelFromAPI: %w", err)
```
cirocosta: i don't think we wrap with pascal-case anywhere, e.g.

```console
$ ag -Q 'fmt.Errorf'
controller/workload/reconciler.go
63:             return ctrl.Result{}, fmt.Errorf("get workload: %w", err)
75:             return r.completeReconciliation(reconcileCtx, workload, fmt.Errorf("get object gvk: %w", err))
126:                            return ctrl.Result{}, fmt.Errorf("update workload status: %w", updateErr)
138:            return ctrl.Result{}, fmt.Errorf("workload not ready")
150:    return fmt.Errorf("supply-chain is not in ready condition")
156:            return nil, fmt.Errorf("workload is missing required labels")
164:                    return nil, fmt.Errorf("get supply chain by label: %w", err)
166:                    return nil, fmt.Errorf("no supply chain found where full selector is satisfied by labels: %v", workload.Labels)
170:            return nil, fmt.Errorf("too many supply chains match the workload selector")

registrar/map_functions.go
53:             mapper.Logger.Error(fmt.Errorf("client list: %w", err), "cluster supply chain to workload requests: client list")

registrar/registrar.go
50:             return fmt.Errorf("cartographer v1alpha1 add to scheme: %w", err)
58:             return fmt.Errorf("register workload controller: %w", err)
62:             return fmt.Errorf("register supply-chain controller: %w", err)
66:             return fmt.Errorf("register pipeline-service controller: %w", err)
79:             return fmt.Errorf("controller new: %w", err)
86:             return fmt.Errorf("watch: %w", err)
98:             return fmt.Errorf("watch: %w", err)
111:            return fmt.Errorf("controller new: %w", err)
118:            return fmt.Errorf("watch: %w", err)
131:            return fmt.Errorf("controller new [pipeline-service]: %w", err)
138:            return fmt.Errorf("watch [pipeline-service]: %w", err)
150:            return fmt.Errorf("index supply chain resource: %w", err)
168:            return fmt.Errorf("index field supply-chain.selector: %w", err)
```

which closely follows go std libs, e.g.,

```
func main() {
        err := http.ListenAndServe("__", nil)
        fmt.Println(err)
}
```

yields

```
listen tcp: address __: missing port in address
```

suggestion:

```suggestion
		return nil, fmt.Errorf("new model from api: %w", err)
```

tests/integration/pipeline_service/pipeline_service_test.go
```diff
@@ -0,0 +1,152 @@
...
+				apiVersion: carto.run/v1alpha1
+				kind: RunTemplate
+				metadata:
+				  namespace: %s
+				  name: my-run-template
+				spec:
+  				  template:
+  				    apiVersion: v1
+  				    kind: ConfigMap
+  				    metadata:
```
cirocosta: I think the reason why this looks strange is due to the mix of tabs and spaces

currently, looks like:
Screen Shot 2021-09-08 at 5 08 48 PM

but should be:

Screen Shot 2021-09-08 at 5 08 10 PM

(^ output from `:set list` in vim)

cirocosta (commented):
left a few comments, mostly nitpicking, but if there's one thing I'd take a deeper look into is the requirement of `kind` and `namespace` in the `runTemplateRef` - I think `kind` should not exist and `namespace` should be optional defaulting to `default`

let me know what you think 😁  good stuff! thx!

pkg/repository/repository.go
emmjohnson: Going to leave this for now and create a new issue to talk about context in repository. Currently they are all TODO

emmjohnson: Not going to address this now. I have opened up issue - https://github.com/vmware-tanzu/cartographer/issues/89

pkg/apis/v1alpha1/pipeline.go
emmjohnson: We have created an issue for this - https://github.com/vmware-tanzu/cartographer/issues/90

pkg/controller/workload/reconciler.go
emmjohnson: We definitely have some work/refactoring to do in the realizer. I have changed all the import names except in registrar.go

emmjohnson (approved):

emmjohnson:
> have you been able to consistently run the integration tests? I quite often hit this snag
> 
> ```
> Summarizing 1 Failure:
> 
> [Fail] Stamping a resource on Pipeline Creation a RunTemplate that produces a Resource a Pipeline that matches the RunTemplateRef [It] Stamps a new Resource
> /home/cirocosta/dev/vmware-tanzu/cartographer/tests/integration/pipeline_service/pipeline_service_test.go:107
> 
> Ran 2 of 2 Specs in 10.880 seconds
> FAIL! -- 1 Passed | 1 Failed | 0 Pending | 0 Skipped
> --- FAIL: TestWebhookIntegration (10.88s)
> FAIL
> ```
> 
> seems like there's some big flakiness there

So this is indeed not flakey and just a real failure. The update status call is requeueing the pipeline for reconcillation. This is in turn creating another stamped resource for the runtemplate. This is probably an infinite cycle. We will address this in very next PR. 😔 

Making the test "pass" for now.

cirocosta:
thx for taking the comments into consideration! 

current check failure:

Screen Shot 2021-09-09 at 12 05 24 PM


blaaaah 😅 

---

> _**emmjohnson** [...] This is probably an infinite cycle. We will address this in very next PR. 😔_

no worries, sgtm 😁

cirocosta (approved):

PR #60, opened 2021-09-02
After CreateOrPatchUnstructuredObject obj is always populated

waciumawanjohi:
In the case that the apiServer returns a previously created object
and that object does not need to be altered, ensure that object
is available to the calling function (and subsequently the
StampContext).

### DIFF
--- a/pkg/repository/repository.go
+++ b/pkg/repository/repository.go
@@ -17,6 +17,7 @@ package repository
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	api_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -65,6 +66,11 @@ func (r *repository) CreateOrPatchUnstructuredObject(obj *unstructured.Unstructu
 		return err
 	} else if r.rc.UnchangedSinceCached(submitted, existingUnstructured) {
 		r.rc.Refresh(submitted)
+
+		objVal := reflect.ValueOf(obj)
+		existingVal := reflect.ValueOf(existingUnstructured)
+		reflect.Indirect(objVal).Set(reflect.Indirect(existingVal))
+
 		return nil
 	} else {
 		createOrPatchErr = r.patchUnstructured(existingUnstructured, obj)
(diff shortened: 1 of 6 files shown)


pkg/repository/repository.go
```diff
@@ -65,6 +66,11 @@ func (r *repository) CreateOrPatchUnstructuredObject(obj *unstructured.Unstructu
 		return err
 	} else if r.rc.UnchangedSinceCached(submitted, existingUnstructured) {
 		r.rc.Refresh(submitted)
+
+		objVal := reflect.ValueOf(obj)
+		existingVal := reflect.ValueOf(existingUnstructured)
+		reflect.Indirect(objVal).Set(reflect.Indirect(existingVal))
+
```
squeedee: I understand this approach copies the complete value into the input reference. But it took me a while to understand this code? Perhaps:

1. Pull it as a utility function that is well documented to explain what is happening here?
2. OR, change the whole signature of CreateOrPatchUnstricturedObject to return the 'fully realize' shape of the object instead of mutating the input parameter.

Neither is blocking, I just think it's worth considering because this is "surprising and complex" code. It's intent is not immediately clear.

squeedee (approved):

pkg/repository/repository.go
cirocosta: > _I just think it's worth considering because this is "surprising and complex" code. It's intent is not immediately clear._

+1

could we do 

```go
                *obj = *existingUnstructured
```

instead? not sure if i'm missing something 🤔  

even if that's not sufficient, I'd really try to avoid introducing any use of reflect - as mentioned, always surprising, easy to mess up etc

pkg/repository/repository.go
```diff
@@ -79,6 +83,12 @@ func (r *repository) CreateOrPatchUnstructuredObject(obj *unstructured.Unstructu
 	return nil
 }
 
+func updateObjWithValuesFromAPIServer(obj *unstructured.Unstructured, existingUnstructured *unstructured.Unstructured) {
+	objVal := reflect.ValueOf(obj)
+	existingVal := reflect.ValueOf(existingUnstructured)
+	reflect.Indirect(objVal).Set(reflect.Indirect(existingVal))
```
cirocosta, line 89: (whops, comment got "eaten" by the changes - https://github.com/vmware-tanzu/cartographer/pull/60#discussion_r701454835)

PR #73, opened 2021-09-07
Introduce rfc-0006-replacing-sources-index

waciumawanjohi:


DanieldeR:
@jwntrs as discussed during IPM could be create some stories for this RFC?

emmjohnson:
Team consensus during pre-IPM - we will move all things that are currently lists that we want to be maps to maps. We will not support aliases for single objects in maps. (i.e. image.image).

DanieldeR:
Issue #103 was raised to implement this RFC.

jwntrs (approved):

PR #74, opened 2021-09-07
Introduce rfc-0007-support-service-definitions

waciumawanjohi:


DanieldeR:
As discussed at Pre-IPM, the scope of this RFC for Beta may be documentation only.

PR #75, opened 2021-09-07
Introduce rfc-0009-supply-chain-switches-template-on-flag

waciumawanjohi:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-0009-supply-chain-switches-template-on-flag/rfc/rfc-0009-supply-chain-switches-template-on-flag.md)

### DIFF
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -95,7 +95,7 @@ spec:
 ```
 
 1. As labels are present in the workload metadata, any behavior possible using just label selectors
-  will be possible using the entire templating context.
+  will be possible using the entire workload.
 
 ### Error conditions
 
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -1,4 +1,4 @@
-# Draft RFC 0009 Supply Chain Switches Template in single step
+# RFC 0009 Supply Chain Switches Templates
 
 ## Summary
 
@@ -45,6 +45,16 @@ would represent a code smell that this primitive is not adequate for Cartographe
 When creating objects of the supply chain, Cartographer has the workload definition available.
 Allowing the supply chain to leverage the workload would add additional power to the conditional.
 
+Each step/resource in a supply chain will continue to define a templateRef, which will continue to
+specify a single template kind (e.g. ClusterImageTemplate). Rather than specifying the name of a
+particular template, authors can provide `options` which is a list of options. An option is
+a `name` and a `selector`. The only selector adopted by this RFC is `MatchFields` (though other
+selectors can be added later). A MatchFields contains:
+- `key`: a path on the workload object. The path must be prefixed by `workload`.
+- `operator`: One of `In`, `NotIn`, `Exists`, `DoesNotExist`
+- `values`: Required and allowed only when the operator is either `In` or `NotIn`. A list of json values.
+  (e.g. both strings and integers are valid values)
+
 ```yaml
 apiVersion: kontinue.io/v1alpha1
 kind: ClusterSupplyChain
@@ -60,12 +70,12 @@ spec:
         options:                                  # <--- a list
         - name: git-template
           selector:
-            matchExpressions:
+            matchFields:
               - key: workload.spec.source.git       # <--- path to field in template context
                 operator: Exists
         - name: imgpkg-bundle-template
           selector:
-            matchExpressions:
+            matchFields:
               - key: workload.spec.source.image       # <--- path to field in template context
                 operator: Exists
 
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -47,7 +47,7 @@ metadata:
 spec:
   selector:
     app: web
-  components:
+  resources:
     - name: source-provider
       templateRef:
         kind: ClusterSourceTemplate
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -53,13 +53,15 @@ spec:
         kind: ClusterSourceTemplate
         options:                                  # <--- a list
         - name: git-template
-          matchExpressions:
-            - key: workload.spec.source.git       # <--- path to field in template context
-              operator: Exists
+          selector:
+            matchExpressions:
+              - key: workload.spec.source.git       # <--- path to field in template context
+                operator: Exists
         - name: imgpkg-bundle-template
-          matchExpressions:
-            - key: workload.spec.source.image       # <--- path to field in template context
-              operator: Exists
+          selector:
+            matchExpressions:
+              - key: workload.spec.source.image       # <--- path to field in template context
+                operator: Exists
 
 ---
 apiVersion: kontinue.io/v1alpha1
@@ -79,20 +81,17 @@ spec:
 1. As labels are present in the workload metadata, any behavior possible using just label selectors
   will be possible using the entire templating context.
 
-## Open choices
+### Error conditions
 
-1. Should options be an ordered list?
-   1. Having options as an ordered list allows 'tie-breakers'. If two options are fulfillable, the earlier one in the
-      list would be chosen.
-   2. Users may be least surprised if options is not ordered. In this case, if multiple options are fulfillable,
-      Cartographer would throw a helpful error.
-2. What syntax should we use for matching fields?
-   1. Cartographer currently has ObservedConditions specifying a `key` and `value`. A condition evaluates to true
-      when the value found at the specified key (path) on the object matches the given value.
-   2. Kubernetes resources "such as Job, Deployment, ReplicaSet, and DaemonSet, support set-based requirements". The
-      syntax for matchExpressions allows for asserting `In` or `NotIn` a specified array of values, or to assert
-      that a key `Exists` or `DoesNotExist`. matchExpressions is a list of assertions, which are ANDed.
-      [link](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) 
+Exactly one template should be chosen at each step. If no template is chosen, or multiple templates match, an error
+should be thrown.
+
+### Matching syntax
+
+Kubernetes resources "such as Job, Deployment, ReplicaSet, and DaemonSet, support set-based requirements". The
+syntax for matchExpressions allows for asserting `In` or `NotIn` a specified array of values, or to assert
+that a key `Exists` or `DoesNotExist`. matchExpressions is a list of assertions, which are ANDed.
+[link](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements)
 
 ## Notes
 
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -1,3 +1,124 @@
-[Draft PR](https://github.com/vmware-tanzu/cartographer/pull/75)
+# Draft RFC <0009> Supply Chain Switches Template Dynamically
 
-[Draft Text](https://github.com/vmware-tanzu/cartographer/blob/rfc-0009-supply-chain-switches-template-on-flag/rfc/rfc-0009-supply-chain-switches-template-on-flag.md)
+## Summary
+
+The Developer Productivity team would like to be able to write 1 supply chain
+which can dynamically use Template A or Template B in a component depending
+on the state of the workload. This RFC proposes methods for achieving that
+outcome.
+
+## Motivation
+
+The Developer Productivity program auto-generates workloads from CLI commands.
+Those workloads match with supplychains and templates that are written by the
+dev-prod team as well. The team would liek to limit the number of supplychains
+that must be written and maintained. Small differences in workload can lead to
+needing an entirely different supply chain. For example, what code source is used.
+
+Further info on their use case can be read [here](https://docs.google.com/document/d/1TVqlNqTyCMlp_yNs9_F80QAIQSC5bxm05vSyWyeGMSw/edit).
+See also this [discussion](https://vmware.slack.com/archives/C01UX69LJCB/p1629408443051200).
+
+## Possible Solutions
+
+Kontinue could choose which template to use based on a flag specified in the workload.
+Alternatively, Kontinue could attempt to deploy the default template and if the necessary
+values are not found, fall back to attempt a different template.
+
+### Switch on Flag
+A supply chain component could reference a template based on the presence of a
+flag in the workload. Below is a proposed example using the keyword `requisite`:
+
+```yaml
+apiVersion: kontinue.io/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: responsible-ops
+spec:
+  selector:
+    app: web
+  components:
+    - name: source-provider
+      templateRefs:                               # <--- now a list
+      - kind: ClusterSourceTemplate
+        name: git-template
+        requisite:
+          path: workload.spec.flags.source-type    # <--- path to field in workload
+          value: git                              # <--- value that will trigger use of this template
+      - kind: ClusterSourceTemplate
+        name: imgpkg-bundle-template
+        requisite:
+          path: workload.spec.flags.source-type    # <--- ideally path is same for all templateRefs in component
+          value: imgpkg-bundle
+
+---
+apiVersion: kontinue.io/v1alpha1
+kind: Workload
+metadata:
+  name: petclinic
+  labels:
+    app: web
+spec:
+  source:
+    git:
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
+  flags:                                           # <--- map[string]string
+    source-type: git
+```
+
+### Switch through fall-back strategy
+Kontinue could allow users to define fall-back values if a given value does not work.
+This could happen in 2 different ways:
+- Allow a supply-chain component to define a fallback template if the first creation attempt fails
+- Allow a template to define a fallback value if the first templated value does not resolve
+
+#### Supply Chain defines fall-back template
+Replace the single template ref for each component with a list of template refs. Kontinue would try them in order and when one successfully deploys, it would proceed to creating the next component, skipping the other template refs for the current component. An example definition:
+
+```
+apiVersion: kontinue.io/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: responsible-ops
+spec:
+  selector:
+    ...
+  components:
+    - name: source-provider
+      templateRefs:
+      - kind: ClusterSourceTemplate
+        name: git-resource-template
+      - kind: ClusterSourceTemplate
+        name: imgpkg-template
+      - kind: ClusterSourceTemplate
+        name: image-template
+```
+
+#### Template defines a fallback value
+a BuildTemplate might have some definition:
+
+```
+apiVersion: kontinue.io/v1alpha1
+kind: ClusterBuildTemplate
+metadata:
+  name: example-build
+spec:
+  params:
+    - name: url
+      default: some-url
+  template:
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: example-build-configmap
+    data:
+      templated-field: $(sources[0].url|params[?(@.name=="url")].value)$
+  imagePath: ...
+```
+
+We can see the introduction of the `|` character to separate possible paths to use to fill this field.
+
+## Cross References and Prior Art
+
+{{Reference other similar implementations, and resources you are using to draw inspiration from}}
--- a/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
+++ b/rfc/rfc-0009-supply-chain-switches-template-on-flag.md
@@ -1,10 +1,11 @@
-# Draft RFC <0009> Supply Chain Switches Template Dynamically
+# Draft RFC 0009 Supply Chain Switches Template in single step
 
 ## Summary
 
-Platform creators will need to handle multiple types of workloads. Cartographer
-must support that behavior through either multiple supply chains, or with
-supply chains that can substitute templates dynamically.
+Platform creators will need to handle multiple types of workloads. While Cartographer
+supports that behavior currently through allowing multiple supply chains, it should
+reduce the number of supply chains necessary by allowing individual supply chains
+to choose between templates within a step.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,124 @@
...
+dev-prod team as well. The team would liek to limit the number of supplychains
+that must be written and maintained. Small differences in workload can lead to
+needing an entirely different supply chain. For example, what code source is used.
+
+Further info on their use case can be read [here](https://docs.google.com/document/d/1TVqlNqTyCMlp_yNs9_F80QAIQSC5bxm05vSyWyeGMSw/edit).
+See also this [discussion](https://vmware.slack.com/archives/C01UX69LJCB/p1629408443051200).
+
+## Possible Solutions
+
+Kontinue could choose which template to use based on a flag specified in the workload.
```
waciumawanjohi: @waciumawanjohi said:
> 
I prefer the strategy of using a flag specified in the workload. Developer intent is clear. It also means that error handling is more straightforward.

waciumawanjohi: @jwntrs replied:
>
I'm also in favour of being explicit in selecting the template, although I don't know if it belongs in the workload, since developers might not know (or be allowed) to select a template.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,124 @@
...
+    git:
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
+  flags:                                           # <--- map[string]string
+    source-type: git
+```
+
+### Switch through fall-back strategy
+Kontinue could allow users to define fall-back values if a given value does not work.
```
waciumawanjohi: @waciumawanjohi said:
> I think the error reporting with this option would quickly become a mess:

`
supply chain failed to be deployed because template ref 1's value 'git.url' could not be found, nor could template ref 2's value 'imgpkg.url'...
`

waciumawanjohi: @jwntrs replied: 
>Having a fall-back strategy to me indicates an order. I'm not sure any one template is more important than others. So to me this doesn't really fit.
>
> It reminds of buildpacks (and maybe that's the intention here). But I think auto-detecting buildpacks only works because you can override that decision at any point.
>
> So I think even if we go with this strategy we will need some way to override.

waciumawanjohi: @waciumawanjohi replied:
> 
We could un-order things by declaring that 1 and only 1 successful template can be found. E.g. we have a list of templates to try. If 0 can be successfully interpolated, we error. If more than 1 can be successfully interpolated, we error. If exactly 1 can be successfully interpolated, we create it and move to the next component.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,124 @@
...
+spec:
+  selector:
+    app: web
+  components:
+    - name: source-provider
+      templateRefs:                               # <--- now a list
+      - kind: ClusterSourceTemplate
+        name: git-template
+        requisite:
+          path: workload.spec.flags.source-type    # <--- path to field in workload
```
waciumawanjohi: @squeedee said
At the least I would prefer: `workload.spec.source.type`

waciumawanjohi: @waciumawanjohi replied:
I'm somewhat agnostic to the name of the field. But, again because the dev prod team is already considering the need for this feature in specifying runtimes, I think the field should have a general name.

waciumawanjohi: @squeedee replied:
I am focusing on the location, not the name.. I think it should be a *child* of `source` (which exists in the spec today) rather than be in a sibling of `source` called `flags`

waciumawanjohi: @jwntrs replied
We're selecting the workload from a label. Maybe we should select the template the same way.

waciumawanjohi: @waciumawanjohi opined:

`labels` like `flags` are a map[string]string, so I'm amenable to that swap.

While I'm not sure the field should be under `metadata` (as labels are), I don't think it should be a child of `spec.source` because there may be use cases where ops teams want to hot-swap build templates in the same way they now want to hot-swap source templates.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,124 @@
...
+
+## Possible Solutions
+
+Kontinue could choose which template to use based on a flag specified in the workload.
+Alternatively, Kontinue could attempt to deploy the default template and if the necessary
+values are not found, fall back to attempt a different template.
+
+### Switch on Flag
+A supply chain component could reference a template based on the presence of a
+flag in the workload. Below is a proposed example using the keyword `requisite`:
```
waciumawanjohi: @squeedee expounded, at length:

I would prefer to infer the type than use flags or a type field in the workload:

Inferr Git:
```yaml
source:
    git:
      url: https://github.com/spring-projects/spring-petclinic.git
      ref:
        branch: main
```

Infer imgpkg:
```yaml
source:
    imgpkgbundle:
      image: https://github.com/spring-projects/spring-petclinic.git
```

where defining both `source.imgpkgbundle` and `source.git` results in an invalid workload.


Then keeping with this implementation in the ClusterSupplyChain :
```yaml
 - kind: ClusterSourceTemplate
        name: git-template
        sourceType: [git|imgpkg]
```

Although I'd like to explore not modifying the ClusterSupplyChain, and instead add more templating support to Templates:

```yaml
kind: ClusterSourceTemplate
metadata:
  name: source
$(if !empty(workload.spec.source.imgPkg))$
spec:
  ... imgpkg spec ...
$(else)$
spec:
  urlPath: .status.artifact.url
  revisionPath: .status.artifact.revision

  template:
    kind: GitRepository
    metadata:
    spec:
      interval: 1m
      url: $(workload.spec.source.git.url)$
      ref: $(workload.spec.source.git.ref)$
      gitImplementation: libgit2
      ignore: ""
$(end)$
```

Of course that means we need to step up with making our templating much better (ideally replacing with something yaml aware, like ytt), and other work would benefit, such as Service Claims/Binding support.

waciumawanjohi: @jwntrs surprised with this response:
The direction I've been given is that this type of rich templating is actually an anti-goal. And we should stick with "just jsonpath" because it's simple, and harder to shoot yourself in the foot.

waciumawanjohi: @waciumawanjohi went into detail: 

This sounds like it can be achieved with the fall-back strategy and a validation to enforce "defining both `source.imgpkgbundle` and `source.git` results in an invalid workload."
I think the fallback strategy is perfectly reasonable. Validations are reasonable, but I don't think it should be achieved with a rule in a webhook. Rather, if both values are defined stamping should fail when the controller tries to create the object in question. Otherwise, we would be coding for this particular use case rather than a general case.
A user team is already considering multiple dimensions for switching a template based on the workload. We don't want to alter the controller/validations every time they come up with a new dimension.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,124 @@
...
+## Motivation
+
+The Developer Productivity program auto-generates workloads from CLI commands.
+Those workloads match with supplychains and templates that are written by the
+dev-prod team as well. The team would liek to limit the number of supplychains
+that must be written and maintained. Small differences in workload can lead to
+needing an entirely different supply chain. For example, what code source is used.
+
+Further info on their use case can be read [here](https://docs.google.com/document/d/1TVqlNqTyCMlp_yNs9_F80QAIQSC5bxm05vSyWyeGMSw/edit).
+See also this [discussion](https://vmware.slack.com/archives/C01UX69LJCB/p1629408443051200).
```
waciumawanjohi: @cirocosta got his ascii-art on with this comment:

just to be sure: are we focusing on the case of _only_ swapping the underlying template under the hood _or_ mutating the supply chain component set?

e.g., given the following supplychain

```
source-provider <--[src]-- image-provider <--[img]-- deployer
    |ref1                    |ref2                     |ref3
gitrepository              kpack/image               apps/deployment
```

are we focusing _solely_ on replacing `ref1`  to stamp something out like `BundleRepository` rather than `GitRepository` in the case of `--imgpkg` rather than `--git` provided via the `tanzu cli` because we want to provide some local source code


```
source-provider <--[src]-- image-provider <--[img]-- deployer
    |new_ref1                   |ref2                     |ref3
bundlerepository              kpack/image               apps/deployment
```
  
_or_, are we talking about mutating the components DAG entirely because with a certain workload having some components don't make sense, like, using `--image` in the `tanzu cli` rather than `--git` or `--imgpkg` because we're providing locally built images (thus, no reason for involving `kpack` or a "repository"):

```
 deployer
    |ref3
apps/deployment
```

?

From the document linked, I imagine both scenarios would be desirable by the developer productivity team, so I think it might be worth calling out explicitly that it's _just_ about the template refs if that's the case

waciumawanjohi: @jwntrs responded:

I think the idea of dynamically changing the shape of the SupplyChain could get really complicated. That second use case seems like a better fit for composable SupplyChains.

waciumawanjohi: @cirocosta agreed:

+1

I'm quite sure the intent of the RFC is indeed just the first (template refs), but I believe it's a good thing to explicitly call that out

waciumawanjohi: @waciumawanjohi also agreed:

Agree with you both:

1. This RFC is only about swapping the templates of a single component.
1. If the [RFC for SupplyChainSnippets](https://github.com/vmware-tanzu/cartographer/blob/main/rfc/rfc-0005-support-composing-supply-chains.md) is accepted, users will then be able to swap whole sections of SupplyChains (as a snippet is proposed to act just like a template).

squeedee:
We're landing ytt support. Wondering if it invalidates this approach as we can switch in template on different sources, or does that make it "more than break glass" ?

DanieldeR:
Bumping the milestone on this to 1.0+ as this is achievable with YTT.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,182 @@
...
+
+```yaml
+apiVersion: kontinue.io/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: responsible-ops
+spec:
+  selector:
+    app: web
+  components:
```
martyspiewak: nit:
```suggestion
  resources:
```

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,182 @@
...
+      ref:
+        branch: main
+```
+
+1. As labels are present in the workload metadata, any behavior possible using just label selectors
+  will be possible using the entire templating context.
+
+## Open choices
+
+1. Should options be an ordered list?
```
martyspiewak: Personally, I think a (helpful) error would be least surprising for users. If we do want an ordered list, though, I'd suggest we need some way of communicating that better in the spec, as I don't know that `options` conveys an ordered list.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,182 @@
...
+## Motivation
+
+The Developer Productivity program auto-generates workloads from CLI commands.
+Those workloads match with supplychains and templates that are written by the
+dev-prod team as well. The team would like to limit the number of supplychains
+that must be written and maintained. Small differences in workload can lead to
+needing an entirely different supply chain. For example, what code source is used.
+
+Further info on their use case can be read [here](https://docs.google.com/document/d/1TVqlNqTyCMlp_yNs9_F80QAIQSC5bxm05vSyWyeGMSw/edit).
+See also this [discussion](https://vmware.slack.com/archives/C01UX69LJCB/p1629408443051200).
```
martyspiewak: The doc and slack thread listed here are not publicly viewable.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,182 @@
...
+## Open choices
+
+1. Should options be an ordered list?
+   1. Having options as an ordered list allows 'tie-breakers'. If two options are fulfillable, the earlier one in the
+      list would be chosen.
+   2. Users may be least surprised if options is not ordered. In this case, if multiple options are fulfillable,
+      Cartographer would throw a helpful error.
+2. What syntax should we use for matching fields?
+   1. Cartographer currently has ObservedConditions specifying a `key` and `value`. A condition evaluates to true
+      when the value found at the specified key (path) on the object matches the given value.
```
martyspiewak, line 91: Was there a reason we chose not to follow the k8s conventions for observedCompletion? Or did we just fail to do sufficient research into prior art?

martyspiewak:
Not necessary to resolve this question to approve this RFC but just a thought:
This model seems to push users towards larger blueprints with far more templates referenced in them, where only a subset of those templates will be relevant for a given owner. Perhaps we will need to re-think our informers for templates, then. 

The way our informers work currently, if a template referenced in a blueprint changes, any owner that matched the blueprint will be re-reconciled. After this RFC is implemented, it is plausible that there will be a significant number of templates in a blueprint that are irrelevant to a given owner, and it doesn't seem worth re-reconciling all of those owners when those templates change.

Again, definitely not necessary to address this as part of this RFC, but just wanted to document the thought for posterity if the RFC is approved.

waciumawanjohi:
> After this RFC is implemented, it is plausible that there will be a significant number of templates in a blueprint that are irrelevant to a given owner, and it doesn't seem worth re-reconciling all of those owners when those templates change.

Sounds like another good quality of deterministic supply chains. As long as we can calculate which templates in a supply chain will be chosen for a workload, we can avoid unnecessary reconciles. (Though that does beg the question, how expensive will that calculation be, will we need to cache the calculation somehow?)

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
waciumawanjohi: Removed links to them. Good catch.

emmjohnson (approved):

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+spec:
+  source:
+    git:                                        # <--- Presence of this field will determine the conditional
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
+```
+
+1. As labels are present in the workload metadata, any behavior possible using just label selectors
+  will be possible using the entire templating context.
```
martyspiewak:
```suggestion
  will be possible using the entire workload.
```

martyspiewak (approved):

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+Allowing the supply chain to leverage the workload would add additional power to the conditional.
+
+Each step/resource in a supply chain will continue to define a templateRef, which will continue to
+specify a single template kind (e.g. ClusterImageTemplate). Rather than specifying the name of a
+particular template, authors can provide `options` which is a list of options. An option is
+a `name` and a `selector`. The only selector adopted by this RFC is `MatchFields` (though other
+selectors can be added later). A MatchFields contains:
+- `key`: a path on the workload object. The path must be prefixed by `workload`.
+- `operator`: One of `In`, `NotIn`, `Exists`, `DoesNotExist`
+- `values`: Required and allowed only when the operator is either `In` or `NotIn`. A list of json values.
```
squeedee, line 55: In/NotIn is probably insufficent for a numeric json type
Boolean types could be better matched with boolean operators too. Propose:

* `Exists` `DoesNotExist` are generic (as they do not expect a `values`)
* `In` `NotIn` are suitable for strings, with a `values` field
* `LessThan`, `GreaterThan`, `Equal`, `NotEqual`, `LessThanOrEqual`, `GreaterThanOrEqual` (and/or) `<`, `>`, `==`, `!=`, `<=`, `>=` for numeric types on a field called `value`
* `Is` (and perhaps `IsNot` against a `value` field that can only be `true` or `false` for booleans.

squeedee: Also I don't think this has to be resolved/implemented immediately. We could support a jsonType in the spec, but throw a `not implemented` error for anything non-string, and work through a healthy design for matchers to implement later.

squeedee (approved):

emmjohnson:
@sclevine do you approve? Wanting to get as many maintainers approvals since we have no formal process (yet).

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+    app: web
+  resources:
+    - name: source-provider
+      templateRef:
+        kind: ClusterSourceTemplate
+        options:                                  # <--- a list
+        - name: git-template
+          selector:
+            matchFields:
+              - key: workload.spec.source.git       # <--- path to field in template context
```
jwntrs, line 74: This is no longer a "path to field in template context" it's a "path to field in the workload". Does that change our syntax here? I kinda like the idea of keeping the `workload` prefix just in case we do want to make other aspects of the system available later on.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+
+However, it is not certain that Cartographer will have the luxury of determinism. It is possible that the
+output of a choreographed resource may be important to the next step of a supply chain. Essentially, what if
+an object emits some values that can be handled by controller A, and other objects that can only be handled
+by controller B? There would need to be a step which could switch on the emitted values. This RFC does not
+handle this case. An RFC to use the entire templating context for template choice would address this use case.
+
+## Alternatives
+
+### Label selectors for choosing templates
```
cirocosta, line 133: I might be missing something, so please correct me if I'm wrong, but it feels like this approach could actually be a separate rule for `options`. e.g., one could make use of both proposals with something like

```yaml
  resources:
    - name: source-provider
      templateRef:
        kind: ClusterSourceTemplate
        options: 
        - name: git-template
          selector:
            matchFields:
              - key: workload.spec.source.git       # would match if the workload has this field filled in
                operator: Exists
        - name: imgpkg-bundle-template
          selector:
            matchLabels:   # or the label set below
              app.tanzu.vmware.com/source-type: imgpkg-bundle
```

although mixing like in the example I provided does seem like an odd approach to designing a supplychain imo

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
jwntrs: nvm I see this is called out explicitly in the bullet above

> key: a path on the workload object. The path must be prefixed by workload

rawlingsj:
LGTM - I'm not on the maintainers list but wanted to express my +1.  I like the approach and praise for finding a solution that keeps the current developer UX consistent, i.e. avoiding adding extra labels to the workload to help select the platform specific resources.

sclevine (approved):

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+
+When creating objects of the supply chain, Cartographer has the workload definition available.
+Allowing the supply chain to leverage the workload would add additional power to the conditional.
+
+Each step/resource in a supply chain will continue to define a templateRef, which will continue to
+specify a single template kind (e.g. ClusterImageTemplate). Rather than specifying the name of a
+particular template, authors can provide `options` which is a list of options. An option is
+a `name` and a `selector`. The only selector adopted by this RFC is `MatchFields` (though other
+selectors can be added later). A MatchFields contains:
+- `key`: a path on the workload object. The path must be prefixed by `workload`.
```
scothis, line 53: I didn't see anything describing how this path is defined, is it JSON Path?

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+would represent a code smell that this primitive is not adequate for Cartographer users.
+
+### Condition template choice on information available from the workload
+
+When creating objects of the supply chain, Cartographer has the workload definition available.
+Allowing the supply chain to leverage the workload would add additional power to the conditional.
+
+Each step/resource in a supply chain will continue to define a templateRef, which will continue to
+specify a single template kind (e.g. ClusterImageTemplate). Rather than specifying the name of a
+particular template, authors can provide `options` which is a list of options. An option is
```
scothis, line 50: Is the `options` field mutually exclusive with `name`, or is name removed and the the options list must be used even when there is a single choice?

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
```diff
@@ -1,3 +1,197 @@
...
+With no further effort, a platform can be created that has multiple supply chains with different
+purposes or requirements. By specifying characteristics of workloads in the workload labels,
+devs can provide information to operators necessary to choose the right supply chain for
+each workload.
+
+While this behavior is currently possible, relying only on this selection would lead to a platform
+with a large number of supply chains.
+
+Workarounds, for example templates that use ytt templating to embed conditional behavior,
+would represent a code smell that this primitive is not adequate for Cartographer users.
```
scothis, line 41: I fully expect use of ytt to manage conditional output to still be in very wide spread use. What this RFC is enabling is the ability to avoid a root level if/else where there is otherwise little commonality. In some cases a template may be able to shift away from using ytt, but I don't expect that to be common.

ytt remains the only meaningful way to:
- conditionally add/remove a field
- iterate over values
- coerce types
- apply logic
- do anything dynamic other than plain substitution

scothis (approved):

cirocosta (approved):

jwntrs (approved):

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
waciumawanjohi: Mutually exclusive. This will not present a breaking change.

rfc/rfc-0009-supply-chain-switches-template-on-flag.md
waciumawanjohi: ytt is useful. But there would be advantages to eliminating its use. [For example](https://github.com/vmware-tanzu/cartographer/pull/72#issuecomment-1028436294).

> ytt remains the only meaningful way to:

Enumerating these uses is valuable! We can begin to observe which are the most common and prioritize developing Cartographer native alternatives!

squeedee:
@waciumawanjohi do you mind merging this?

PR #77, opened 2021-09-07
Introduce rfc-0010-supply-chains-supply-default-params

waciumawanjohi:


rfc/rfc-0010-supply-chains-supply-default-params.md
```diff
@@ -0,0 +1,78 @@
...
+          params:
+            - name: some-param
+              value: $(workload.spec.params[?(@.name=="some-param")].value)$
+              default: "a great value"  # <--- new field
+        ```
+   2. The supply-chain can use the name-matching pattern that the templates use.
+      In the same way that the template assumes that the supply chain will have
+      a param with a matching name (or else the default value should be
+      substituted), the supply chain can assume the workload will have a matching
+      name param.
```
waciumawanjohi: I believe this approach is the easiest to implement. So much so that I have a working example on the following branch:
https://github.com/vmware-tanzu/cartographer/tree/test_default_params

To run the relevant feature tests:
`make test-kuttl focus=workload-does-not-provide-param`
and
`make test-kuttl focus=workload-provided-params`

Note that this approach is a breaking change to the supply chain API. (As a result, I expect most other tests on that branch are failing).

rfc/rfc-0010-supply-chains-supply-default-params.md
```diff
@@ -0,0 +1,78 @@
...
+        components:
+        - name: some-component
+          templateRef:
+            ...
+          params:
+            - name: some-param
+              value: $(workload.spec.params[?(@.name=="some-param")].value)$
+              default: "a great value"  # <--- new field
+        ```
+   2. The supply-chain can use the name-matching pattern that the templates use.
```
emmjohnson: So we will allow defaults for templates and supply chains. What is the order of precedence? 
- workload
- template
- supply chain?

For example, the supply chain is the primary default. The template can override the supply chain and the workload can override the template (or if no template default would override the supply chain).

rfc/rfc-0010-supply-chains-supply-default-params.md
```diff
@@ -0,0 +1,78 @@
...
+      apiVersion: carto.run/v1alpha1
+      kind: ClusterSupplyChain
+      spec:
+        components:
+        - name: some-component
+          templateRef:
+            ...
+          params:
+            - name: some-param
+              value: $(workload.spec.params[?(@.name=="some-param")].value)$
```
emmjohnson, line 174: I don't like having another field here. I think we can get away with just name and value. Where values can be over written by the precedence order defined/documented.

emmjohnson: I also think we don't need default anywhere, value would suffice.

idoru: what if we replaced value with something along the lines of like `prefer-workload-default: (true|false)`

rfc/rfc-0010-supply-chains-supply-default-params.md
waciumawanjohi: In this proposal:
Templates can define a param with a default. This implicitly allows the supply chain to overwrite it.
Supply chains can define a param with a default. This implicitly allows the workload to overwrite it.

This means that workloads has the highest precedence, but supply chains can decide not to define a default, allowing the template default to be used regardless of what is defined in the workload.

rfc/rfc-0010-supply-chains-supply-default-params.md
waciumawanjohi: > I think we can get away with just name and value.

On which object?

Are you suggesting that in every place that the keyword`default` is used, change that keyword to `value`?

DanieldeR:
Issue #156 would be fixed by addressing this RFC.

waciumawanjohi:
TODOs:
1. Change from "possible solutions" to solo solution 1. ii.
2. Add the ability for supply chain to define either a default field or a value field. Default can be overwritten by workload.
3. Workload must specify the component name and the param name for each component. This is because some templates may have name collisions for params. (e.g., template 1 wants to know 'disk space' and template 3 wants to know 'disk space' but these should be radically different values for different use cases).

sclevine:
I worry about parameterizing templates such that values from `workload.spec.params` end up in `params`. This creates an interface where the SupplyChain author can't easily swap out templates, because they're required to thread workload parameters through special, template-specific parameter names. Do we actually need to support jsonpath interpolation in SupplyChain at all? Maybe all values injected via the SupplyChain could be constant? Maybe templates could 

What if we:
1. Drop all interpolation in the SupplyChain itself (i.e., only support constant params)
2. Allow SupplyChain-provided params to be defaulted in templates (e.g., via `supplyChainParams:`)
3. Separately, allow Workload-provided params to be defaulted in templates (e.g., via `workloadParams:`)

> Workload must specify the component name and the param name for each component. 

Similarly, I worry about coupling Workloads and SupplyChains. Ideally, renaming a component in the SupplyChain shouldn't break Workloads.

rfc/rfc-0010-supply-chains-supply-default-params.md
```diff
@@ -0,0 +1,78 @@
...
+      kind: ClusterSourceTemplate
+      spec:
+        params:
+          - name: some-param             # <--- templates do not use jsonpath to specify supplychain location
+            default: "a fantastic value"
+      template:
+        apiVersion: carto.run/v1alpha1
+        kind: Deliverable
+        value:
+          first: $(params[?(@.name=="some-param")].value)$
```
sclevine: Could we flatten `params` into a map when it's made available in templates?

waciumawanjohi: Could we? Yes!
Should we? I think that it will be confusing for users to create entries in the params array in a template, and then refer to those entries with jsonpath for a map. The map notation is certainly simpler, it's just less tethered to any field defined on any object.

waciumawanjohi:
> I worry about parameterizing templates such that values from `workload.spec.params` end up in `params`. This creates an interface where the SupplyChain author can't easily swap out templates, because they're required to thread workload parameters through special, template-specific parameter names. Do we actually need to support jsonpath interpolation in SupplyChain at all? Maybe all values injected via the SupplyChain could be constant? Maybe templates could
> 
> What if we:
> 
> 1. Drop all interpolation in the SupplyChain itself (i.e., only support constant params)
> 2. Allow SupplyChain-provided params to be defaulted in templates (e.g., via `supplyChainParams:`)
> 3. Separately, allow Workload-provided params to be defaulted in templates (e.g., via `workloadParams:`)

#3 Sounds like it introduces the pattern you are concerned about. I also don't see why a template author should be empowered to decide what choices a dev can make that can't be overridden by the app operator.

The proposal we landed on will keep the app operator in control by allowing them, for every parameter, to either:

1. Specify nothing. In which case the default value in the template would apply. A matching param value specified by the workload would be ignored.
2. Specify a 'value' field. In which case that value would override the template's default. A matching param value specified by the workload would be ignored.
3. Specify a 'default' field. If the workload specifies a matching param, that value would override the supply chain and the template default. If the workload does not specify a matching param, the supply chain default would override the template value.

We certainly want to ease the app operator's ability to plug and play different templates. But I don't think we can get to a point where they can do so without some awareness of the inputs a template requires.

waciumawanjohi:
> > Workload must specify the component name and the param name for each component.
> 
> Similarly, I worry about coupling Workloads and SupplyChains. Ideally, renaming a component in the SupplyChain shouldn't break Workloads.

I agree, this is a pain point. But I think it is necessary pain. Imagine two templates that ask for a parameter like 'disk-space'. One is using 'disk space' to know how much space to allocate for a test, another is passing that value in to the image being built. (Very hand-wavy example, but I think it illustrates the problem)

Those two templates should not have to have knowledge of each other. And at the same time, the name collision must be disambiguated. The supply chain has been (I think rightly) the mediator between templates and workloads. So the supply chain should have some way of passing the right 'disk-space' value from the workload to template A and template B. (Note that in our current architecture, this problem could occur and would require the supply chain author to enforce that workloads specified two differently named fields in workload.spec.params.) I hear the concern that component name would ideally be some string that the supply chain author can alter at will. The alternatives are:

1. workload.spec.params are prepended with the template name
2. supply chains add another field to params such as workload-param-name. If that field is empty, we could assume that the workload param name in question is the same as the param name. example:

```yaml
kind: ClusterSupplyChain
metadata: ...
spec:
  components:
    - name: image-provider
      templateRef: ...
      params:
        - name: disk-space
          default: 10GB
          workload-param-name: disk-space-for-images
```

rfc/rfc-0010-supply-chains-supply-default-params.md
sclevine: Does the user really expect the jsonpath query to operate on the literal structure of the `params` field in `spec`? Maybe we could change the name of (or even remove) `params` to make this more clear?

It's a relatively common pattern to use a list of objects for parameters that are exposed as a map later (e.g., for env vars).

Also, I don't think that most users will be especially familiar with jsonpath syntax.

sclevine:
> I agree, this is a pain point. But I think it is necessary pain. Imagine two templates that ask for a parameter like 'disk-space'. One is using 'disk space' to know how much space to allocate for a test, another is passing that value in to the image being built. (Very hand-wavy example, but I think it illustrates the problem)

I think the solution to this is: template authors should understand that Workload parameters are global. They should each contribute to a unified Workload UX, not expose details of the SupplyChain in the Workload

sclevine:
>  I also don't see why a template author should be empowered to decide what choices a dev can make that can't be overridden by the app operator.

Hm, what if we:

1. Merge all `Workload` params into a single `params` map that's accessible in the template
2. Allow constant (non-jsonpath) parameter overrides in the SupplyChain (per-component **and** per-SupplyChain)
3. Allow parameter defaults in the templates

That would prevent parameters from being threaded through the SupplyChain, but allow operators to specify template parameters when needed.

jwntrs:
Just want to see if I can summarize some opinions here.

1. We should make params a map

2. Seems like we agree on the precedence of params:
workload
supply chain components
supply chain
template

3. Seems like we don't agree on an approach to address one of the original points brought up in the rfc, which is, that there will be certain params that the workload shouldn't be allowed to override. To this there are two approaches:

A) Use a `value`/`default` where appropriate. Where `defaults` can be overridden `values` cannot. This means a workload won't have a `default`, because it takes highest precedence. Template won't have `value` because it has the lowest precedence. Within the supply chain, params can have both `value` and `default`. 

When picking what value to use the ordering becomes:
```
supply chain component value
supply chain value
workload value
supply chain component default
supply chain default
template default
```

B) Use `value` everywhere, but supply chain params would have an extra field `overridable: (true|false)` that indicates whether a param is overridable by the workload. 

When picking what value to use the ordering becomes:
```
supply chain component value (overridable: false)
supply chain value (overridable: false)
workload value
supply chain component value (overridable: true)
supply chain value (overridable: true)
template value                                       <=== maybe this is the only place we keep default?
```

Update: changed `override: true` to `overrideable: true` as per https://github.com/vmware-tanzu/cartographer/pull/77#issuecomment-966662636

jwntrs:
Let's settle on an approach so we can move forward with an implementation. It looks like this will be a breaking change so we should try and implement sooner rather than later.

jwntrs:
I also think @waciumawanjohi made a good point about naming collisions in templates, but in the interest of moving this along I think we should push that decision out.

waciumawanjohi:
While I'm somewhat agnostic as to using `default|value` or using a flag, I will point out:
1. We already use the `default` noun, with no user complaints. That argues against changing templates to use a `value` field.
2. The template delegates upstream whether a parameter will be overwritten. (Currently by specifying a parameter `default`). The same pattern should apply to the supply chain, it should delegate to the workload. Using the verb `override` to mean that the supply chain overrides the workload is not delegation. Alternatively, a flag on the supply-chain value like `overridable` conceptually maps to the model of deciding whether or not to delegate to the workload. If a value were to have the `overridable` flag set to true, it would mean the workload parameter value would take precedence.

emmjohnson (approved):

PR #78, opened 2021-09-08
separate tools from application go.mod

cirocosta:
it's been great to keep track of the tools we need and their specific
versions by tracking them with `go.mod`, but it turns out that there are
tons of deps that get unecessarily introduced into our `go.mod` that
would be checked for, say, licenses if we were to scan out repository.

by breaking those go.mod apart, we can then proceed with scans solely on
the main `go.mod` and ignore the tools deps.

---

_thx @scothis for the inspiration 😁_


scothis:
I've recently taken to splitting out each tool into a separate module to avoid internal conflicts between tool dependencies. For example: (this may also be pure overkill 🤷)

https://github.com/k8s-service-bindings/service-binding-controller/blob/5b4514d7d3ae70809707913a180379364a611cde/Makefile#L9-L12

You can also setup Dependabot for the new module to automatically bump to the latest tool versions (that pass CI):

https://github.com/k8s-service-bindings/service-binding-controller/blob/5b4514d7d3ae70809707913a180379364a611cde/.github/dependabot.yml#L18-L29

cirocosta:
good point! totally forgot about updating dependabot 😅

> _I've recently taken to splitting out each tool into a separate module to avoid internal conflicts between tool dependencies._ 

that's neat, I'll leave a single go mod for the tools for now as it seems like there are no incompatibilities (although ... k8s deps being k8s deps .. I imagine will hit it at some point in the future, but then we cross that bridge at that time)

cirocosta:
(rebased)

cirocosta:
hmmm, actually, this doesn't work well with the `go:generate` annotations we have throughout the code

e.g.

```
pkg/registrar/registrar.go
17://go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
```

such that

```console
$ make generate
go generate ./...
no required module provides package github.com/maxbrunsfeld/counterfeiter/v6; to add it:
        go get github.com/maxbrunsfeld/counterfeiter/v6
pkg/conditions/condition_manager.go:17: running "go": exit status 1
no required module provides package github.com/maxbrunsfeld/counterfeiter/v6; to add it:
        go get github.com/maxbrunsfeld/counterfeiter/v6
...
```

---

i'll mark it as draft before getting `generate` properly working

scothis:
> ```
> pkg/registrar/registrar.go
> 17://go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
> ```

I assume you need to reference the go.mod in the hack dir, like `-mod=mod -modfile hack/go.mod`. Which raises an interesting question about the working directory this command is executed from...

cirocosta:
dropping this for now - if it turns out that those dependencies are indeed getting in the way, we can then come back to this

PR #79, opened 2021-09-08
Add unit tests for registrar functions

### LINKED ISSUE #19: Missing Tests: registrar.go (all of it)
There is no coverage for this file whatsoever :dagger: :D
https://github.com/vmware-tanzu/cartographer/blob/main/pkg/registrar/registrar.go

waciumawanjohi:
Moves anonymous function for indexing supply chains to
the supply chain definition file.

closes #19

### DIFF
--- a/pkg/apis/v1alpha1/cluster_supply_chain.go
+++ b/pkg/apis/v1alpha1/cluster_supply_chain.go
@@ -24,6 +24,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -120,6 +121,20 @@ func (c *ClusterSupplyChain) ValidateUpdate(_ runtime.Object) error {
 
 func (c *ClusterSupplyChain) ValidateDelete() error {
 	return nil
+}
+
+func GetSelectorsFromObject(o client.Object) []string {
+	var res []string
+	res = []string{}
+
+	sc, ok := o.(*ClusterSupplyChain)
+	if ok {
+		for key, value := range sc.Spec.Selector {
+			res = append(res, fmt.Sprintf("%s: %s", key, value))
+		}
+	}
+
+	return res
 }
 
 type SupplyChainSpec struct {
--- a/pkg/apis/v1alpha1/cluster_supply_chain_test.go
+++ b/pkg/apis/v1alpha1/cluster_supply_chain_test.go
@@ -429,4 +429,59 @@ var _ = Describe("ClusterSupplyChain", func() {
 		})
 
 	})
+
+	Describe("GetSelectorsFromObject", func() {
+		var expectedSelectors, actualSelectors []string
+		Context("when object is a supply chain", func() {
+			var sc *v1alpha1.ClusterSupplyChain
+
+			BeforeEach(func() {
+				sc = &v1alpha1.ClusterSupplyChain{Spec: v1alpha1.SupplyChainSpec{}}
+			})
+			Context("given a supply chain with 0 selectors", func() {
+				BeforeEach(func() {
+					actualSelectors = v1alpha1.GetSelectorsFromObject(sc)
+				})
+				It("returns an empty list", func() {
+					expectedSelectors = []string{}
+					Expect(actualSelectors).To(ConsistOf(expectedSelectors))
+				})
+			})
+			Context("given a supply chain with 1 selector", func() {
+				BeforeEach(func() {
+					sc.Spec.Selector = map[string]string{"some-key": "some-val"}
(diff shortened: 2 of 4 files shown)


pkg/apis/v1alpha1/cluster_supply_chain.go
```diff
@@ -122,6 +123,20 @@ func (c *ClusterSupplyChain) ValidateDelete() error {
...
+	var res []string
+	res = []string{}
+
+	sc, ok := o.(*ClusterSupplyChain)
+	if ok {
+		for key, value := range sc.Spec.Selector {
+			res = append(res, fmt.Sprintf("%s: %s", key, value))
+		}
+	}
+
```
cirocosta, line 136: it looks like we intend to never use this with something that is not a `*ClusterSupplyChain`, which makes me think that in case of `!ok`, we should even go to the "extremes" of `panic(fmt.Errorf("must be of type blabla"))` which is a feedback we'd get really quick in our integration / e2e tests

e.g.

```diff
diff --git a/pkg/apis/v1alpha1/cluster_supply_chain.go b/pkg/apis/v1alpha1/cluster_supply_chain.go
index cf60a20..ba71d43 100644
--- a/pkg/apis/v1alpha1/cluster_supply_chain.go
+++ b/pkg/apis/v1alpha1/cluster_supply_chain.go
@@ -127,6 +127,8 @@ func GetSelectorsFromObject(o client.Object) []string {
        var res []string
        res = []string{}

+       panic(fmt.Errorf("foo"))
+
        sc, ok := o.(*ClusterSupplyChain)
        if ok {
                for key, value := range sc.Spec.Selector {
```

leads to an almost immediate

```
        /home/cirocosta/go/pkg/mod/k8s.io/client-go@v0.22.1/tools/cache/shared_informer.go:410 +0x449
created by sigs.k8s.io/controller-runtime/pkg/cache/internal.(*specificInformersMap).Start.func1
        /home/cirocosta/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.0/pkg/cache/internal/informers_map.go:151 +0x111
panic: foo [recovered]
        panic: foo [recovered]
        panic: foo [recovered]
        panic: foo

goroutine 159 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc00010ec00})
        /home/cirocosta/go/pkg/mod/k8s.io/apimachinery@v0.22.1/pkg/util/runtime/runtime.go:55 +0xd8
panic({0x140f8a0, 0xc000788010})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc00046c0e0})
        /home/cirocosta/go/pkg/mod/k8s.io/apimachinery@v0.22.1/pkg/util/runtime/runtime.go:55 +0xd8
panic({0x140f8a0, 0xc000788010})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc000464be0})
        /home/cirocosta/go/pkg/mod/k8s.io/apimachinery@v0.22.1/pkg/util/runtime/runtime.go:55 +0xd8
panic({0x140f8a0, 0xc000788010})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1.GetSelectorsFromObject({0xc00020d080, 0xc00020d080})
```

(I know purposefully adding `panic()`s in the code can be something frowned upon by some, but this seems to be one of those cases where we have to adhere to an interface - the func signature that's expected - but we also want to guard against a programming mistake)

cirocosta: another option might be to actually be quite open: an indexer on `spec.selector` works on _anything_ that .. has a `spec.selector` - that could be done with something like https://pkg.go.dev/k8s.io/apimachinery@v0.21.1/pkg/apis/meta/v1/unstructured#NestedMap

cirocosta: (feel free to refuse both options, just a suggestion - leaving to your judgment 😁 )

cirocosta:
is this PR complete? e.g., registrar also takes care of registering the reconcilers with the manager so that the manager ... manages them, but I don't see tests there - did you decide to purposefully not include? (fine with me!)

thanks!

waciumawanjohi:
> is this PR complete? e.g., registrar also takes care of registering the reconcilers with the manager so that the manager ... manages them, but I don't see tests there - did you decide to purposefully not include? (fine with me!)
> 
> thanks!

PR is complete. Functions that were essentially stringing together calls to Controller Runtime seemed both difficult to test and of dubious value.

pkg/apis/v1alpha1/cluster_supply_chain.go
waciumawanjohi: I added the `ok` to the type check to avoid a panic (default behavior is that golang panics if the type check fails). My concern about a panic is that it would take down the whole controller. While I don't expect this function to be called by Controller Runtime with the wrong object type, if it were I want our code to be robust to that issue.

cirocosta (approved):

PR #97, opened 2021-09-10
fix: update test to reflect newly added resources

cirocosta:
Signed-off-by: Ciro S. Costa <ciroscosta@vmware.com>

### DIFF
--- a/pkg/registrar/registrar_test.go
+++ b/pkg/registrar/registrar_test.go
@@ -58,7 +58,7 @@ var _ = Describe("Registrar", func() {
 					Group:   "carto.run",
 					Version: "v1alpha1",
 				}
-				Expect(len(scheme.KnownTypes(gv))).To(Equal(21))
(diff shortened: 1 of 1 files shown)


emmjohnson:
Was it failing before this? Seems like we will have a hard time remembering to add them here?

emmjohnson (approved):

waciumawanjohi:
> Was it failing before this? Seems like we will have a hard time remembering to add them here?

Line 61 is the test to make sure that we'll remember to add them. (With a helpful comment on line 62)

waciumawanjohi (approved):

cirocosta:
> _Was it failing before this? _

it didn't appear as a fail in anyone's PR because of multiple PRs being merged together 

some projects use things like https://github.com/bors-ng/bors-ng to avoid that kind of thing, but imo, not very necessary with the number of PRs we have

PR #104, opened 2021-09-13
Issue 57/reject templates with namespace

### LINKED ISSUE #57: Cartographer rejects *Templates that attempt to set the namespace of the templated objects
_"I want to protect users from dangerous practices"_

```gherkin
Given a cluster with Cartographer
When a *Template that defines `spec.template.metadata.namespace` is submitted
Then the template is rejected
```

Examples:
This template would be rejected
```yaml
kind: ClusterBuildTemplate
metadata:
  name: kpack-template
spec:
  imagePath: .data.average_color_of_the_universe
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: test-configmap-build
      namespace: default # <--- the offending line
    data:
      average_color_of_the_universe: "Cosmic latte"

```

ekcasey:
Resolves #57

Adds validating webhook for
* `clusterconfigtemplates`
* `clusterimagetemplates`
* `clustersourcetemplates`
* `clustertemplates`

To test this behavior:
1. add `metadata.namespace` to the templates in the example `suppy-chain-templates.yaml`
2. apply the templates to the cluster
3. observe the following error messages
```
> k apply -f app-operator

clustersupplychain.carto.run/supply-chain created
Error from server (invalid template: template should not set metadata.namespace on the child object): error when creating "app-operator/supply-chain-templates.yaml": admission webhook "source-template-validator.cartographer.com" denied the request: invalid template: template should not set metadata.namespace on the child object
Error from server (invalid template: template should not set metadata.namespace on the child object): error when creating "app-operator/supply-chain-templates.yaml": admission webhook "image-template-validator.cartographer.com" denied the request: invalid template: template should not set metadata.namespace on the child object
Error from server (invalid template: template should not set metadata.namespace on the child object): error when creating "app-operator/supply-chain-templates.yaml": admission webhook "template-validator.cartographer.com" denied the request: invalid template: template should not set metadata.namespace on the child object
```

Questions:
1. Are we okay with the text of the error message?
2. I held off on adding other validations that seemed potentially sensible (e.g. does the stamped object have a kind?). Do we want more validations and if so, should we create another issue or add them to this PR?


config/webhook/webhook.yaml
```diff
@@ -33,6 +33,63 @@ webhooks:
         path: /validate-carto-run-v1alpha1-clustersupplychain
     sideEffects: None
     admissionReviewVersions: ["v1", "v1beta1"]
+  - name: config-template-validator.cartographer.com
```
ekcasey, line 36: I added these manually rather than using kubebuilder b/c it appeared like the supply chain validator was added manually? Would we prefer that these were autogenerated by kubebuilder?

pkg/apis/v1alpha1/cluster_config_template.go
```diff
@@ -36,14 +37,27 @@ type ClusterConfigTemplate struct {
 
 type ConfigTemplateSpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
-	Template   runtime.RawExtension `json:"template"`
-	ConfigPath string               `json:"configPath"`
-	Params     DefaultParams        `json:"params,omitempty"`
+	TemplateSpec `json:",inline"`
```
ekcasey, line 39: Happy to undo this refactor if folks don't like it, but it seemed convenient and also expressive of the intent - all templates share a core spec but may have extra fields specific to their type.

ekcasey:
~Looks like I missed a copyright header somewhere~, taking a look now.

Update: The issue was actually caused by a slight differenced in a generated file due to the older version of go I was using locally. Upgrading to go 1.17 fixed the failure in `make copyright` in the test workflow.

config/webhook/webhook.yaml
squeedee: That's on me, little experience with kubebuilder and I was dealing with a lot of conflicting information on the internet so I did what works. If there's a kubebuilder annotation then by all means let's use it!

pkg/apis/v1alpha1/cluster_config_template.go
squeedee: Now that they are sharing a core spec, perhaps it should have it's own file (so as not to imply a `parent->child` relationship between `cluster_template->cluster_[config|image|source]_template`?

squeedee (approved):

pkg/apis/v1alpha1/cluster_config_template.go
ekcasey: In the case of `ClusterTemplate`, the spec is simply `TemplateSpec`,  given that ClusterTemplate is the most basic possible template, with no added fields for exposing outputs (a similar thing is implied by the naming). So if we moved `TemplateSpec` into its own file, `cluster_template.go` would not contain the spec object for that template, which might be a little bit weird?

I think maybe I am trying to imply what you are suggesting we avoid implying - that ClusterXTemplate is a ClusterTemplate + X. But it is possible I am misunderstanding the concern.

squeedee: Yeah I think the underpinning here is that they're all siblings, and not 'subclasses' even if they look like it. But I see the point on naming which makes my minor concern seem too energetic :D

PR #108, opened 2021-09-14
Add ytt support for templates

scothis:
In addition to the current tempesting scheme using JSON Path,
Cartographer now support a sibling scheme that uses ytt. ytt has rich
support for applying data values to structured yaml with the full
expressiveness of a programming language that is executed within a
sandbox.

As a sibling to each Template field on the components is a Ytt field.
These fields are mutually exclusive. The ytt template is passed resolved
with the template context values bound as data values. For example, with
SupplyChain processing the Workload fields are available within ytt at
`data.values.workload`.

The ytt binary is added to the container by ko. At the moment only
linux-amd64 is supported.


scothis:
NOTE: the new code paths are not tested. Pushing up an early commit to make sure I didn't break the existing functionality.

pkg/apis/v1alpha1/common.go
```diff
@@ -28,6 +29,12 @@ type DefaultParam struct {
 	DefaultValue apiextensionsv1.JSON `json:"default"`
 }
 
+type ResourceTemplate struct {
```
waciumawanjohi: Big fan of pulling this out as a struct that can be inlined into the various *Templates.

scothis: folded it into the TemplateSpec that caused a fun little conflict

scothis:
Added unit tests

pkg/apis/v1alpha1/cluster_template.go
```diff
@@ -63,12 +64,20 @@ func (c *ClusterTemplate) ValidateDelete() error {
...
-	obj := metav1.PartialObjectMetadata{}
-	if err := json.Unmarshal(t.Template.Raw, &obj); err != nil {
-		return fmt.Errorf("invalid template: failed to parse object metadata: %w", err)
+	if t.Template == nil && t.Ytt == "" {
+		return fmt.Errorf("invalid template: must specify template or ytt, found neither")
 	}
-	if obj.Namespace != metav1.NamespaceNone {
-		return errors.New("invalid template: template should not set metadata.namespace on the child object")
+	if t.Template != nil && t.Ytt != "" {
+		return fmt.Errorf("invalid template: must specify template or ytt, found both")
```
squeedee: thoughts on "invalid template: must specify one of template or ytt, found both"?

scothis: added

pkg/templates/stamper.go
```diff
@@ -169,6 +173,79 @@ func (s *Stamper) Stamp(resourceTemplate []byte) (*unstructured.Unstructured, er
...
+	if kodata, ok := os.LookupEnv("KO_DATA_PATH"); ok {
+		ytt = path.Join(kodata, fmt.Sprintf("ytt-%s-%s", runtime.GOOS, runtime.GOARCH))
+	}
+
+	args := []string{"-f", "-"}
+	stdin := bytes.NewReader([]byte(template))
+	stdout := bytes.NewBuffer([]byte{})
+	stderr := bytes.NewBuffer([]byte{})
+
+	// TODO this would be cleaner if s.TemplatingContext was a known type
```
squeedee: Go 2.0 will have generics :P 

Jokes aside. It was a known type until Pipeline came along :/

scothis: Iterating over all the keys in the context end up being more robust

squeedee: can we nix the comment then?

scothis: done

squeedee (approved):
Looks great Scott.

Also thanks for leading the way with adding deeper logging, we intend to add more like it.

I would really like to see a serious kuttl example (not going to block this merge though). By serious I mean an example where loops or conditions are used (or both). 

Because `ytt` is intended as break glass, I assume we do not intend to use it to solve service templating requirements?

PR #112, opened 2021-09-15
Errors if an input references an unknown component

### LINKED ISSUE #3: Operator sees an error if they fail to specify an existing component
Recommend we update https://github.com/vmware-tanzu/cartographer/blob/main/tests/integration/supply_chain_validation_test.go#L18-337 to  include:
* When a component refers to an unspecified component.

# Acceptance 
[TBD]

Flows vmware-tanzu/cartographer-private#118

ekcasey:
* Also tweaks provider type mismatch error messages to be (hopefully) less confusing

Signed-off-by: Emily Casey <ecasey@vmware.com>

Resolves #3

### DIFF
--- a/pkg/apis/v1alpha1/cluster_supply_chain.go
+++ b/pkg/apis/v1alpha1/cluster_supply_chain.go
@@ -63,37 +63,49 @@ func (c *ClusterSupplyChain) validateNewState() error {
 	}
 
 	for _, component := range c.Spec.Components {
-		err := c.validateComponentRef(component.Sources, component, "ClusterSourceTemplate", "Source")
-		if err != nil {
-			return err
+		if err := c.validateComponentRefs(component.Sources, "ClusterSourceTemplate"); err != nil {
+			return fmt.Errorf(
+				"invalid sources for component '%s': %w",
+				component.Name,
+				err,
+			)
 		}
 
-		err = c.validateComponentRef(component.Images, component, "ClusterImageTemplate", "Image")
-		if err != nil {
-			return err
+		if err := c.validateComponentRefs(component.Images, "ClusterImageTemplate"); err != nil {
+			return fmt.Errorf(
+				"invalid images for component '%s': %w",
+				component.Name,
+				err,
+			)
 		}
 
-		err = c.validateComponentRef(component.Configs, component, "ClusterConfigTemplate", "Config")
-		if err != nil {
-			return err
+		if err := c.validateComponentRefs(component.Configs, "ClusterConfigTemplate"); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/apis/v1alpha1/cluster_supply_chain.go
```diff
@@ -63,37 +63,49 @@ func (c *ClusterSupplyChain) validateNewState() error {
...
+				"'%s' is provided by unknown component '%s'",
+				ref.Name,
+				ref.Component,
+			)
 		}
 		if referencedComponent.TemplateRef.Kind != targetKind {
 			return fmt.Errorf(
-				"%s '%s' for '%s' component must reference a %s",
-				sourceField,
+				"component '%s' providing '%s' must reference a %s",
```
ekcasey, line 106: Looking for feedback on this. Personally, I found that this change made the error messages more understandable, but I am curious what the maintainers think.

tests/integration/supply_chain/supply_chain_validation_test.go
```diff
@@ -273,7 +273,7 @@ var _ = Describe("SupplyChainValidation", func() {
 			It("Rejects the supply chain", func() {
 				err = c.Create(ctx, supplyChain)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring("Source 'provider' for 'build-image-provider' component must reference a ClusterSourceTemplate"))
+				Expect(err.Error()).To(ContainSubstring("component 'provider' providing 'solo-source-provider' must reference a ClusterSourceTemplate"))
```
ekcasey, line 276: I wonder if testing the exact text of the error messages twice is overkill. Maybe we should test the formatting in the unit tests and merely test that an error occurred in the integration tests?

pkg/apis/v1alpha1/cluster_supply_chain.go
squeedee: Looks good to me. We struggle with consistent log messages and there should be time carved out to review more of them. But I appreciate this.

emmjohnson (approved):

tests/integration/supply_chain/supply_chain_validation_test.go
squeedee: I prefer to include important user facing value in integrations (which i think of as features). The nature of the message is important to this story (we see debugging issues from user's that drove better messaging here). But more so, this is an outside-in testing habit and I'm not sure all the maintainers keep the same views as me. It's another topic that's due for review but timeliness is more important right now.

squeedee (approved):

PR #113, opened 2021-09-15
5 emit results

### LINKED ISSUE #5: Emit results on Pipeline CRD
```
As a supply chain author
I want to emit the latest run's _outputs_ 
So that I can trigger work on test results
```

* [x] copies results if available at build
* [x] copies results when available asynchronously
* [x] copies results when available asynchronously - resiliance - this should just *work* - need e2e test - issue created #114
* [x] should we iterate thru all outputs and return what we can instead of erroring and returning nil for outputs? - issue creatd #110 

## Notes
Do not worry about ‘success yet’

squeedee:
Work for #5 where the output changes of a stamped object are recorded on the pipeline.status


Makefile
```diff
@@ -50,7 +64,7 @@ coverage:
 	go tool cover -html=coverage.out -o coverage.html
 	open coverage.html
 
-lint:
+lint: copyright
```
idoru, line 67: ❤️

pkg/realizer/pipeline/realizer.go
```diff
@@ -42,18 +43,18 @@ type TemplatingContext struct {
 	Pipeline *v1alpha1.Pipeline `json:"pipeline"`
 }
 
-func (p *pipelineRealizer) Realize(pipeline *v1alpha1.Pipeline, logger logr.Logger, repository repository.Repository) *v1.Condition {
+func (p *pipelineRealizer) Realize(pipeline *v1alpha1.Pipeline, logger logr.Logger, repository repository.Repository) (*v1.Condition, templates.Outputs, *unstructured.Unstructured) {
 	pipeline.Spec.RunTemplateRef.Kind = "RunTemplate"
 	if pipeline.Spec.RunTemplateRef.Namespace == "" {
```
idoru, line 48: Missing some test coverage for the namespace defaulting here (although it's not new in this PR), just raising it if that's a concern.

idoru (approved):
Looks good to me. I can't help but feel `pipeline`'s `Realizer.Realize()` interface is starting to look a little bloated. Not sure if that's something we need to care about now.

PR #119, opened 2021-09-16
Reference sources, images, configs, params by name

idoru:


examples/source-to-knative-service/app-operator/supply-chain-templates.yaml
```diff
@@ -137,7 +137,7 @@ spec:
                     spec:
                       containers:
                         - name: workload
-                          image: $(images[0].image)$
+                          image: $(images.image.image)$
```
cirocosta, line 140: 😅  maybe we could update the supplychain.yaml to be a little nicer, like, `images.app.image` 

might be worth doing the same for `sources`, so we do like, `sources.app.url` ..

cirocosta:
reminded me of https://concourse-ci.org/jobs.html#schema.step.task-step.input_mapping

> A map from task input names to concrete names in the build plan. This allows a task with generic input names to be used multiple times in the same plan, mapping its inputs to specific resources within the plan.
>
> ---
> The following example demonstrates a task with a generic release-repo input being mapped to more specific artifact names:
> 
> ```
> plan:
> - get: diego-release
> - get: cf-release
> - get: ci-scripts
> - task: audit-diego-release
>   file: ci-scripts/audit-release.yml
>   input_mapping: {release-repo: diego-release}
> - task: audit-cf-release
>   file: ci-scripts/audit-release.yml
>   input_mapping: {release-repo: cf-release}
> ```

(not saying we should implement, but I could see that being requested at some point)

cirocosta:
made me wonder if we should be more restrictive with regards to the alphabet allowed for the `name` 🤔 e.g., to avoid `name: "foo bar"`, `name: "foo\u200bbar", ..., similar to how `ConfigMap`s perform validation on their keys

applying the perfectly-valid yaml 

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo
data:
  "\u200b": foo
```

yields 

```
The ConfigMap "foo" is invalid: data[​]: Invalid value: "\u200b": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

site/content/docs/reference.md
```diff
@@ -164,16 +164,16 @@ spec:
       # 
       # in a template, these can be consumed as: 
       #
-      #    $(sources[<idx>]url)$
-      #    $(sources[<idx>]revision)$
+      #    $(sources.<name>.url)$
```
cirocosta, line 167: being _that_ one: with the introduction of `source.(url|revision)` etc, we should update the docs

site/content/docs/reference.md
```diff
@@ -312,7 +315,7 @@ spec:
         name: java-builder
       source:
         blob:
-          url: $(sources[0].url)$
+          url: $(sources.provider.url)$
 ```
 
 _ref: [pkg/apis/v1alpha1/cluster_image_template.go](https://github.com/vmware-tanzu/cartographer/blob/v0.0.3/pkg/apis/v1alpha1/cluster_image_template.go)_
```
cirocosta, line 334: unrelated (might be worth a separate issue/pr): we should update these references 😅 it was very useful when the docs repo was separate (so we could ensure no drift would occur), but being in a single repo, I think it shouldn't reference a particular revision (unless we're referring to how something _was_ rather than _is_)

cirocosta:
given that this is a breaking change for those already authoring templates, wdyt of starting a `CHANGELOG.md` where we could record major changes like this?

cirocosta:
@idoru @emmjohnson looks like a few conflicts popped up

Screen Shot 2021-09-20 at 10 31 35 AM

examples/source-to-knative-service/app-operator/supply-chain-templates.yaml
emmjohnson: Will create issue for this

cirocosta (approved):
lgtm 😁

Issue #99, opened 2021-09-10
GitWritter with Tekton Example


DanieldeR:
A current gap that exists is the ability to write to a git repository. The purpose of this issue is to create an example supply chain with the last step leveraging the pipeline service and Tekton to write some YAML to a git repository.

## Acceptance Criteria
An example exists where:
* A supply chain writes some yaml (likely a knative service wrapped in a Kapp App) to a git repository
* Usage of the supply chain is documented in both OSS and Commericial docs

**Note** While we may eventually need the ability to specify the git repository in the workload, for now having the operator hard code it in a Tekton pipeline is sufficient.

squeedee:
Leaving open as it gets rolled into #405

PR #120, opened 2021-09-16
Remove pipeline experiment as it is now part of the code

emmjohnson:


cirocosta (approved):

cirocosta:
I just realized that this will actually end up also getting rid of all the examples etc - I think it's worth the whole removal anyway, but just a reminder for anyone reading this that it _might_ be useful in the future (when tackling https://github.com/vmware-tanzu/cartographer/issues/99) to come back  to previous revisions and pick up the examples that were there already.

(still lgtm 😁 )

PR #133, opened 2021-09-21
Move deliverable into test only path

### LINKED ISSUE #130: remove deliverable from crd, move it so it is available for tests

emmjohnson:
Closes #130 


pkg/registrar/registrar_test.go
```diff
@@ -58,7 +58,7 @@ var _ = Describe("Registrar", func() {
 					Group:   "carto.run",
 					Version: "v1alpha1",
 				}
-				Expect(len(scheme.KnownTypes(gv))).To(Equal(25))
+				Expect(len(scheme.KnownTypes(gv))).To(Equal(23))
```
cirocosta, line 61: at first I thought 'oh, 23? shouldn't it be 24?', but I guess the deduction of `2` is due to `*List` being registered 👍

cirocosta:
dang, I just realized that we actually have `Deliverable` mentions in our `site`:

```
site/content/docs/install.md
47:^                    deliverables.carto.run             CustomResourceDefinition  -       -    create  -       reconcile  -   -

site/content/docs/about.md
14:* Deliverable ([Spec Reference](reference.md#deliverable))
16:The supply chain allows users to define the steps that an application must pass through in order to create an image. The deliverable portion of a Cartographer workflow allows the user to define the steps that the built image must pass through. By combining supply chains with delivery, the entire path to production can be specified and thereby choreographed by Cartographer.
20:Both the supply chain and delivery consist of components that are specified via Templates. Each template acts as a wrapper for existing Kubernetes resources and allows them to be used with Cartographer. There are currently four different types of templates that can be use in a Cartographer supply chain or deliverable:

site/content/docs/reference.md
143:  # deliverable state. (required, at least 1)

```

I think it'd be good to make sure we get rid of that

kuttl-test.yaml
```diff
@@ -18,6 +18,8 @@ startControlPlane: true
 testDirs:
   - tests/kuttl
 crdDir: config/crd/bases
+manifestDirs:
+  - tests/crd
```
cirocosta, line 22: so, my understanding is that we're doing a plain move of the CustomResourceDefinition object that has been generated before to that directory, and then just submitting to Kubernetes when running `kuttl`, but there's no go-based definition of the `struct` that `controller-gen` would generate, right?

if that's the case, I'd say it'd be good if we could strip down the CustomResourceDefinition to a minimum, removing those descriptions like

```
        openAPIV3Schema:
          properties:
            apiVersion:
              description: 'APIVersion defines the versioned schema of this representation
                of an object. Servers should convert recognized schemas to the latest
                internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
              type: string
            kind:
              description: 'Kind is a string value representing the REST resource this
                object represents. Servers may infer this from the endpoint the client
                submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-       kinds'
              type: string
```

and the annotation that indicates that this was generated:

```
  metadata:
    annotations:
      controller-gen.kubebuilder.io/version: v0.6.2
```

cirocosta:
oh btw, I just remembered that for the `Test` CRD used in integration we have

```
          SchemeGroupVersion = schema.GroupVersion{
                  Group:   "test.run",
                  Version: "v1alpha1",
          }
```

which imo is nice that allows one to quickly tell that such resource is not part of the main set of resources.

I think either here or in another PR we should move this object to such apiversion - wdyt?

kuttl-test.yaml
emmjohnson: Not going to take the time to do this. I think we have some work to do around both "test" crds at the moment.

squeedee (approved):
Good enough for beta!

PR #134, opened 2021-09-21
81 pipeline params

### LINKED ISSUE #81: Pipeline Service should be able to handle params
As an operator, I want to be able to write values into the pipeline which are picked up by the runTemplate.

AC
```gherkin
Given a pipeline, CRD
When a value is written into the inputs field
And the runTemplate refers to that field
Then the object is stamped out with value in that field defined.
```

```yaml
kind: Pipeline
metadata:
  name: tests
spec:
  runTemplateName: tekton-runner # use the `RunTemplate` named ‘tekton-runner’ to
                                 # figure out how to stamp out “*Run” objects that
                                 # represents “running” (e.g. tekton/PipelineRun)

  inputs:                        # at the moment, you can pass anything to this field,
    url: http://git.dev/d3adb33f.tgz     # but it could (should?) be something
                                         # strongly typed

---

kind: RunTemplate
metadata:
  name: run-with-source
  labels:
    app.tanzu.vmware.com/type: test 
spec:
  completion:
    succeeded: {key: 'status.conditions.#(type=="Succeeded").status', value: "True"}
    Failed: {key: 'status.conditions.#(type=="Succeeded").status', valueNot: "True"}

  template:
    pipelineRef: 
      name: my-tekton-pipeline                   # <= (1)
    workspaces:
      - name: shared-data
        volumeClaimTemplate:
          spec:
            accessModes:
              - ReadWriteOnce
            resources:
              requests:
                storage: 5Gi
    params:
      - name: blob-url
        value: '$(pipeline.spec.inputs.url)'    # <= (2)
```

would result in an object

```yaml
spec:
  workspaces:
    - name: shared-data
      volumeClaimTemplate:
        spec:
          accessModes:
            - ReadWriteOnce
          resources:
            requests:
              storage: 5Gi
  params:
    - name: blob-url
      value: http://git.dev/d3adb33f.tgz
```

# Notes:
The top level field should be called `Inputs`. This field should be a map[string]JSONobject.

# Out of scope
Default values

waciumawanjohi:


waciumawanjohi:
Work should be squashed before merging.

emmjohnson (approved):

cirocosta:
```
35a2020dc397539a2a9c9b4e2d4e69810fc7a882 hack: turn docker-login resilient to bad json
```

addresses the flakiness to e2e setup

cirocosta:
(pushed a rebase)

PR #136, opened 2021-09-21
Use local refs for docs

emmjohnson:


cirocosta (approved):
niice, looks good! went one by one, all working good 👍

cirocosta:
it'd be good to include an explanation for `ytt` too though (separate PR imo) - one doing `k explain clustersourcetemplate` can tell that the field exists but wouldn't figure out how / why to use it

PR #137, opened 2021-09-21
4 Emit only on success

### LINKED ISSUE #4: Only emit results on success
As a pipeline author
I want the status of my Pipeline to be based on the _most recently submitted_ and _successful_ PipelineRun 
So that I can trigger work only when a test passes

# Basic Spec
```gherkin
Given a pipeline and runTemplate that have created an object
When that object has a Succeeded condition that is true
Then the Pipeline's Status includes the output fields defined on the runTemplate
```

```gherkin
Given a pipeline and runTemplate that has created an object
When that object has a Succeeded condition that is not true
Then the Pipeline's Status is still that of a previous successful run
```

# Notes:
## Hard coding 'success'
`Success` has been scoped to:
When `stampedObject.Status.Conditions[@type='succeeded'].Status` == true 

## Compare based on creationTimestamp
1. We expect one pipeline can kick off multiple stampedobjects at once.
2. Those objects can succeed in indeterminate order.
3. We expect to emit status from the most recently submitted object which is successful.
4. In this story, most recently successful will be judged by their creationTimestamp.

# Out of scope:
## Ordering objects defined more than once
This scheme has a corner case for when a pipeline has spec A (creates an object), is updated to spec B (creates another object), and then is updated to spec A (will not always create a new object because the original exists). In that case, when both A and B are successful, the most recently submitted successful object should be A (it is so recent that it is the current spec of the pipeline). The current story need not handle this corner case.

## Time collisions
One could imagine a scenario where 1 pipeline submitted two objects in such close succession that they have the same creationTimestamp. In that scenario, it would be impossible to judge which is the most recently submitted. This story need not handle that case. There may be a follow-on story to create some mechanism to move from timestamps to some monotonically increasing series. This series would need to be appended to the objects created by the pipeline. Story authors should remember that new object creation can be triggered by changes to the pipeline, the runTemplate or (possibly) some 3rd object to which they refer.

waciumawanjohi:
Pipeline will emit an output of the templated object that was most recently submitted to the APIServer and which has a condition type=Succeeded and status=True.

### DIFF
--- a/pkg/realizer/pipeline/realizer.go
+++ b/pkg/realizer/pipeline/realizer.go
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
 		return TemplateStampFailureCondition(fmt.Errorf("%s: %w", errorMessage, err)), nil, nil
 	}
 
-	err = repository.EnsureObjectExistsOnCluster(stampedObject, false)
+	err = repository.EnsureObjectExistsOnCluster(stampedObject.DeepCopy(), false)
 	if err != nil {
 		errorMessage := "could not create object"
 		logger.Error(err, errorMessage)
 		return StampedObjectRejectedByAPIServerCondition(fmt.Errorf("%s: %w", errorMessage, err)), nil, nil
 	}
 
-	outputs, err := template.GetOutput(stampedObject)
+	stampedObject.SetLabels(labels)
+	allPipelineStampedObjects, err := repository.ListUnstructured(stampedObject)
+	if err != nil {
+		errorMessage := fmt.Sprintf("could not list pipeline objects: %s", err.Error())
+		logger.Info(errorMessage)
+		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
+	}
+	// TODO: handle the case where the pipeline has changed what type of object it stamps
+	// TODO: handle the case where the pipeline has changed what runTemplate pairs with
+
+	outputs, err := template.GetOutput(pipeline.Status.Outputs, allPipelineStampedObjects)
 
 	if err != nil {
 		errorMessage := fmt.Sprintf("could not get output: %s", err.Error())
--- a/pkg/realizer/pipeline/realizer.go
+++ b/pkg/realizer/pipeline/realizer.go
@@ -87,22 +87,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
 		return StampedObjectRejectedByAPIServerCondition(fmt.Errorf("%s: %w", errorMessage, err)), nil, nil
 	}
 
-	stampedObject.SetLabels(labels)
-	allPipelineStampedObjects, err := repository.ListUnstructured(stampedObject)
+	objectForListCall := stampedObject.DeepCopy()
+	objectForListCall.SetLabels(labels)
+
+	allPipelineStampedObjects, err := repository.ListUnstructured(objectForListCall)
 	if err != nil {
-		errorMessage := fmt.Sprintf("could not list pipeline objects: %s", err.Error())
-		logger.Info(errorMessage)
-		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
+		err := fmt.Errorf("could not list pipeline objects: %w", err)
+		logger.Info(err.Error())
+		return FailedToListCreatedObjectsCondition(err), nil, stampedObject
 	}
-	// TODO: handle the case where the pipeline has changed what type of object it stamps
-	// TODO: handle the case where the pipeline has changed what runTemplate pairs with
 
-	outputs, err := template.GetOutput(pipeline.Status.Outputs, allPipelineStampedObjects)
-
+	outputs, err := template.GetOutput(allPipelineStampedObjects)
 	if err != nil {
 		errorMessage := fmt.Sprintf("could not get output: %s", err.Error())
 		logger.Info(errorMessage)
 		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
+	}
+	if len(outputs) == 0 {
+		outputs = pipeline.Status.Outputs
 	}
 
 	return RunTemplateReadyCondition(), outputs, stampedObject
--- a/pkg/templates/run_template.go
+++ b/pkg/templates/run_template.go
@@ -31,23 +31,24 @@ type Outputs map[string]apiextensionsv1.JSON
 type RunTemplate interface {
 	GetName() string
 	GetResourceTemplate() v1alpha1.TemplateSpec
-	GetOutput(outputs map[string]apiextensionsv1.JSON, stampedObjects []*unstructured.Unstructured) (Outputs, error)
+	GetOutput(stampedObjects []*unstructured.Unstructured) (Outputs, error)
 }
 
 type runTemplate struct {
 	template *v1alpha1.RunTemplate
 }
 
-func (t runTemplate) GetOutput(outputs map[string]apiextensionsv1.JSON, stampedObjects []*unstructured.Unstructured) (Outputs, error) {
+func (t runTemplate) GetOutput(stampedObjects []*unstructured.Unstructured) (Outputs, error) {
 	var (
 		updateError                        error
-		everyObjectErrored, outputUpdated  bool
+		everyObjectErrored                 bool
 		mostRecentlySubmittedSuccesfulTime *time.Time
 	)
 
+	outputs := Outputs{}
+
 	evaluator := eval.EvaluatorBuilder()
 
-	outputUpdated = false
 	everyObjectErrored = true
 
 	for _, stampedObject := range stampedObjects {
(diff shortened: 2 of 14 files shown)


tests/integration/pipeline_service/pipeline_service_test.go
```diff
@@ -349,42 +456,60 @@ var _ = Describe("Stamping a resource on Pipeline Creation", func() {
...
-			err = yaml.Unmarshal([]byte(pipelineYaml), pipelineDefinition)
+			testsList := &testapi.TestList{}
+
+			Eventually(func() ([]testapi.Test, error) {
+				err := c.List(ctx, testsList, opts...)
+				return testsList.Items, err
+			}).Should(HaveLen(1))
+
+			// This is in order to ensure gen 1 object and gen 2 object have different creationTimestamps
+			time.Sleep(time.Second)
```
emmjohnson, line 488: Is a whole second necessary here? Can we sleep less? (same for second sleep below)

pkg/realizer/pipeline/realizer.go
```diff
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
...
 
-	outputs, err := template.GetOutput(stampedObject)
+	stampedObject.SetLabels(labels)
+	allPipelineStampedObjects, err := repository.ListUnstructured(stampedObject)
+	if err != nil {
+		errorMessage := fmt.Sprintf("could not list pipeline objects: %s", err.Error())
+		logger.Info(errorMessage)
+		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
+	}
+	// TODO: handle the case where the pipeline has changed what type of object it stamps
```
emmjohnson: Are there issues created for these?

pkg/realizer/pipeline/realizer.go
```diff
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
...
-	err = repository.EnsureObjectExistsOnCluster(stampedObject, false)
+	err = repository.EnsureObjectExistsOnCluster(stampedObject.DeepCopy(), false)
 	if err != nil {
 		errorMessage := "could not create object"
 		logger.Error(err, errorMessage)
 		return StampedObjectRejectedByAPIServerCondition(fmt.Errorf("%s: %w", errorMessage, err)), nil, nil
 	}
 
-	outputs, err := template.GetOutput(stampedObject)
+	stampedObject.SetLabels(labels)
```
emmjohnson: SetLabels is called from within the stampContext.Stamp() - this call should not be necessary

pkg/realizer/pipeline/realizer.go
```diff
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
...
+	allPipelineStampedObjects, err := repository.ListUnstructured(stampedObject)
+	if err != nil {
+		errorMessage := fmt.Sprintf("could not list pipeline objects: %s", err.Error())
+		logger.Info(errorMessage)
+		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
```
emmjohnson: This condition does not make sense in this context

pkg/realizer/pipeline/realizer.go
```diff
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
...
+	err = repository.EnsureObjectExistsOnCluster(stampedObject.DeepCopy(), false)
 	if err != nil {
 		errorMessage := "could not create object"
 		logger.Error(err, errorMessage)
 		return StampedObjectRejectedByAPIServerCondition(fmt.Errorf("%s: %w", errorMessage, err)), nil, nil
 	}
 
-	outputs, err := template.GetOutput(stampedObject)
+	stampedObject.SetLabels(labels)
+	allPipelineStampedObjects, err := repository.ListUnstructured(stampedObject)
```
emmjohnson: @squeedee and I are thinking that this could be a pain point in the near future. We assume there will be 1000s of objects that get returned here. Can we optimize this? This should be it's own exploration/story.

pkg/templates/run_template.go
```diff
@@ -30,35 +31,103 @@ type Outputs map[string]apiextensionsv1.JSON
...
+			}
+
+			outputs = provisionalOutputs
+			outputUpdated = true
+		}
+
+		if objectErr != nil {
+			updateError = objectErr
+		} else {
+			everyObjectErrored = false
```
emmjohnson, line 84: everyObjectUnsuccessful might be more meaningful

pkg/realizer/pipeline/realizer.go
```diff
@@ -80,14 +80,24 @@ func (p *pipelineRealizer) Realize(ctx context.Context, pipeline *v1alpha1.Pipel
...
+	if err != nil {
+		errorMessage := fmt.Sprintf("could not list pipeline objects: %s", err.Error())
+		logger.Info(errorMessage)
+		return OutputPathNotSatisfiedCondition(err), nil, stampedObject
+	}
+	// TODO: handle the case where the pipeline has changed what type of object it stamps
+	// TODO: handle the case where the pipeline has changed what runTemplate pairs with
+
+	outputs, err := template.GetOutput(pipeline.Status.Outputs, allPipelineStampedObjects)
```
emmjohnson: We don't use pipeline.Status.Outputs anywhere in GetOutput. Let's not pass it in and instead on return check if outputs is empty. If outputs is empty set to pipeline.Status.Outputs.

```
	outputs, err := template.GetOutput(allPipelineStampedObjects)
	if err != nil {
		//
	}
	if len(outputs) == 0 {
		outputs = pipeline.Status.Outputs
	}
```

emmjohnson (changes requested):
See comments made

tests/integration/pipeline_service/pipeline_service_test.go
waciumawanjohi: We need to sleep for long enough for the creationTimestamp to be different for these three objects. Omitting the sleep led to failures.

Here's the output of doing a `k get -o yaml` on an object: `creationTimestamp: "2021-09-17T16:02:30Z"` No subsecond precision is reported.

Suggests that yes, we need to have an entire second in order to get different timestamps.

pkg/realizer/pipeline/realizer.go
waciumawanjohi: Additional labels can be set on the stamped object (for example, any label defined in the runTemplate's template). We only want to use the four labels defined in the labels variable.

pkg/templates/run_template.go
waciumawanjohi: I wonder if "unsuccessful" will get mixed in with the concept of the Succeeded condition. If for every object, the succeeded condition is false, but the variable everyObjectUnsuccessful is false, would that be surprising?

pkg/realizer/pipeline/realizer.go
waciumawanjohi: 👍

cirocosta:
@waciumawanjohi please make sure you rebase so the branch gets the changes to improve ci flakiness 😁

pkg/realizer/pipeline/realizer.go
waciumawanjohi: 👍

tests/integration/pipeline_service/pipeline_service_test.go
squeedee: Oh geeze that worries me on a deeper 'co-incidental timing' level. The design we discussed of a "sequential counter of our own" on the pipeline (referencing each 'cause to stamp' incrementally) makes even more sense in light of this.

pkg/realizer/pipeline/realizer.go
squeedee: Ahhh this makes sense now. This line only impacts the in memory object, which is used for selecting other objects to gain status from. It is also returned to the caller, which I understand does not impact it's upstream use (as a source for an informer) but it would still surprise me that it's a mutation.

Could we deepcopy it as `listingStampedObject` or something, so we don't mutate the return value?

pkg/templates/run_template.go
squeedee: `noSuccessesFound` perhaps? The intent is that you're searching for a success, correct?

pkg/realizer/pipeline/realizer.go
waciumawanjohi: #142 and #143

pkg/realizer/pipeline/realizer.go
waciumawanjohi: See issue #144

pkg/templates/run_template.go
emmjohnson: I just don't like Error - it is confusing me with a client error. Approving

emmjohnson (approved):

PR #141, opened 2021-09-22
address docker-login.sh flakiness

cirocosta:
Signed-off-by: Ciro S. Costa <ciroscosta@vmware.com>

### DIFF
--- a/hack/docker-login.sh
+++ b/hack/docker-login.sh
@@ -35,7 +35,11 @@ update_or_create_docker_config() {
 
         readonly qry='.auths["%s"] = {"auth": "%s"}'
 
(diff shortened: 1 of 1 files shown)


emmjohnson (approved):

hack/docker-login.sh
```diff
@@ -35,7 +35,11 @@ update_or_create_docker_config() {
 
         readonly qry='.auths["%s"] = {"auth": "%s"}'
 
-        test -s $config_file || echo '{}' >$config_file
+        local current_content=$(jq '.' $config_file)
```
cirocosta, line 38: this way we can ensure that we're always dealing with a proper json file there, catching errors like file not existing or not being a proper json

PR #146, opened 2021-09-23
Caret versions on go in validation and releaser, and allow overriding versions in setup-binaries

tbr11:
- Go 1.x is compatible so using caret versions will pick any 1.x version, should reduce cycles on releaser workflow
- Overriding versions in setup-binaries lets us rebuild a particular sha even if a dependency version gets pulled or otherwise needs to be updated


cirocosta (approved):
👍 lgtm!

PR #151, opened 2021-09-23
github: add issue template

cirocosta:
this is a straight port of what we had before in the gitlab repository.

hitting "new issue" will bring:

Screen Shot 2021-09-23 at 2 04 55 PM

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,17 @@
+_"I want to <x>"_
+
+```gherkin
+Given <Some Condition>
+When <Something Happens>
+Then <This other thing should happen?
+```
+
+
+### Example
+
+<Code snippets that illustrate the when/then blocks>
+
+
+## Additional Context
+
+<Additional notes>


tbr11 (approved):

idoru:
I like the motivation to do this; but this feels a lot like a feature request template.

I'd love to see us have a specific template for both features and bug reports.

Considering the template content itself, if this is to be a feature request template, I'd like to get input on the problem as well as the proposed solution. To that end, what do you think about prepending what you have with something like this:

```
### Problem description

<Please describe the problem you're trying to solve>

### Proposed solution
```

In terms of a bug report, I'd expect something more like:

```
### Bug description:
A clear and concise description of what the bug is.

### To reproduce:
Steps to reproduce the behavior (example):
1. Apply attached supply chains and workloads
2. Observe status of Workload

### Expected behavior:
A clear and concise description of what you expected to happen.

### Actual behavior:
A clear and concise description of what actually happened.

### Logs, k8s object dumps:
Logs or k8s object dumps are always helpful! Add logs to help explain your problem. Dump objects in yaml `-o yaml` format.

### Versions:
 - Infrastructure (kind, TKG, GKE etc)
 - cartographer version
 - k8s version
 - ... other versions of related software being used

### Deployment info:
If possible, share your (redacted) manifests used to deploy.

If you used any deployment strategy it'd be helpful to point it out and share as
much about it as possible (e.g.  `kubectl apply` vs `kapp deploy`)

### Additional context:
Add any other context about the problem here.
```

Flying by the seat of my pants on the `Logs/Versions/Deployment info` section so I'm sure it's a bit wrong/missing stuff that would be more useful, but I just wanted to try moving the idea in this direction.

WDYT?

cirocosta:
@idoru +1 on multiple templates - I think we can achieve with [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) - e.g., see [concourse's ISSUE_TEMPLATE dir](https://github.com/concourse/concourse/tree/master/.github/ISSUE_TEMPLATE)

Screen Shot 2021-09-27 at 12 29 42 PM


I'm personally not a huge fan of GitHub discussions (which is used under `concourse/concourse` for general questions / submitting feature request), but we could even mix those.

how about starting a new issue for tracking these improvements? btw, if you feel strongly about reverting this PR, happy to accept a PR reverting it.

thxx!

PR #164, opened 2021-09-27
Housekeeping

squeedee:
Just 2 quick houskeeping steps

1. Introduce PHONIES to avoid future confusion with our Makefiles
2. Lint the hack scripts (to the level of error) with shellcheck.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,18 @@
+.PHONY: build
 build: gen-objects gen-manifests
 	go build -o build/cartographer ./cmd/cartographer
+
+.PHONY: run
 run: build
 	build/cartographer
 
+.PHONY: gen-objects
 gen-objects:
 	go run sigs.k8s.io/controller-tools/cmd/controller-gen \
                 object \
                 paths=./pkg/apis/v1alpha1
+
+.PHONY: gen-manifests
 gen-manifests:
 	go run sigs.k8s.io/controller-tools/cmd/controller-gen \
 		crd \
@@ -16,11 +22,13 @@ gen-manifests:
 		-f ./hack/boilerplate.go.txt \
 		config/crd/bases
 
+.PHONY: test-gen-objects
 test-gen-objects:
 	go run sigs.k8s.io/controller-tools/cmd/controller-gen \
                 object \
                 paths=./tests/integration/pipeline_service/testapi
 
+.PHONY: test-gen-manifests
 test-gen-manifests:
 	go run sigs.k8s.io/controller-tools/cmd/controller-gen \
 		crd \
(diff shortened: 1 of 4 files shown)


hack/setup.sh
```diff
@@ -114,7 +114,7 @@ show_usage_help() {
 display_vars() {
         echo "Variables:
 
-	COMMANDS: 	$@
+	COMMANDS: 	$*
```
cirocosta, line 117: is this from shellsheck? IIUC, $@ and $* don't really matter in this particular case

hack/Makefile
```diff
@@ -0,0 +1,4 @@
+.PHONY: lint
+
+lint:
+	shellcheck *.sh -S error # eventually we want to lift the severity and catch warnings too.
```
cirocosta, line 4: I think we should update `./CONTRIBUTING.md` - there we mention all the dev dependencies needed

Makefile
```diff
@@ -1,12 +1,18 @@
+.PHONY: build
```
cirocosta, line 1: I think `phony'`ing `build `is indeed necessary, but, what'd be the reason for the others? I'm not suuuper familiar with `make` details like this, but from what I understand, as the other targets don't produce anything that matches a target, we'd not have nay executions being skipped

cirocosta (approved):
lgtm  👍  just left a few questions

hack/Makefile
cirocosta: hmmm I'm surprised this didn't break `ci` 🤔  https://github.com/vmware-tanzu/cartographer/blob/cce05c1d5853deef994635a62eb4d2b11d7fbac1/.github/workflows/validation.yaml#L46

I guess that comes installed  by default on their ubuntu setup?

https://github.com/vmware-tanzu/cartographer/runs/3723206626#step:5:247

hack/setup.sh
squeedee: Yeah it was from shellcheck

hack/Makefile
squeedee: That's a good point. I guess it must?

Makefile
squeedee: The reason is future proofing. The day you create a file that matches a Make target is the day you spend hours trying to work out why something isn't triggering.

squeedee: By way of example, you accidentally create a file called test. This is common as in `touch foo` or `vim test`

Then you finish your work, add the files you care about `make test && git push`

cirocosta: got it, makes sense

thanks!

cirocosta (approved):
lgtm 😁

PR #170, opened 2021-09-27
Make scripts more cross-platform

### LINKED ISSUE #168: Update RELEASE_DATE in release.sh
RELEASE_DATE=${RELEASE_DATE:-$(date -Iseconds)} in hack/release.sh is not supported on mac

idoru:
In particular, should work on Mac OS now.

Release date is now calculated at UTC for consistency. This conveniently avoids issues with supporting ISO8601 date format timezone.

[#168]

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -21,7 +21,7 @@ readonly ROOT=$(cd $(dirname $0)/.. && pwd)
 readonly SCRATCH=${SCRATCH:-$(mktemp -d)}
 readonly REGISTRY=${REGISTRY:-"$($ROOT/hack/ip.py):5000"}
 readonly BUNDLE=${BUNDLE:-$REGISTRY/cartographer-bundle}
-readonly RELEASE_DATE=${RELEASE_DATE:-$(date -Iseconds)}
+readonly RELEASE_DATE=${RELEASE_DATE:-$(TZ=UTC date +"%Y-%m-%dT%H:%M:%SZ")}
 
 readonly YTT_VERSION=0.36.0
 readonly YTT_CHECKSUM=d81ecf6c47209f6ac527e503a6fd85e999c3c2f8369e972794047bddc7e5fbe2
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -216,7 +216,7 @@ git_previous_version() {
         git tag --sort=-v:refname -l |
                 grep -A30 $version_filter |
                 tail -n +2 |
-                grep -P '^v\d+\.\d+\.\d+$' |
+                grep -E '^v\d+\.\d+\.\d+$' |
```
cirocosta: uh this one I believe is not 1:1 - I used `-P` 

```
       -P, --perl-regexp
              Interpret PATTERNS as Perl-compatible regular expressions (PCREs).  This option is experimental when combined  with  the  -z
              (--null-data) option, and grep -P may warn of unimplemented features.
```

so that `\d+` would match digits - **I think we could stick with `-E` which is in macos' `grep`, but then perhaps use `[:digit:]` or plain `[0-9]`?**

with the current path, on macos:

```
	RELEASE_VERSION:	v0.0.0-dev
	PREVIOUS_VERSION:	v0.0.6
```

but linux (ubuntu 20.04)

```
        RELEASE_VERSION:        v0.0.0-dev
        PREVIOUS_VERSION:
```

_(^ empty previous version)_

hack/release.sh
```diff
@@ -21,7 +21,7 @@ readonly ROOT=$(cd $(dirname $0)/.. && pwd)
 readonly SCRATCH=${SCRATCH:-$(mktemp -d)}
 readonly REGISTRY=${REGISTRY:-"$($ROOT/hack/ip.py):5000"}
 readonly BUNDLE=${BUNDLE:-$REGISTRY/cartographer-bundle}
-readonly RELEASE_DATE=${RELEASE_DATE:-$(date -Iseconds)}
+readonly RELEASE_DATE=${RELEASE_DATE:-$(TZ=UTC date +"%Y-%m-%dT%H:%M:%SZ")}
```
cirocosta, line 24: nice - works great :grin:

hack/release.sh
idoru: Ah yes, good catch. I thought my pair and I had tested it on Linux but I see now that `\d` only matches literal `d`. Pushing a fix.

cirocosta (approved):
lgtm 😁

PR #172, opened 2021-09-28
Add Netlify config

### LINKED ISSUE #159: Deploy the OSS Docs
The documentation for Cartographer needs to be deployed so that it can be accessed publically. 

For the time being, they will need to stay password-protected until the repo is again moved to open. 

## Acceptance Criteria
* The docs are deployed 
* Password protection is enabled

jonasrosland:
This will add a `netlify.toml` configuration file to specify versions and build commands, and make a temporary fix for the site so we can see it deployed properly.

Signed-off-by: Jonas Rosland <jrosland@vmware.com>

### DIFF
--- a/site/config.yaml
+++ b/site/config.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-baseURL: "https://template.dev/"
+#baseURL: "https://template.dev/"
 languageCode: "en-us"
 title: "Cartographer"
 theme: "template"
--- a/site/netlify.toml
+++ b/site/netlify.toml
@@ -1,0 +1,30 @@
+[build]
+publish = "public"
+command = "hugo --gc --minify"
+
+[context.production.environment]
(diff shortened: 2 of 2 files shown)


cirocosta:
nice, @jonasrosland 😁

looking at https://github.com/vmware-tanzu/cartographer/pull/170#issuecomment-928510688, it makes me wonder if we should be building on _any_ changes - is there a way of configuring the auto-build so that it's only for changes under `./site`? (or .. maybe we got a build because it's the first run of it?)

also, I noticed that the endpoint is reachable without any auth - is that fine? or something we should configure (given that this repo is still private)? this is one of the acceptance criteria defined in https://github.com/vmware-tanzu/cartographer/issues/159 

thank you!

jonasrosland:
Site has been locked down with a password, shared in Slack. Right now Netlify builds on each run (only takes a minute), but we can change the notification behavior for sure.

squeedee (approved):

Issue #167, opened 2021-09-27
Create secondary Docker config.json


DanieldeR:
in hack/setup we are mutating local docker config - can we create a new config and convince docker to use that?

cirocosta:
I _believe_ it totally doable - I'm sure `ko` would do the right thing (honor `DOCKER_CONFIG`) as it's based on [gccr](https://github.com/google/go-containerregistry/), but I think `imgpkg` is too, so I think _yes_, we can do it 😁 

any suggestions on where should it go? temporary file? a `.gitignore`d json (`docker-config.json`) in the root directory of the repository?

cirocosta:
#218 
#175

PR #175, opened 2021-09-28
use /tmp/cartographer-docker/ for docker config

idoru:
- fix handling for missing docker config
- shellchecked

[#167]

### DIFF
--- a/hack/docker-login.sh
+++ b/hack/docker-login.sh
@@ -18,37 +18,37 @@ set -o pipefail
 set -o nounset
 
 readonly DOCKER_REGISTRY=${DOCKER_REGISTRY:-"https://index.docker.io/v1/"}
-readonly DOCKER_CONFIG=${DOCKER_CONFIG:-"$(realpath ~/.docker/config.json)"}
-readonly DOCKER_USERNAME=$DOCKER_USERNAME
-readonly DOCKER_PASSWORD=$DOCKER_PASSWORD
+readonly DOCKER_CONFIG=${DOCKER_CONFIG:-"$(realpath ~/.docker)"}
+readonly DOCKER_USERNAME="${DOCKER_USERNAME}"
+readonly DOCKER_PASSWORD="${DOCKER_PASSWORD}"
 
 main() {
-        local auth=$(basic_auth $DOCKER_USERNAME $DOCKER_PASSWORD)
+        local auth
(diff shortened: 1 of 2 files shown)


hack/docker-login.sh
```diff
@@ -18,37 +18,39 @@ set -o pipefail
...
-        local current_content=$(jq '.' $config_file)
-        if [[ "$current_content" == "" ]]; then
-                echo '{}' >$config_file
+        if [ ! -f "${config_file}" ] || [ -z "$(jq '.' "$config_file")" ]; then
+                echo '{}' > "$config_file"
         fi
 
-        config=$(jq "$(printf "$qry" $server $token)" < "$config_file")
-        echo "$config" > $config_file
+        query="$(printf '.auths["%s"] = {"auth": "%s"}' "$server" "$token")"
```
cirocosta, line 39: given that we're generating a docker config that's just for this, I'd say we could instead just not bother at all about updates and always produce a fresh configuration - wdyt?

_if_ that's the case, we could do the equivalent of

```
DOCKER_CONFIG=$(mktemp -d) ko login --username admin --password admin 10.188.0.3:5000
```


which would be enough to generate a fresh docker config

hack/setup.sh
```diff
@@ -121,21 +124,23 @@ display_vars() {
...
 
         echo -e "\n\nregistry credentials:\n
-	username: admin
-	password: admin
-	"
+        username: admin
+        password: admin
+        "
+
+        export DOCKER_CONFIG="/tmp/cartographer-docker"
```
cirocosta: I think we shouldn't use `export` for that as it's bringing side-effects. e.g., CI passes because it runs commands one after another (`./hack/setup.sh cluster cartographer`), so when we execute `cartographer`, it then has already `DOCKER_CONFIG` properly set _because_ this function executed before (i.e, works due to the side-effect) - `./hack/setup.sh cluster ; ./hack/setup.sh cartographer` wouldn't work, I believe (`imgpkg` and `ko` need those credentials, but they wouldn't find).

hack/docker-login.sh
idoru: Fair enough. I've gone with just generating the file anew every time the setup is run. I opted to stick with the template for now as it looks simple enough.

squeedee (approved):

PR #176, opened 2021-09-28
Introduce rfc-0011-first-class-templates

squeedee:


rfc/rfc-0011-first-class-templates.md
```diff
@@ -0,0 +1,124 @@
...
+  name: petclinic
+  labels:
+    supply-chain: "a-good-one"
+spec:
+  source:
+    git:
+      url: https://github.com/spring-projects/spring-petclinic.git
+```
+
+# Supply Chain
```
squeedee, line 41: I expect the most contention will be around what's changed here.

Supply chain now has templating, but I hope by using well constrained `owner` crds for our templates we avoid them being complex or confusing.

PR #183, opened 2021-09-29
169 simplify test crd

### LINKED ISSUE #169: Simplify to One CRD
We now have two test crds, I think we could get away with one. There is some cleanup work to do.

* [x] Extend Test CRD to include map[string]interface()
* [x] Remove delivery CRD
* [x] In kuttl tests, use test crd instead of delivery
* [x] Move Test CRD from pipeline service to /tests
* [x] Improve directory structure of test CRD

squeedee:
* Moved the Test CRD
* Made the makefile rebuild it for test-* targets
* Ensured all tests used the Test CRD instead of the old Deliverable CRD.

See [#169]


Makefile
```diff
@@ -22,21 +22,25 @@ gen-manifests:
 		-f ./hack/boilerplate.go.txt \
 		config/crd/bases
 
-.PHONY: test-gen-objects
-test-gen-objects:
+tests/resources/zz_generated.deepcopy.go: tests/resources/*.go
```
waciumawanjohi: What is the purpose of adding `tests/resources/*.go`?

kuttl-test.yaml
```diff
@@ -19,7 +19,7 @@ testDirs:
   - tests/kuttl
 crdDir: config/crd/bases
 manifestDirs:
-  - tests/crd
+  - tests/resources/crds
```
waciumawanjohi, line 22: Both paths here install CRDs. Do we assume that the behavior of the two fields (crdDir and manifestDirs) is the same? Is there a risk that kuttl won't wait for the test CRD to be installed before running, courting the risk of flakes?

(I don't find the documentation of these two fields to be very clear... https://kuttl.dev/docs/testing/reference.html#testsuite)

squeedee: I understand your concern, but if it becomes an issue it's probably something to address with a PR to kuttl.

All I can add is that it works, and has done so for a while, without flakes. Putting our test crd in the config/crd/bases directory would be messy, and unfortunately crdDir is singular.

Makefile
squeedee: This is how `make` was intended to be used. you can call `test-gen-objects` now and it will not execute controller-gen unless one of the files specced in `tests/resources/*.go` has a newer timestamp than `tests/resources/zz_generated.deepcopy.go`

Technically it should be a list of `*.go` files _except_ `zz_generated.deepcopy.go`, but Make will elide the file for us anyway. 

I added this (and a couple of other make targets) so that `make test-unit` and `make test-kuttle` ensure the generated manifests and deepcopy are re-built as needed, without making us wait for them when they are unchanged.

squeedee: glad you asked though. I've forgotten more than I realized wrt Make. I need to replace the wildcard matching with distinct lists. will submit a patch

kuttl-test.yaml
idoru: The kuttl code indicates a potential race condition. Things in crdDir are waited for: 

https://github.com/kudobuilder/kuttl/blob/main/pkg/test/harness.go#L474-L484

But sadly, manifests, which are applied later are not:

https://github.com/kudobuilder/kuttl/blob/main/pkg/test/harness.go#L493-L497

@squeedee I'm inferring from their code that the design intent is `manifestDirs` isn't expected to introduce CRDs the tests depend upon, so our kuttl PR would be to change `crdDir string` -> `crdDirs []string` ?

waciumawanjohi: An option that does not necessitate upstream changes to kuttl is to mirror the CRDs in config/crd/bases in tests/resources/crds. Then we can point kuttl to tests/resources/crds in the crdDir.

Makefile
waciumawanjohi: Cool. Makes sense!

waciumawanjohi (approved):

kuttl-test.yaml
squeedee: @waciumawanjohi sounds good, could create a chore if you think a prophylactic effort is needed.

PR #188, opened 2021-10-01
provide a sample how to get the value from the param

bmoussaud:


idoru (approved):
Thanks for the fix. For what it's worth, we're hoping to remove the need to dig into the workload to access parameters like this in the first place.

Issue #23, opened 2021-08-25
Running an Informer on a Simple Change


cirocosta:
We want to use informers to switch to an event driven mechanism for stamped resource updates.

* Use informers to cause a new reconciliation.
* only add informers to GVKs that we are stamping
* run the whole reconcile loop, don't worry about which downstream component depends on the informed component. That's resolved and simplified by "when creating an object, cache any mutations"
* Team agrees that writing a reasonable end to end (perhaps based on the examples) test will help with confidence in such a massive change. May not need to do this for the spike, but worth considering?

Outcome: 
  * To stop putting in a 5 second reconcile

Tasks:
* [x] Remove supply chain dependence on retries
  * [x] delete retries, look for failing tests
  * [x] add tests to integrations that test informers
  * [x] ensure we use a GVK keyed cache to limit adding too many listeners
  * [x] test restart resilience
  * [x] clean up errors vs. status
* [x] do the same for workload
  * [x] delete retries, look for failing tests
  * [x] add tests to integrations that test informers
  * [x] ensure we use a GVK keyed cache to limit adding too many listeners
  * [x] test restart resilience
  * [x] clean up errors vs. status
* [x] do the same for deliverable
  * [x] delete retries, look for failing tests
  * [x] add tests to integrations that test informers
  * [x] ensure we use a GVK keyed cache to limit adding too many listeners
  * [x] test restart resilience
  * [x] clean up errors vs. status
* [x] do the same for delivery
  * [x] delete retries, look for failing tests
  * [x] add tests to integrations that test informers
  * [x] ensure we use a GVK keyed cache to limit adding too many listeners
  * [x] test restart resilience
  * [x] clean up errors vs. status
* [x] add note in contributing to the effect of: 
      > If you have an error condition, set the status of the object. Do not return the error to controller-runtime UNLESS an informer (.Watch) will not trigger a reconcile. If an informer will not cause a reconcile, then return an error so that controller-runtime will retry”
 
Note:
This has already been accomplished by the pipeline service

emmjohnson:
This has not been done anywhere besides pipeline. Opening up now.

waciumawanjohi:
TODO: Delete flaky tests from integration. See here: https://github.com/vmware-tanzu/cartographer/issues/111 and https://github.com/vmware-tanzu/cartographer/pull/195

- [ ] Delete from workload test
- [ ] Delete from deliverable test

squeedee:
[...]

PR #195, opened 2021-10-04
fixes flaky test

squeedee:
- just extends the timeout, hopefully that is enough.
- ultimately, story to use informers will make this test redundant
[#23]

### DIFF
--- a/tests/integration/supply_chain/workload_reconciler_test.go
+++ b/tests/integration/supply_chain/workload_reconciler_test.go
@@ -244,7 +244,7 @@ var _ = Describe("WorkloadReconciler", func() {
 				}
 
 				return logLine.Timestamp - previousSeconds
-			}, 2*time.Second).Should(BeNumerically(">", 1.0))
+			}, 2500*time.Millisecond).Should(BeNumerically(">", 1.0))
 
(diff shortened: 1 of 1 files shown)


cirocosta:
_> ultimately, story to use informers will make this test redundant_

aah maybe I'm misunderstanding the test, but I thought that polling interval wasn't relevant because of how what we're testing here is all about shorticuiting the interval in the first place 🤔  isn't that the case?

cirocosta (approved):
lgtm

squeedee:
> _> ultimately, story to use informers will make this test redundant_
> 
> aah maybe I'm misunderstanding the test, but I thought that polling interval wasn't relevant because of how what we're testing here is all about shorticuiting the interval in the first place 🤔 isn't that the case?

The way I read it is, this test ensures that backoff is shortcut by observing a supply chain change. It's really simple to prove that it's applied but not that the component chain's speed is reset to "fast"

If the speed isn't reset, then any component affected by the supply chain change will immediately restamp. but we might still wait hours to copy results down the chain. Happy to be corrected.

So assuming I'm right about the above, then there's no concerns about backoff when we use informers instead, because updates to the stamped resource trigger a reconcile

PR #196, opened 2021-10-04
Seperate kuttl tests for each controller

### LINKED ISSUE #171: Make the name in the sources/configs/image definitions optional
### Description of problem

As an operator, if I define a template that consumes a single upstream component (i.e. a single item in `images:`), and I reference that component by its shortcut name (.i.e `image` instead of `images.some-image.image`), then the name in the supply chain definition becomes meaningless.

### Proposed solution

Should we make this field optional?

### Example

```yaml
kind: ClusterSupplyChain
metadata:
  name: supply-chain
spec:
  selector:

  components:
    - name: source-provider
      templateRef:
        kind: ClusterSourceTemplate
        name: source

    - name: image-builder
      templateRef:
        kind: ClusterImageTemplate
        name: image
      sources:
        - component: source-provider
          name: source                       #< =============== make this field optional

    - name: deployer
      templateRef:
        kind: ClusterTemplate
        name: app-deploy
      images:
        - component: image-builder
          name: image                         #< =============== make this field optional
```

```yaml
kind: ClusterImageTemplate
metadata:
  name: image
spec:
  imagePath: .status.latestImage

  template:
    apiVersion: kpack.io/v1alpha1
    kind: Image
    metadata:
    spec:
      tag: #@ data.values.image_prefix + "$(workload.metadata.name)$"
      serviceAccount: service-account
      builder:
        kind: ClusterBuilder
        name: go-builder
      source:
        blob:
          url: $(source.url)$       #< =============== if I use the shortcut, the name above is never used
```

## Additional Context

We always provide a map of all the components that you consume (.i.e. `images.some-name.mage`), so if the name is missing from the supply chain definition, maybe we don't add it to this map?

squeedee:
- create make targets `test-kuttl-pipeline-service` and `test-kuttl-supply-chain`
- create seperate folders for pipeline-service and supply-chain
- ensure `make test-kuttle` will test both directories
- ensure `make list-kuttle` shows all tests, still focusable.

closes #171


cirocosta (approved):
lgtm 😁

Issue #198, opened 2021-10-05
provide single-yaml installation


cirocosta:
### Description of problem

Installing Cartographer is not as straightforward as it should be - rather than optimizing for the air-gapped case right of the bat, we should provide a simple installation method for those willing to fetch images from public repositories *ala* `k apply -f <release.yaml>`.

### Proposed solution

For the current release, we could update it providing images coming from
DockerHub (projectcartographer) and point folks at installing from there.

I think we can achieve that with something like below:

```
gh release download v0.0.6 bundle.tar
imgpkg copy --tar bundle.tar --to-repo projectcartographer --lock-output lock.yaml
imgpkg pull --lock lock.yaml -o bundle
kbld -f ./bundle -f cartographer.yaml
```

Given that ultimately it's a matter of relocating to DockerHub and then making
the YAML available, that should do it.

For future releases though, we could update the releases scripts to have them
producing not only the `bundle.tar`, which can be relocated etc, but also a
`cartographer.yaml` file that points at DockerHub images.

cirocosta:
done #200

PR #200, opened 2021-10-05
docs: update installation instructions

cirocosta:
from the commit message:

```
given that this repository now is open, it makes sense to provide
instructions for how to install cartographer based on images that we
publish to dockerhub.

the approach I took here was to keep the other usecases (which are ..
good usecases), but immediatelly provide the instructions that folks
would expect: `k apply -f <release>`.
```

note: while this is related to #198 (i.e., it addresses the need of having an easy way to install now by providing a yaml file there and the docs update), it does not address updating the workflow

### DIFF
--- a/site/content/docs/install.md
+++ b/site/content/docs/install.md
@@ -1,18 +1,34 @@
 # Installing Cartographer
 
-### Pre-requisites
+There are three recommended ways of installing Cartographer:
 
-1. A running Kubernetes cluster (v1.19+) with admin permissions
+1. with a single YAML file (quick!)
+2. using an [imgpkg bundle], making easier for relocation
+3. using a [carvel Packaging] objects, for an opinionated package-based
+   workflow
 
-In case you don't, [kind] is a great alternative for running Kubernetes
-locally, specially if you already have Docker installed (that is all it takes).
+If all you want to do is get started quick and fetching images from DockerHub
+is not an issue, [`1`](#1-single-file-quick-install) is the way to go.
+Otherwise, check out the other approaches (especially if you want to have
+everything relocated to a registry of your own).
 
-2. [cert-manager], so that certificates can be created and maintained to
-   support the controller's [admission webhook]
+tl;dr _(having [cert-manager] already installed)_:
 
-### Steps
+```bash
+kubectl create namespace cartographer-system
+kubectl apply -f https://github.com/vmware-tanzu/cartographer/releases/latest/download/cartographer.yaml
+```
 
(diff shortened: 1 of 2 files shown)


site/content/docs/install.md
```diff
@@ -1,18 +1,35 @@
 # Installing Cartographer
 
-### Pre-requisites
+There are three recommended ways of installing Cartographer:
 
-1. A running Kubernetes cluster (v1.19+) with admin permissions
+1. with a single YAML file (quick!)
+2. using an [imgpkg bundle], making easier for relocation
```
cirocosta, line 6: missing ref

site/content/docs/install.md
```diff
@@ -1,18 +1,35 @@
...
+3. using a [carvel packaging]
 
-In case you don't, [kind] is a great alternative for running Kubernetes
-locally, specially if you already have Docker installed (that is all it takes).
+If all you want to do is get started quick and fetching images from DockerHub
+is not an issue, `1` is the way to go. Otherwise, check out the other
```
cirocosta:
```suggestion
is not an issue, [`1`](#1-single-file-quick-install) is the way to go. Otherwise, check out the other
```

site/content/docs/install.md
```diff
@@ -55,14 +74,65 @@ ps.: although we recommend using [kapp] as provided in the instructions you'll
...
+
+
+## 2. Bundle tarball
+
+This installation method is great for those that want to host Cartographer's
+image and installation objects by themselves - either so they live in a public
+container image registry of their own, or for air-gapped scenarios where bits
+need to be moved in an offline fashion.
 
 First, head to the [releases page] and download the `bundle.tar` file available
```
cirocosta, line 139: missing ref

site/content/docs/install.md
```diff
@@ -55,14 +72,65 @@ ps.: although we recommend using [kapp] as provided in the instructions you'll
...
+To install it, first create the namespace where the controller will be
+installed:
+
+```bash
+kubectl create namespace cartographer-system
+```
+
+and then, submit the objects included in the release:
+
+```
```
cirocosta: missed `bash` qualifier

DanieldeR (approved):
The changes need to be made to the versioned docs as well. I pushed a new commit to make the same changes to /site/content/docs/v0.0.6/

I'm going to create an issue to better document the documentation process. It's a tab bit confusing and untenable at the moment.

The rest of the changes look great! (I haven't tried them though)

cirocosta:
got it, awesome! thank you! 🙌

PR #203, opened 2021-10-05
make release publish images to dockerhub

cirocosta:
in #200 I added instructions for one to leverage a `cartographer.yaml` file that we can distribute in our releases so that all it takes to install Cartographer is a `k apply -f <latest_release>`, but I hadn't updated the automation for doing so. Here is where I tackle the automation (+ updates to the script to surface that single yaml file).

please see the commit messages for details about the changes.

(no, I haven't pushed a tag for a release yet, but, have manually ran the same steps through and it looks good)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -55,8 +55,20 @@ jobs:
           KUBEBUILDER_ASSETS: /usr/local/bin
         run: make test
 
-      - name: run e2e example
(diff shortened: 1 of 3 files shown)


squeedee (approved):
All I ask is that we cut an v0.0.7 once this and #200 land, so that we have guidance for users and to check it end-to-end. I tried to cut a sample but again had issues with hack/release.sh talking to my cluster :/

PR #215, opened 2021-10-06
Draft: Introduce rfc-0012-discoverable-input-fields

waciumawanjohi:


rfc/rfc-0012-discoverable-input-fields.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Possible Solutions
+
+When a template is submitted to the cluster it is reconciled by a controller.
+That controller adds a status field to the template. In the status there is a field: `expectedInputs`.
+This is an array of
+1. All of the named params used in the template.
+2. All of the workload fields used in the template.
+
+When a supply chain is submitted to the cluster, the supply chain controller updates its status
```
scothis, line 22: Type checking is typically either static at compile time, or dynamic at runtime. This feels like it's trying to do static analysis at runtime.

rfc/rfc-0012-discoverable-input-fields.md
```diff
@@ -0,0 +1,79 @@
...
+all of the inputs expected by the new supply chain.
+2. We want workload authors to be able to quickly know what inputs their supply chain expects.
+
+## Possible Solutions
+
+When a template is submitted to the cluster it is reconciled by a controller.
+That controller adds a status field to the template. In the status there is a field: `expectedInputs`.
+This is an array of
+1. All of the named params used in the template.
+2. All of the workload fields used in the template.
```
scothis, line 20: How do you expect this information to be maintained as the template language evolves? I worry without careful planning, this contract could end up limiting the expressiveness of the templates. For example, how do you see ytt based templates working in this model?

Issue #12, opened 2021-09-01
Create RFC for delivery


DanieldeR:
* [x] Get RFC into repository [Stephen Levine]
* [ ] Generate stories from RFC [@squeedee]

Requirements:
https://docs.google.com/document/d/1n0jp0ud7j-6mLzbNNcC0EC5HAgCeDbHs0chFe9iBDsc

squeedee:
@jwntrs please try and copy the existing yaml and rfp info into this issue?

jwntrs:
I just linked to the private doc for now. We can update once we get started on a public RFC.

squeedee:
Stephen Levine will have an RFC ready for us today, and then I'll create initial stories.

jwntrs:
@squeedee I wrote up https://github.com/vmware-tanzu/cartographer/issues/220. Feel free to add/change whatever you want

PR #221, opened 2021-10-08
RFC: Delivery

sclevine:
[Readable](https://github.com/sclevine/cartographer/blob/patch-1/rfc/rfc-0013-delivery.md).

[#12]


sclevine:
Addresses #12

squeedee (approved):
As this was part of our last pre-oss team discussion, we have already decided that it's ratified. At least enough to begin work.

Issue #163, opened 2021-09-27
Select a resource by labels


jwntrs:
As an operator, I want to let developers associate a tekton/pipeline with their workload via label selection.

For example:

Developer defines:
```yaml
apiVersion: carto.run/v1alpha1
kind: Workload
metadata:
  name: dev-1-master
  labels:
    app.tanzu.vmware.com/workload-type: web
    pipelines.carto.run/group: dev-1                      # ========= DEFINE LABEL ===============
spec:
  ...
```

```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: pipeline-1
  labels:
    pipelines.carto.run/group: dev-1                 # ========= MATCHES LABEL ===============
    pipelines.carto.run/stage: testing
spec:
  params:
    - name: source-url
    - name: source-revision
  tasks:
    - name: test
      params:
        - name: source-url
          value: $(params.source-url)
        - name: source-revision
          value: $(params.source-revision)
      taskSpec:
        params:
          - name: source-url
          - name: source-revision
        steps:
          - name: test
            image: golang
            script: |-
              git clone $(params.source-url) && git co $(params.source-revision)

              make test
```

Operator defines:

```yaml
apiVersion: carto.run/v1alpha1
kind: ClusterSourceTemplate
metadata:
  name: testing-pipeline
spec:
  urlPath: .status.someUrl
  revisionPath: .status.someRevision

  template:
    apiVersion: carto.run/v1alpha1
    kind: Pipeline
    metadata:
      name: $(workload.name)$
    spec:
      runTemplateName: tekton

      selector:                                 # ========= SELECT BY LABEL ===============
        resource:
          apiVersion: tekton.dev/v1beta1
          kind: Pipeline
        matchingLabels: 
          pipelines.carto.run/group: $(workload.metadata.labels['pipelines\.carto\.run/group'])$
```

```yaml
apiVersion: kontinue.io/v1alpha1
kind: RunTemplate
metadata:
  name: tekton
spec:

  template:
    apiVersion: tekton.dev/v1beta1
    kind: PipelineRun
    metadata:
      generateName: $(pipeline.metadata.name)-
    spec:
      pipelineRef: {name: $(selected.metadata.name)}    # ========= REFERENCE SELECTED OBJECT =======

      params:
        - name: source-url
          value: $(pipeline.spec.inputs.url)
        - name: source-revision
          value: $(pipeline.spec.inputs.revision)
```

[...]

PR #230, opened 2021-10-13
Select resource by labels

idoru:
Implements #163 

If 0 or more than 1 resources match the supplied labels then an error is generated.


idoru:
Reopening as draft.

PR #234, opened 2021-10-14
Implement delivery controller

martyspiewak:


pkg/registrar/registrar.go
```diff
@@ -30,6 +30,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/vmware-tanzu/cartographer/pkg/controller/delivery"
+
```
cirocosta: nit: extra whitespace

```suggestion
```

pkg/repository/repository.go
```diff
@@ -57,6 +59,21 @@ func NewRepository(client client.Client, repoCache RepoCache) Repository {
...
 
+func (r *repository) GetDelivery(name string) (*v1alpha1.ClusterDelivery, error) {
+	delivery := &v1alpha1.ClusterDelivery{}
+
+	key := client.ObjectKey{
+		Name: name,
+	}
+
+	err := r.cl.Get(context.TODO(), key, delivery)
+	if err != nil && api_errors.IsNotFound(err) {
```
cirocosta: it looks like some of the methods do this check, thus, performing it within repository, while others leave for higher levels to do so.

e.g., here at the reconciler level we check 

https://github.com/vmware-tanzu/cartographer/blob/eabeb980942d346c6e35c00b2fa229f084d83050/pkg/controller/delivery/reconciler.go#L53-L61

vs in workload's reconciler

https://github.com/vmware-tanzu/cartographer/blob/77ee9657b2af7dbb2d68353419f8f246da58ad5b/pkg/controller/workload/reconciler.go#L59-L66

is there a particular reason we should go with one over the other? I think it'd be good if we could keep them very a-like and adopt a single pattern 😁

tests/integration/delivery/delivery_suite_test.go
```diff
@@ -0,0 +1,157 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/storage/names"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
+
```
cirocosta: nit: extra newline

tests/integration/delivery/delivery_suite_test.go
```diff
@@ -0,0 +1,157 @@
...
+	cancel           context.CancelFunc
+	controllerError  chan error
+	controller       *root.Command
+	controllerBuffer *gbytes.Buffer
+)
+
+const DebugControlPlane = false
+
+var _ = BeforeSuite(func() {
+	var err error
```
cirocosta, line 62: dang, so sad we have to have the same setup one again 😅  it'd be great if we could soon get around having a more approachable suite that lets us not have to repeat this (I don't think we necessarily have to address is in this PR though, just voicing a thought :p )

tests/integration/delivery/delivery_test.go
```diff
@@ -0,0 +1,171 @@
...
+			deliveryYaml := utils.HereYaml(`
+				---
+				apiVersion: carto.run/v1alpha1
+				kind: ClusterDelivery
+				metadata:
+				  name: my-delivery
+				spec:
+			      resources:
+			        - name: my-first-resource
+					  templateRef:
```
cirocosta, line 55: Screen Shot 2021-10-14 at 3 36 30 PM

there seems to be a lack of `\t` over there

tests/integration/delivery/delivery_test.go
```diff
@@ -0,0 +1,171 @@
...
+		AfterEach(func() {
+			for _, obj := range cleanups {
+				_ = c.Delete(ctx, obj, &client.DeleteOptions{})
+			}
+		})
+
+		Context("the referenced resource exists", func() {
+			BeforeEach(func() {
+				clusterSourceTemplateYaml := utils.HereYaml(`
+                  ---
```
cirocosta: same as the other comment - for things like this, I find using vim w/ `:set list` very useful (although replacing `\t` by anything visible would do the trick too I guess)

tests/integration/supply_chain/supply_chain_suite_test.go
```diff
@@ -30,19 +28,20 @@ import (
...
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/storage/names"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
+
```
cirocosta: nit: extra newline

```suggestion
```

tests/integration/pipeline_service/pipeline_service_suite_test.go
```diff
@@ -29,20 +27,21 @@ import (
...
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/storage/names"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
+
```
cirocosta: nit: extra newline

```suggestion
```

pkg/repository/repository_test.go
```diff
@@ -22,9 +22,11 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
cirocosta: not a big deal, but we usually call this import `kerrors` (like, `kubernetes errors ...`)

```
tests/helpers/client.go
21:     "k8s.io/apimachinery/pkg/api/errors"

pkg/repository/repository_test.go
25:     apierrors "k8s.io/apimachinery/pkg/api/errors"

pkg/repository/repository.go
21:     api_errors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/supplychain/reconciler_test.go
27:     kerrors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/supplychain/reconciler.go
23:     kerrors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/pipeline/reconciler.go
24:     kerrors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/pipeline/reconciler_test.go
27:     kerrors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/workload/reconciler.go
23:     kerrors "k8s.io/apimachinery/pkg/api/errors"

pkg/controller/workload/reconciler_test.go
27:     kerrors "k8s.io/apimachinery/pkg/api/errors"
```

I think `apierrors` is more "go"y, but it'd be not very consistent with the rest of the codebase

pkg/controller/delivery/reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/cartographer/pkg/conditions"
+
```
cirocosta: nit: extra newline

```suggestion
```

pkg/controller/delivery/reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+		r.conditionManager.AddPositive(TemplatesFoundCondition())
+		return nil
+	} else {
+		r.conditionManager.AddPositive(TemplatesNotFoundCondition(missing))
+		return fmt.Errorf("encountered errors fetching resources: %s", strings.Join(missing, ", "))
+	}
+}
+
+func (r *Reconciler) completeReconciliation(delivery *v1alpha1.ClusterDelivery, reconcileError error) (ctrl.Result, error) {
+	delivery.Status.Conditions, _ = r.conditionManager.Finalize()
```
cirocosta, line 89: is ignoring `changed` here the right approach? IIRC we use that returned value to make sure we don't issue more status updates than needed

pkg/repository/repository.go
```diff
@@ -57,6 +59,21 @@ func NewRepository(client client.Client, repoCache RepoCache) Repository {
...
+		return nil, nil
+	}
+
+	return delivery, err
```
cirocosta: IIRC we _usually_ leave the last return as the happy path, having captured the error and wrapped in early returns, but I'll defer to folks who have contributed more to the codebase recently to speak to it as it might not really be the prevailing pattern 😁

pkg/repository/repository.go
```diff
@@ -165,6 +203,7 @@ func (r *repository) createUnstructured(obj *unstructured.Unstructured) error {
 
 func (r *repository) patchUnstructured(existingObj *unstructured.Unstructured, obj *unstructured.Unstructured) error {
 	submitted := obj.DeepCopy()
+	// FIXME: I'm untested. What am I for? Patch doesn't block on RV's (is this a historical artifact of .Update?)
 	obj.SetResourceVersion(existingObj.GetResourceVersion())
```
cirocosta, line 212: good find! I _think_ we indeed did `Update` at some point back in time as we moved from `..OrUpdate..` to `..OrPatch..` (doing a quick lookup through the history of the previous repository)

```diff
diff --git a/pkg/realizer/component.go b/pkg/realizer/component.go
index 4284a43..442bc2e 100644
--- a/pkg/realizer/component.go
+++ b/pkg/realizer/component.go
@@ -57,7 +57,7 @@ func (r *componentRealizer) Do(component *v1alpha1.SupplyChainComponent, supplyC
                }
        }

-       err = r.repo.CreateOrUpdateUnstructuredObject(stampedObject)
+       err = r.repo.CreateOrPatchUnstructuredObject(stampedObject)
        if err != nil {
                return nil, ApplyStampedObjectError{
                        Err:           err,
```

tests/helpers/client.go
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package helpers
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
```
cirocosta:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

tests/helpers/client.go
```diff
@@ -0,0 +1,41 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
cirocosta:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

pkg/apis/v1alpha1/cluster_deployment_template.go
```diff
@@ -0,0 +1,49 @@
...
+// +kubebuilder:object:generate=true
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
```
cirocosta, line 26: ha, it might actually be worth getting rid of the status subresource for the template objects 🤔  we use the templates as "blueprints" only, being "just data" (like a ConfigMap), so it might actually be worth not having the subresource at all

(might be worth doing it in a separate issue for all templates though - I'd say we should keep in this PR so it doesn't look very apart).

cirocosta:
looking good! I left a few comments, mostly nitpicking though 👍

pkg/repository/repository.go
martyspiewak: Added an issue to standardize this: #239

tests/integration/delivery/delivery_suite_test.go
martyspiewak: I agree- maybe we can do that separately though

pkg/controller/delivery/reconciler.go
martyspiewak: Good catch- going to tackle this in a separate issue: #241

pkg/repository/repository.go
martyspiewak: Interesting- we're going to take a deeper look at this in a separate issue: #242

pkg/apis/v1alpha1/cluster_deployment_template.go
martyspiewak: Interesting. We made an issue to address this: #243

cirocosta (approved):

Issue #47, opened 2021-08-25
When Template is updated, immediately reconcile any matching workloads


cirocosta:
I want updating a template to fix workload issues without waiting for the exponential backoff.

```gherkin
Given a StarTemplate whose inner template is malformed in a way that prevents proper object submission
And a workload that is in exponential backoff from erroring on this component submission
When the template is patched
Then the workload immediately begins the reconciliation process
```

# Example
## Before

```
kind: Workload
status:
  conditions:
    - type: ComponentsUnableToSubmit
      status: "True"
      reason: TemplateRejectedByAPIServer
    - type: Ready
      status: "False"
      reason: ComponentsUnableToSubmit
```
## After
In no more than 10 seconds after submitting new template:
```
status:
  conditions:
    - type: ComponentsUnableToSubmit
      status: "False"
      reason: ComponentSubmissionComplete
    - type: Ready
      status: "True"
      reason: Ready
```

emmjohnson:
We should do this for deliverable/delivery also

waciumawanjohi:
Resolved in #304

PR #237, opened 2021-10-18
220 deliverable

emmjohnson:


tests/integration/delivery/delivery_suite_test.go
```diff
@@ -0,0 +1,157 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/storage/names"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
+
+	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
```
emmjohnson: extra new line

tests/integration/pipeline_service/pipeline_service_suite_test.go
```diff
@@ -29,20 +27,21 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/storage/names"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
```
emmjohnson: extra new line

tests/integration/supply_chain/supply_chain_suite_test.go
```diff
@@ -30,19 +28,20 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/storage/names"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	"github.com/vmware-tanzu/cartographer/tests/helpers"
```
emmjohnson: extra new line

pkg/apis/v1alpha1/common.go
```diff
@@ -40,6 +75,8 @@ func GetAPITemplate(templateKind string) (client.Object, error) {
 		template = &ClusterConfigTemplate{}
 	case "ClusterTemplate":
 		template = &ClusterTemplate{}
+	case "ClusterDeploymentTemplate":
```
cirocosta, line 78: per-RFC, it should actually be `ClusterDeployTemplate` 

> _- **ClusterDeployTemplates** are cluster-scoped resources that enable application operators to specify resource templates that deploy environments. ClusterDeployTemplates take a single source artifact and return a single deployment artifact._

https://github.com/vmware-tanzu/cartographer/blob/a90f8c8571a47d72f964b94ca77da5ca7108367a/rfc/rfc-0013-delivery.md?plain=1#L23

is there a particular reason for changing or slip/typo?

pkg/apis/v1alpha1/workload.go
```diff
@@ -79,40 +78,10 @@ type WorkloadSpec struct {
...
-	APIVersion string `json:"apiVersion,omitempty"`
-}
-
 type WorkloadStatus struct {
-	ObservedGeneration int64                        `json:"observedGeneration,omitempty"`
-	Conditions         []metav1.Condition           `json:"conditions,omitempty"`
-	SupplyChainRef     WorkloadSupplyChainReference `json:"supplyChainRef,omitempty"`
+	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
+	Conditions         []metav1.Condition `json:"conditions,omitempty"`
+	SupplyChainRef     OwnerReference     `json:"supplyChainRef,omitempty"`
```
cirocosta: hmmm I'm not sure about the name 😬  I'd say it's probably better to go with something like `ObjectReference` (like, `corev1.ObjectReference`) as this is not _really_ an "owner reference" in the sense that the supply chain doesn't _own_ the workload.

cirocosta: even from the point of view of making it "1:1" to `corev1`/`metav1`, `metav1.OwnerReference` is really different from `corev1.ObjectReference`:

- https://github.com/kubernetes/apimachinery/blob/4a9e16b3571218c2e15b8d0319f0b8e0e3fbacf5/pkg/apis/meta/v1/types.go#L303-L330

vs

- https://github.com/kubernetes/api/blob/03c9f537e2640508ed90ba19ba03f43fa5a2f76f/core/v1/types.go#L5388-L5440

pkg/apis/v1alpha1/cluster_delivery.go
```diff
@@ -55,6 +56,9 @@ type ClusterDeliveryStatus struct {
 type ClusterDeliveryResource struct {
 	Name        string                           `json:"name"`
 	TemplateRef DeliveryClusterTemplateReference `json:"templateRef"`
+	Params      []Param                          `json:"params,omitempty"`
+	Sources     []ResourceReference              `json:"sources,omitempty"`
```
cirocosta, line 60: in the RFC, when it talks about deployment, it always refers to _single source_ 


> _- **ClusterDeployTemplates** are cluster-scoped resources that enable application operators to specify resource templates that deploy environments._
>  _ClusterDeployTemplates take a single source artifact and return a single deployment artifact._ 
>  
>  _source -> deployment_
>
> (see [detailed explanation](https://github.com/vmware-tanzu/cartographer/blob/a90f8c8571a47d72f964b94ca77da5ca7108367a/rfc/rfc-0013-delivery.md?plain=1#L23-L24)) 


having in the examples a singular form, for instance:

```yaml
kind: ClusterDelivery
metadata:
  name: delivery
spec:
  selector:
    app.tanzu.vmware.com/deliverable-type: web
  resources:
    - name: config-provider
      templateRef:
        kind: ClusterSourceTemplate
        name: git-repository
    - name: deployer
      templateRef:
        kind: ClusterDeployTemplate
        name: app-deploy
      source:                                       #       (!)
        resource: config-provider
```

(see [deploy only example](https://github.com/vmware-tanzu/cartographer/blob/a90f8c8571a47d72f964b94ca77da5ca7108367a/rfc/rfc-0013-delivery.md?plain=1#L172))

although it still does allow the use of plural in order resources, like

```yaml
# ...
kind: ClusterDelivery
metadata:
  name: delivery
spec:
  selector:
    app.tanzu.vmware.com/deliverable-type: web
  resources:
# ...
    - name: promoter
      templateRef:
        kind: ClusterTemplate
        name: git-merge
      sources:
      - resource: performance-tester # need way to merge parallel tests eventually
```

(see [staging example](https://github.com/vmware-tanzu/cartographer/blob/a90f8c8571a47d72f964b94ca77da5ca7108367a/rfc/rfc-0013-delivery.md?plain=1#L327-L332))

 are we planning to have the singular form in?

pkg/controller/deliverable/reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+	deliverable.Status.Conditions, changed = r.conditionManager.Finalize()
+
+	var updateErr error
+	if changed || (deliverable.Status.ObservedGeneration != deliverable.Generation) {
+		deliverable.Status.ObservedGeneration = deliverable.Generation
+		updateErr = r.repo.StatusUpdate(deliverable)
+		if updateErr != nil {
+			logger.Error(updateErr, "update error")
+			if err == nil {
+				logger.Info("finished")
```
cirocosta: I think for this function we could actually `defer logger.Info("finished")` at the beginning of the function (i.e., line 115) so we don't have to annotate it everywhere but guarantee that it'll always be logged

cirocosta: hmm actually, i wonder if we should instead move the `finished` logging to a line right after the log `started` in `Reconcile()`.

IIUC, the `start & finish` pair of logs is useful to let us figure out where a reconciliation ultimately started & completed for a given object (not adding any extra information to the `finished`), so I'd imagine something there like

```go
func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	logger := logr.FromContext(ctx).
		WithValues("name", req.Name, "namespace", req.Namespace)
	ctx = logr.NewContext(ctx, logger)

	logger.Info("started")
	defer logger.Info("finished")
```

could do well.

_(having the same pattern across all reconcilers, and removing the `finished` calls in this function)_

wdyt?

pkg/controller/deliverable/reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+		WithValues("name", req.Name, "namespace", req.Namespace)
+	ctx = logr.NewContext(ctx, logger)
+	logger.Info("started")
+
+	reconcileCtx := logr.NewContext(ctx, logger)
+
+	deliverable, err := r.repo.GetDeliverable(req.Name, req.Namespace)
+	if err != nil || deliverable == nil {
+		if kerrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
cirocosta, line 61: very related to https://github.com/vmware-tanzu/cartographer/pull/237/files#r732141047 - I think in the whole branch started at line `60` we miss a "finished" log (which the `defer` proposed there in the comment would solve)

pkg/controller/deliverable/reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+	if len(deliverable.Labels) == 0 {
+		r.conditionManager.AddPositive(DeliverableMissingLabelsCondition())
+		return nil, fmt.Errorf("deliverable is missing required labels")
+	}
+
+	deliveries, err := r.repo.GetDeliveriesForDeliverable(deliverable)
+	if err != nil || len(deliveries) == 0 {
+		r.conditionManager.AddPositive(DeliveryNotFoundCondition(deliverable.Labels))
+
+		if err != nil {
```
cirocosta: wdyt of flattening this all out?

e.g.:

```go
          if err != nil {
                  r.conditionManager.AddPositive(DeliveryNotFoundCondition(deliverable.Labels))
                  return nil, fmt.Errorf("get delivery by label: %w", err)
          }

          if len(deliveries) == 0 {
                  r.conditionManager.AddPositive(DeliveryNotFoundCondition(deliverable.Labels))
                  return nil, fmt.Errorf("no delivery found where full selector is satisfied by labels: %v", deliverable.Labels)
          }

          if len(deliveries) > 1 {
                  r.conditionManager.AddPositive(TooManyDeliveryMatchesCondition())
                  return nil, fmt.Errorf("too many deliveries match the deliverable selector label")
          }
```

I think the less nesting the better as then it becomes more apparent what the necessary conditions to be fulfilled are, but that might be just personal taste

pkg/registrar/registrar.go
```diff
@@ -147,6 +153,38 @@ func registerDeliveryController(mgr manager.Manager) error {
...
+	); err != nil {
+		return fmt.Errorf("watch: %w", err)
+	}
+
+	mapper := Mapper{
+		Client: mgr.GetClient(),
+		Logger: mgr.GetLogger().WithName("deliverable"),
+	}
+
+	if err := ctrl.Watch(
```
cirocosta, line 178: so the only thing _not_ being watched is `templateRef` - I know that's something that we still want to do for `supplychain` (see https://github.com/vmware-tanzu/cartographer/issues/47), but makes me wonder if given that we're starting here with some fresh code (already adding a test for mapper, etc), we should also do this one for already - wdyt?

imo getting reconciliations shortcircuiting an exponential backoff is pretty great as I'd imagine people mostly get to big backoffs exactly because templates are not correct - sounds like a nice good win for not a lot of complexity (happy to see this in another story though - perhaps just a comment to https://github.com/vmware-tanzu/cartographer/issues/47 mentioning that it should be done for delivery too?)

cirocosta (commented):
niice, exciting to see it all coming together!  it'll be interesting to see it plugged in with the work on #238

I left a couple comments, feel free to create issues / defer to other PRs, whatever works best 😁

pkg/apis/v1alpha1/common.go
emmjohnson: Per our inception chat, we had decided that it would produce a deployment thus naming it ClusterDeploymentTemplate

pkg/registrar/registrar.go
emmjohnson: Added to issue #47

pkg/apis/v1alpha1/cluster_delivery.go
emmjohnson: Let's move this convo to slack. Leaving as is for now.

cirocosta (approved):

PR #238, opened 2021-10-18
GitWriter / Gitops example

### LINKED ISSUE #26: docs/example: delivery outside of SupplyChain
> "I want to run my configured containers on k8s"

```
Given a customer who has used Cartographer to take a workload through a supply-chain to config
When they read Cartographer documentation
Then they confidently configure Kapp to deploy the config
```

### Problem  

Cartographer currently takes workloads and turns them into k8s configuration, ready to be deployed on a cluster. Customers need a paved road to actually do this deployment. We should tell customers how they can use Carvel tooling to achieve this goal.

### Acceptance Criteria
A Cartographer use case example of using Carvel to deploy and application

_(ported from docs repo)_

### LINKED ISSUE #99: GitWritter with Tekton Example

waciumawanjohi:
closes #26  
closes #99


examples/gitops-to-app/app-operator/supply-chain-templates.yaml
```diff
@@ -0,0 +1,77 @@
...
+spec:
+  template:
+    apiVersion: kappctrl.k14s.io/v1alpha1
+    kind: App
+    metadata:
+      name: $(workload.metadata.name)$
+    spec:
+      serviceAccountName: default
+      fetch:
+        - git:
```
cirocosta: I think this is already addressed locally, but, let's not forget to use `http` rather than `git` 😁

examples/source-to-gitops/app-operator/supply-chain-templates.yaml
```diff
@@ -0,0 +1,204 @@
...
+    apiVersion: carto.run/v1alpha1
+    kind: Pipeline
+    metadata:
+      name: $(workload.metadata.name)$-git-writer
+    spec:
+      runTemplateRef:
+        name: git-writer
+
+      inputs:
+        input_config_map_name: $(workload.metadata.name)$
```
cirocosta, line 195: hmmm I think this is troublesome - IIUC, there will never be an update to this spec, thus, there'll never be more than one PipelineRun for that configmap (Tekton by itself will _not_ detect changes to the configmap and automatically create new runs) 🤔 

it might sound hacky, but, one way of getting that """right""" that I can think of would be to make the `ClusterTemplate` that's  creating the ConfigMap promote its `.data` as a config (thus, making it a ClusterConfigTemplate instead) which then we can apply a hashing function to in the ClusterTemplate that's stamping out this Pipeline

e.g.


```
ClusterImageTemplate         ClusterConfigTemplate             ClusterTemplate

  image-builder      <--img--    config-creator      <--config--  git-writer


  kpack/Image                      v1/ConfigMap                carto.run/Pipeline
                                                                 spec:
                                                                   inputs:
                                                                     digest: #@ sha256.sum(data.values.config)
```

wdyt?

examples/source-to-gitops/values.yaml
```diff
@@ -19,3 +19,10 @@ registry:
   server: https://index.docker.io/v1/
   username: projectcartographer
   password: a-very-hard-to-guess-password
+git_writer:
```
cirocosta, line 22: I imagine in a real-world use we'd expect this to come from the `Workload` object, do you envision it in that way too? 

I'm thinking in terms of the developer being able to stick with that Workload interface to define "hey, here's where final k8s config should be checked into" 🤔

hack/docker-compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+version: "3"
+
+networks:
+  gitea:
+    external: false
+
+services:
+  server:
+    image: gitea/gitea:1.15.3
```
cirocosta, line 23: wdyt of gitea? has it been useful for the setup or .. not really? looking at the config file under `hack`, seems like there's quite a bit to configure 😢  I think I remember @matthewmcnew mentioning looking at an approach for having git repositories easily configured (a long time ago I worked on https://github.com/cirocosta/gitserver-http and https://github.com/cirocosta/gpusher thing to help with that, but not sure how applicable it'd be)

examples/source-to-gitops/values.yaml
waciumawanjohi: > we'd expect this to come from the Workload object, do you envision it in that way too?

Yes.

Right now, these fields are fed into the carto/Pipeline object as params, so that they can be overwritten by params from the workload.

If you are asking, "Should this become a first class field of `Workload`?" I'd support that, as our examples and defaults are moving towards that gitops process.

hack/docker-compose.yaml
waciumawanjohi: > has it been useful for the setup

A local repo for our tests does bring value. Indeed, the example still needs some tweaking, because 2 people running the tests at the same time could cause a race condition and cause a spurious failure. (My first quick fix would be to make commits to new branches with random names)

The issue with gitea is not only that there's a fair bit to configure, but also that it doesn't yet work! (We're able to use what is there to make manual commits on a local machine, but wiring up the git-writer to use it failed).

examples/source-to-gitops/app-operator/supply-chain-templates.yaml
waciumawanjohi: Good catch. I've just echoed the data in the git-writer, very hacky. Later we can work on getting the git-writer to consume from the data passed.

waciumawanjohi:
Closing in favor of https://github.com/vmware-tanzu/cartographer/pull/282

PR #256, opened 2021-10-20
Remove condition change handling

squeedee:
Investigated why we still use changed:
* All our tests still work without it
* We introduced pkg/conditions/condition_manager.go#102-110
  which we believe ensures the status object emitted by finalize()
  is identical when nothing has changed.
* manually ensured that client.Status().Update() does not cause a
  resourceVersion bump when applying an identical status.

### DIFF
--- a/pkg/conditions/condition_manager.go
+++ b/pkg/conditions/condition_manager.go
@@ -54,8 +54,7 @@ type ConditionManager interface {
 
 	// Finalize	returns all conditions
 	// not idempotent! subsequent finalizes will keep adding Parent conditions
-	// The changed result represents whether the conditions have changed enough to warrant an update to the APIServer
-	Finalize() (conditions []metav1.Condition, changed bool)
+	Finalize() (conditions []metav1.Condition)
 
 	// IsSuccessful can be called any time. Start's off true, but an
 	// add of an unsuccessful condition (Positive-False or Negative-True)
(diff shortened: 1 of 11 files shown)


squeedee:
@waciumawanjohi @cirocosta @idoru can you all review please?

We're pretty certain all the issues this code was built for are now covered, but may have missed something.

waciumawanjohi:
This looks like a reasonable change to me.

waciumawanjohi (approved):

idoru:
Doesn't the old implementation avoid unnecessary calls to StatusUpdate() ?

squeedee:
> Doesn't the old implementation avoid unnecessary calls to StatusUpdate() ?

@idoru correct, however calls to StatusUpdate do nothing, as far as we can tell. So avoiding a possible (depending on caching) network call vs simplicity in the code. 

When I say "do nothing", even when called on retries (when in error state) or on refreshInterval of 5s, the call to Status.Update() causes no change in the resourceVersion

cirocosta:
> _When I say "do nothing", even when called on retries (when in error state) or on refreshInterval of 5s, the call to Status.Update() causes no change in the resourceVersion_

that's true, but I'd imagine the requests are definitely being made to kube-apiserver - is that right? it's certainly no big deal at a small scale, but with the current use of polling, we might quickly run out of requests available for hitting the apiserver without throttling

---

I think we should strive to not have those requests issued @squeedee

Issue #59, opened 2021-08-25
If a supply-chain's component is deleted, delete the object it stamped out


cirocosta:
_"I want auto-garbage collection when I delete supply-chain components"_

```gherkin
Given a Kontinue workload/supply-chain that has stamped out objects
When a supply-chain's existing component is deleted
Then the previously created object is deleted
```

Examples:

Initial State
```yaml
apiVersion: carto.run/v1alpha1
kind: ClusterBuildTemplate
metadata:
  name: first-template
spec:
  imagePath: .data.average_color_of_the_universe
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: first-build
    data:
      average_color_of_the_universe: "Cosmic latte"

---

apiVersion: carto.run/v1alpha1
kind: ClusterBuildTemplate
metadata:
  name: second-template
spec:
  imagePath: .data.best_litigant_against_board
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: second-build
    data:
      best_litigant_against_board: "Brown"

---

apiVersion: carto.run/v1alpha1
kind: ClusterSupplyChain
metadata:
  name: responsible-ops
spec:
  selector:
    integration-test: "workload-supply-chain-hardcoded-templates"
  components:
    - name: first-image-provider
      templateRef:
        kind: ClusterBuildTemplate
        name: first-template
    - name: second-image-provider
      templateRef:
        kind: ClusterBuildTemplate
        name: second-template

---

# STAMPED OUT OBJECT
apiVersion: v1
kind: ConfigMap
metadata:
  name: first-build
  labels:
    carto.run/workload-name: petclinic
    carto.run/cluster-supply-chain-name: responsible-ops
    carto.run/component-name: first-image-provider
    carto.run/cluster-build-template-name: first-template
  ownerReferences:
    - apiVersion: carto.run/v1alpha1
      kind: Workload
      ...
data:
  average_color_of_the_universe: "Cosmic latte"

---

# STAMPED OUT OBJECT
apiVersion: v1
kind: ConfigMap
metadata:
  name: second-build
  labels:
    carto.run/workload-name: petclinic
    carto.run/cluster-supply-chain-name: responsible-ops
    carto.run/component-name: second-image-provider
    carto.run/cluster-build-template-name: second-template
  ownerReferences:
    - apiVersion: carto.run/v1alpha1
      kind: Workload
      ...
data:
  best_litigant_against_board: "Brown"
```

[...]

Issue #56, opened 2021-08-25
If a supply-chain is deleted, delete the objects created by that supply-chain


cirocosta:
_"I want auto-garbage collection when I delete supply-chains"_

```gherkin
Given a Cartographer workload/supply-chain that has stamped out objects
When a supply-chain is deleted
Then the previously created objects are deleted
```

Examples:
```yaml
apiVersion: carto.run/v1alpha1
kind: ClusterSupplyChain
metadata:
  name: responsible-ops
spec:
  selector:
    integration-test: "workload-supply-chain-hardcoded-templates"
  components:
    - name: first-image-provider
      templateRef:
        kind: ClusterBuildTemplate
        name: first-template
    - name: second-image-provider
      templateRef:
        kind: ClusterBuildTemplate
        name: second-template

---

# SOME STAMPED OUT OBJECTS
```

When ClusterSupplyChain `responsible-ops` is deleted, the stamped out objects are deleted.

# Notes:
A likely place to insert this logic is in the supply-chain reconciler here:
https://github.com/vmware-tanzu/cartographer/blob/main/pkg/controller/supplychain/reconciler.go

Issue #54, opened 2021-08-25
If a *Template is deleted, delete the objects stamped out from that template


cirocosta:
_"I want auto-garbage collection when I delete *Templates"_

```gherkin
Given a Cartographer workload/supply-chain/*Template that has stamped out an object
When the *Template is deleted
Then the previously created object is deleted
```

Examples:
```yaml
apiVersion: carto.run/v1alpha1
kind: ClusterBuildTemplate
metadata:
  name: first-template
spec:
  imagePath: .data.average_color_of_the_universe
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: first-build
    data:
      average_color_of_the_universe: "Cosmic latte"

---

apiVersion: carto.run/v1alpha1
kind: ClusterSupplyChain
metadata:
  name: responsible-ops
spec:
  selector:
    integration-test: "workload-supply-chain-hardcoded-templates"
  components:
    - name: first-image-provider
      templateRef:
        kind: ClusterBuildTemplate
        name: first-template

---

# STAMPED OUT OBJECT
apiVersion: v1
kind: ConfigMap
metadata:
  name: first-build
  labels:
    carto.run/workload-name: petclinic
    carto.run/cluster-supply-chain-name: responsible-ops
    carto.run/component-name: first-image-provider
    carto.run/cluster-build-template-name: first-template
  ownerReferences:
    - apiVersion: carto.run/v1alpha1
      kind: Workload
      ...
data:
  average_color_of_the_universe: "Cosmic latte"

```

when `first-template` is deleted, `first-build` should no longer exist in the cluster.

# Notes
## Which controller?
If the SupplyChain cannot find the template for a component, its Ready condition will be False. This is a viable candidate for where to put the logic for this garbage collection. The logic cannot go into the workload reconciler without a refactor, as the workload will not enter its reconciliation function with a supply-chain that is not Ready.

## Differentiate between an error and an absence.
Controller-Runtime's client differentiates between a general error when trying to find an object and not finding the object. In the latter case, the error returned is `k8s.io/apimachinery/pkg/api/errors` `IsNotFound`
https://github.com/kubernetes-sigs/controller-runtime/blob/7d83250a445f2b5371e39e4197d3d0024f95fbdc/pkg/cache/internal/cache_reader.go#L52-L71

https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound

PR #274, opened 2021-10-28
First draft of RFC 0014

paulcwarren:


davidmirror-ops:
Hey @paulcwarren, thank you for this RFC.
We have a section in the weekly community meetings where all the maintainers team discuss new RFCs.
We would like to have you in the next meeting, happening Wednesday Nov. 3 at 11:00AM EDT (you can find more info on the [agenda](https://docs.google.com/document/d/1HwsjzxpsNI0l1sVAUia4A65lhrkfSF-_XfKoZUHI120/edit?usp=sharing))
Just let us know if you could attend to send you a calendar invite :)

Thanks again!

rfc/rfc-0014-change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Possible Solutions
+As the cartographer is already generally aware of workload inputs and outputs (artifacts) and how those artifacts from one component (service) are inputs into another.  It should, therefore, be possible to output, as additional status, the relationship between these artifacts and the service that produced it.
+
+For example:
+
+```
+status:
+   outputs:
+   - type: image
```
jwntrs, line 29: What happens when another downstream outputs gets created with this image (e.g. config)? Are we going to have two separate output trees that contain this information?

```yaml
status:
   outputs:
   - type: image
     value: abc123
     passed: [image-builder]
     from:
     - type: source
       value: def567
       passed: [workload, source-cloner, unit-tester]

   - type: config
     value:  shasum(some-giant-config)
     passed: [config-builder]
     from:
     - type: image   # <========= does this whole tree repeat again?
       value: abc123
       passed: [image-builder]
       from:
       - type: source
         value: def567
         passed: [workload, source-cloner, unit-tester]
```

rfc/rfc-0014-change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+Providing a real world example.  Given a supply chain that has several inputs, let’s say source code image and a debug flag.  When a developer applies a workload after changing their source code and turning on debug.  Several services in the supply chain may trigger at the same time and these inputs will traverse through the supply chain at different times.  But the developer, initiating the debug session will want to wait for that specific source code change to arrive “in cluster” in an image prepped for debugging before attaching their debugger.  So, it is important to know when both inputs have fully traversed the supply chain.
+
+## Possible Solutions
+As the cartographer is already generally aware of workload inputs and outputs (artifacts) and how those artifacts from one component (service) are inputs into another.  It should, therefore, be possible to output, as additional status, the relationship between these artifacts and the service that produced it.
+
+For example:
+
+```
+status:
+   outputs:
```
jwntrs, line 28: > But the kpack controller may also do work in response to a base OS image update. Nothing to do with the supply chain but impacting it none-the-less. As a result the supply chain may be triggered part way through by this input.

What does the output look like when this happens? Does it add a second image as follows:

```yaml
status:
   outputs:
   - type: image
     value: abc123
     passed: [image-builder]
     from:
     - type: source
       value: def567
       passed: [workload, source-cloner, unit-tester]

   - type: image
     value: xyz890
     passed: [image-builder]
     from:
     - type: source
       value: def567
       passed: [workload, source-cloner, unit-tester]
```

jwntrs (commented):
I left a couple inline comments. 

Overall, it seems like making things output-centric could lead to a bunch of duplication in the status. Maybe we could build out an artifact tree instead. Example below.

Also the ClusterSourceTemplates produce two values, `url` and `revision` so we should probably include both in our model.

```yaml
status:
  artifacts:
    - from: source-cloner
      type: source
      values:
        url: github.com/some/repo
        revision: blah10234
      produces: []     # <==== currently running tests
    - from: source-cloner
      type: source
      values:
        url: github.com/some/repo
        revision: def456
      produces:
        - from: unit-tester
          type: source
          values:
            url: github.com/some/repo
            revision: def456
          produces:
            - from: image-builder
              type: image
              values:
                image: abc123
              produces:
                - from: config-provider
                  type: config
                  values:
                    config: shasum(some-giant-config)
                  produces: []
            - type: image
              values:
                image: xyz890
              produces: []     # <==== hasn't produced config yet
```

Maybe instead of having `type` and `values` we could have something like:
```yaml
    - from: source-cloner
      source:
        url: github.com/some/repo
        revision: blah10234
      produces: []
```


Other questions:
How many artifact trees do we keep in the status?

Do we invalidate all the artifacts if the supply chain changes?

rfc/rfc-0014-change-tracking.md
sclevine: Maybe the nested entries should just be refs to the top-level. E.g.,
```yaml
status:
   outputs:
   - type: source
      value: def567
      passed: [workload, source-cloner, unit-tester]

   - type: image
     value: abc123
     passed: [image-builder]
     from:
     - type: source
       valueRef: def567

   - type: config
     value:  shasum(some-giant-config)
     passed: [config-builder]
     from:
     - type: image
       valueRef: abc123
```

jwntrs: @sclevine a source produces two values, `url` and `revision` so we probably want to track both. How would a `valueRef` work for cases like that?

rfc/rfc-0014-change-tracking.md
```diff
@@ -0,0 +1,46 @@
+# RFC 0014 Change Tracking
+
+## Summary
+Implement a change tracking feature to make it possible for clients to track input data as it traverses through the supply chain.
+Motivation
+Context
+At the core of the architecture of cartographer is the concept that a supply chain is the orchestration of objects and their controllers.
```
alexbarbato: nit: My understanding is that the core of cartographer is actually `choreography` versus `orchestration` as stated here.

alexbarbato (commented):
A super nit, but going to provide it anyway. It was great to get to review this :D

alexbarbato:
I love this RFC.

It does feel like it's attempting to solve a problem that is potentially precluded by a smaller problem - how does a developer get an understanding of how their supply chain "ran"?

Right now I find myself using `lineage` or `tree`, but even in those cases it's not really explicit beyond giving me a high level tree. I wonder how much of this RFC becomes less painful for a developer if they have a clearer way to view the progress of a supply chain at all.

(I could be being totally naive and maybe this already exists OR this is the exact RFC to try to solve that)

waciumawanjohi:
This RFC relates to several previous RFCs:
1. [RFC 3](https://github.com/vmware-tanzu/cartographer/blob/main/rfc/rfc-0003-intermediate-value-crds.md) addresses this case exactly. The RFC proposes using separate objects in the cluster to store state. These separate state holding objects are referred to as Intermediate CRDs. The RFC notes that state could be stored on the Workload instead.
2. [RFC 4](https://github.com/vmware-tanzu/cartographer/blob/main/rfc/rfc-0004-handle-multiple-supply-chains-selecting-workload.md) explores the case where more than 1 supply chain reconciles with a workload. It proposes a separate object in the cluster that represents the logical pairing between supply chain and workload. This object could be used to store state.
3. [RFC 11](https://github.com/vmware-tanzu/cartographer/blob/main/rfc/rfc-0011-first-class-templates.md) Proposes using Intermediate CRDs to stamp out objects.

waciumawanjohi:
If we are to store state about outputs on the workload, I prefer @jwntrs' tree where the root is a top level output (e.g. the output of a resource that only depends on the workload, not on any other template's output).

> How many artifact trees do we keep in the status?

Any artifact that is currently exposed or consumed should be kept in the tree. So if `source-cloner` has exposed 99 revisions over a period, and `unit-tester` is currently processing revision N and `image-builder` is processing revision N-1 and `image-builder`'s most recently completed work is revision N-2, then no tree for N-3 or earlier should be maintained. The moment no logical entity in the supply chain is processing or exposing an output or child of an output, that output is historical.

waciumawanjohi:
> Do we invalidate all the artifacts if the supply chain changes?

Yes. We already have stories for deleting the stamped out objects in this case. (See https://github.com/vmware-tanzu/cartographer/issues/56, https://github.com/vmware-tanzu/cartographer/issues/59, https://github.com/vmware-tanzu/cartographer/issues/54). We should also delete the references to their outputs.

rfc/rfc-0014-change-tracking.md
sclevine: Might need IDs/UUIDs or something similar

cirocosta: +1 on IDs for each entry (perhaps running a hash function on the value? given that we have the values themselves once per output, we can always find the pre-image of the hashing result if we want need to surface the underlying content of the hash) 

I wonder if `workload` itself could be considered an """output"""  which could be referenced like any other value so that we can always point back to the workload as it was at such point in time 🤔

rfc/rfc-0014-change-tracking.md
jwntrs: Addressed by https://github.com/vmware-tanzu/cartographer/pull/274#discussion_r740389186 as well

rfc/rfc-0014-change-tracking.md
sclevine: > perhaps running a hash function on the value

💯  Maybe sha256 of ordered JSON?

> I wonder if workload itself could be considered an """output""" which could be referenced like any other value so that we can always point back to the workload as it was at such point in time 🤔

This could get complicated, because it isn't possible to guarantee that all related artifacts are generated from the same Workload spec. We could store revisions of the Workload spec as artifacts, and ref them as inputs of every non-Workload artifact. But I'm not sure if that's worth it, given that we aren't going to be able to capture every input with Cartographer anyways. Might be better to rely on Tekton Chains for Tekton and SBoM for Kpack images for provenance.

sclevine: Updated suggestion:
```yaml
  artifacts:
  - source: # oneOf(source,image,config)
      id: <SHA:string>
      uri: <:string>
      revision: <:string>
      passed: <resource-names:[]string>
      from:
      - id: <SHA:string>
  - image: # oneOf(source,image,config)
      id: <SHA:string>
      digest: <"sha256:"digest:string>
      passed: <resource-names:[]string>
      from:
      - id: <SHA:string>
  - config: # oneOf(source,image,config)
      id: <SHA:string>
      config: <:string>
      passed: <resource-names:[]string>
      from:
      - id: <SHA:string>
```

rfc/rfc-0014-change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+       value: def567
+       passed: [workload, source-cloner, unit-tester]
+```
+
+Here we see the run image `abc123`, produced by the `image-builder` service, and “containing” the source code image `def567`. 
+
+Using the above example again when a workload is applied with the cli, updating both the source image and debug flags, it can then wait for the workload status to be Ready and for an output status showing the last artifact in the supply chain contains the original source image and the debug flag values we expect.
+
+## Concerns/Questions
+How much does a consumer of this output status need to understand about supply chains?  Does a consumer have to “know” what the last artifact produced by a supply chain is?  And how?
```
hectorj2f, line 48: I'd advocate to let the customer know when its artifact didn't reach the desired state. That means you need to have at least three states e.g. (Pending, Finished or Failed). Otherwise it is difficult to understand what in a coreography went wrong or whether the chain is still under reconciliation.

rfc/rfc-0014-change-tracking.md
```diff
@@ -0,0 +1,46 @@
...
+
+Providing a real world example.  Given a supply chain that has several inputs, let’s say source code image and a debug flag.  When a developer applies a workload after changing their source code and turning on debug.  Several services in the supply chain may trigger at the same time and these inputs will traverse through the supply chain at different times.  But the developer, initiating the debug session will want to wait for that specific source code change to arrive “in cluster” in an image prepped for debugging before attaching their debugger.  So, it is important to know when both inputs have fully traversed the supply chain.
+
+## Possible Solutions
+As the cartographer is already generally aware of workload inputs and outputs (artifacts) and how those artifacts from one component (service) are inputs into another.  It should, therefore, be possible to output, as additional status, the relationship between these artifacts and the service that produced it.
```
hectorj2f, line 22: When I read this statement, I get the impression the list of inputs/outputs defines some kind of orchestraction which opposes to what we propose here following a choreography model (as an event broker). 

In contrast, I'd expect each stage in the supply chain to watch on its required input resource (without any explicit definition) to generate any output, and viceversa. I assume stages are heterogenous and can be added dynamically without having to change a whole ClusterSupplyChain definition (e.g. i mean updating all the outputs and inputs refs in the whole document).

hectorj2f: Perhaps the real problem is how to determine if a stage reached its desired state or not. In that direction, I'd suggest to rely on standarized conditions https://github.com/deads2k/enhancements/blob/e4e99850831a45b924f66f67ecd388c36a23ab6c/keps/sig-api-machinery/1623-standardize-conditions/README.m.  You could identify when a stage has reached its desired state by assuming a `HasCondition("Ready")` was added to the status.conditions of the managed resources independently of its nature.

rfc/rfc-0014-change-tracking.md
waciumawanjohi: From the [Nov 22 Office Hours](https://docs.google.com/document/d/1ImIh7qBrOLOvGMCzY6AURhE-a68IE9_EbCf0g5s18vc/edit):
Issue #378 points out that it is useful to note not only the component name that has yielded an artifact, but also the object that artifact has been read from. E.g.
```yaml
  artifacts:
  - source: # oneOf(source,image,config)
      id: <SHA:string>
      uri: <:string>
      revision: <:string>
      passed: <resource-names:[]string>
      from:
      - id: <SHA:string>
        kind: <:string>
        apiVersion: <:string>
        name: <:string>
        namespace: <:string>
        resourceVersion: <:string>
```

waciumawanjohi: Clarification:
```yaml
  artifacts:
  - source: # oneOf(source,image,config)
      id: <SHA:string>
      uri: <:string>
      revision: <:string>
      passed:
      - resource-name: <:string>
        kind: <:string>
        apiVersion: <:string>
        name: <:string>
        namespace: <:string>
        resourceVersion: <:string>
      from:
      - id: <SHA:string>
```
first item in the list of passed should be the resource that originally created the artifact.

rfc/rfc-0014-change-tracking.md
paulcwarren: Changed.

rfc/rfc-0014-change-tracking.md
waciumawanjohi: > when its artifact didn't reach the desired state

The artifact is the result of a successful state change. For example, let us say a `ClusterSupplyChain` has a `ClusterSourceTemplate` which templates a FluxCD GitRepository (as in the [examples](https://github.com/vmware-tanzu/cartographer/tree/main/examples/source-to-knative-service)). The artifact of this template is some SHA. At the moment that the artifact, the SHA, is known to Cartographer, the template has succeeded in reaching the desired state. It is impossible to decorate _the artifact_ with any status other than `Successful`/`Finished`.

That said, thinking about UX to enable devs and ops to understand what resources in the SupplyChain are in a pending or failed state is valuable.

hectorj2f: @waciumawanjohi  Yes, that is right. However there is a lack of visibility during the 20 min or more that could take a pipeline take to product an artifact.

jwntrs:
In the office hours this week (Jan 10th, 2022), we talked about how we could extend this type of artifact tracing to track artifacts across clusters.

Assuming you started with a SupplyChain that looked something like the following (from [git-writer example](https://github.com/vmware-tanzu/cartographer/tree/main/examples/gitwriter-sc)): 

[source-provider] -> [image-builder] -> [config] -> [git-writer]

But you made a small change and wrapped the git-writer in the SupplyChain with a `ClusterSourceTemplate`, it would let you emit the revision/url of the committed config. By doing this you would end up with an artifact representing the commit to the config-repo as the last item in the status of your workload.

So your workload status might end up looking like this:
```yaml
artifacts:
  - source: source
      id: 000
      url: https://github.com/my-org/my-source
      revision: abc123
      passed:
      - resource-name: source-provider
        kind: GitRepository
  - source: image
      id: 111
      image: gcri.io/asdfsdfasdf
      url: https://github.com/my-org/my-repo
      revision: abcdef12345
      passed:
      - resource-name: image-builder
        apiVersion: kpack.io/v1alpha2
        kind: Image
      from:
      - id: 000
  - source: config
      id: 222
      config: shasum(of-some-config)
      passed:
      - resource-name: my-config
        apiVersion: v1
        kind: ConfigMap
      from:
      - id: 111
  - source: source
      id: 333
      url: https://github.com/my-org/my-config
      revision: xyz789
      passed:
      - resource-name: config-writer
        kind: Runnable
      from:
      - id: 222
```

Now lets assume you are following the [basic-delivery example](https://github.com/vmware-tanzu/cartographer/tree/main/examples/basic-delivery) on another cluster.

You would end up with a Deliverable with the following status:
```yaml
artifacts:
  - source: source
      id: 333
      url: https://github.com/my-org/my-config
      revision: xyz789
      passed:
      - resource-name: config-provider
        kind: GitRepository
 - source: some other-artifact
   ...
```

You can see how we could correlate these two completely independent resources, based solely on their artifacts.

jwntrs:
@rawlingsj made a really insightful comment in another unrelated thread that we should consider here:
> Customers will likely not be willing to give access from a dev cluster (and developer users) to others such as staging / production.  Instead it will probably be easier to send events out of those locked down clusters to a centralized place for users to get visibility into their environments.

My comment [here](https://github.com/vmware-tanzu/cartographer/pull/274#issuecomment-1010335810) definitely assumes read access to all clusters. So maybe we should start thinking about writing out artifacts as k8s events at the same time as writing them to the workload/deliverable status. That way if someone is already emitting k8s events from their prod clusters, we can integrate into their existing tooling.

emmjohnson:
Several RFCs attempt to close this out. Health status makes a new approach possible. Closing this PR.

PR #293, opened 2021-11-03
Replace SupplyChain retries with Informers for Templates

emmjohnson:
[#23]

Co-authored-by: Rasheed Abdul-Aziz <rabdulaziz@vmware.com>


pkg/controller/delivery/reconciler.go
```diff
@@ -70,6 +70,7 @@ func (r *Reconciler) reconcileDelivery(delivery *v1alpha1.ClusterDelivery) error
 	var missing []string
 	for _, resource := range delivery.Spec.Resources {
 		_, err := r.repo.GetDeliveryClusterTemplate(resource.TemplateRef)
+		// FIXME: should be checking if err IsNotFound
```
cirocosta, line 73: is this something we want to tackle before merge / write a story? what's the ideal here? if the err is not a NotFound, then completely bail out because it's something bad?

pkg/registrar/map_functions.go
```diff
@@ -133,6 +138,66 @@ func (mapper *Mapper) RunTemplateToRunnableRequests(object client.Object) []reco
...
+
+	list := &v1alpha1.ClusterSupplyChainList{}
+
+	err = mapper.Client.List(
+		context.TODO(),
+		list,
+	)
+
+	if err != nil {
+		mapper.Logger.Error(err, "template to supply chain requests")
```
cirocosta: wdyt of

```suggestion
		mapper.Logger.Error(err, "list cluster supplychain")
```

?

`template to supply chain requests` would be describing the whole function, rather than just the problem that actually occurred

pkg/registrar/registrar.go
```diff
@@ -137,6 +137,20 @@ func registerSupplyChainController(mgr manager.Manager) error {
...
+		Client: mgr.GetClient(),
+		Logger: mgr.GetLogger().WithName("supply-chain"),
+	}
+
+	for _, template := range v1alpha1.ValidSupplyChainTemplates {
+		if err := ctrl.Watch(
+			&source.Kind{Type: template},
+			handler.EnqueueRequestsFromMapFunc(mapper.TemplateToSupplyChainRequests),
+		); err != nil {
+			return fmt.Errorf("watch template (%v): %w", template, err)
```
cirocosta: I don't think `%v` is a good choice here - it'll produce a "stringified" output that's not really that helpful

e.g.,

```go
          template := &v1alpha1.ClusterSourceTemplate{}
          panic(fmt.Errorf("watch template (%v): %w", template, errors.New("foo")))
```

leads to

```
panic: watch template (&{{ } {      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {{nil  []}  } {}}): foo

goroutine 1 [running]:
main.main()
        ...main.go:13 +0xa8
```

(note that we can't just grab  `GroupVersionKind()` formatting on `%s` either as the gvk wouldn't be filled)

pkg/registrar/map_functions.go
```diff
@@ -25,13 +27,16 @@ import (
...
+//counterfeiter:generate sigs.k8s.io/controller-runtime/pkg/client.Client
+
 //counterfeiter:generate . Logger
 type Logger interface {
 	Error(err error, msg string, keysAndValues ...interface{})
 }
 
 type Mapper struct {
 	Client client.Client
+	// fixme We should accept the context, not the logger - then we get the right logger and so does the client
```
cirocosta, line 39: makes me wonder if we should update our linter setting to remind us about fixmes/todos 🤔 

https://github.com/matoous/godox

pkg/controller/delivery/reconciler.go
squeedee: We have an example of this change for supply chain reconciler in this PR. If it's not found, there is no 'Error' to report and it should _not_ return a reconciler error. This is fixed in the PR for delivery-informers.

pkg/registrar/map_functions.go
squeedee: No thanks. this needs to be in a metric style static analysis tool such as Sonarqube. These are placeholders that need to be adressed _eventually_. I leave markers so that anyone passing them by while making changes understands there is intent to change. Imagine you think to yourself, while working in this code "this really should be a context" - this `fixme` validates your thoughts and saves you searching for a story. 

if we get a good static analysis tool running, we'll be able to decide when to add stories as needed, and we can annotate fixme/todo with story links.

pkg/registrar/map_functions.go
squeedee: new commit to fix

pkg/registrar/registrar.go
squeedee: no what we want is the type name, but reflection returns errors of it's own... what a rabbit hole.

squeedee: As a side note, the root cause of us missing the low-quality here is the complete absence of useful testing for registrar.

pkg/controller/delivery/reconciler.go
cirocosta: oh, I see - cool! thx!

pkg/registrar/map_functions.go
cirocosta: makes sense! 👍

cirocosta (approved):
lgtm! nice

PR #304, opened 2021-11-05
23 informers workload

squeedee:


tests/integration/supplychain/workload_reconciler_test.go
```diff
@@ -274,4 +278,164 @@ var _ = Describe("WorkloadReconciler", func() {
...
+
+			test = &resources.Test{}
+
+			// FIXME: make this more obvious
+			Eventually(func() ([]v1.Condition, error) {
+				err := c.Get(ctx, client.ObjectKey{Name: "test-resource", Namespace: testNS}, test)
+				return test.Status.Conditions, err
+			}).Should(BeNil())
+
+			_ = c.Get(ctx, client.ObjectKey{Name: "test-resource", Namespace: testNS}, test)
```
emmjohnson: Remove this

tests/kuttl/supplychain/templates-ytt/01-assert.yaml
```diff
@@ -0,0 +1,27 @@
...
+#metadata:
+#  name: responsible-ops---consume-output-of-resources
+#status:
+#  conditions:
+#    - type: TemplatesReady
+#      status: "True"
+#      reason: Ready
+#    - type: Ready
+#      status: "True"
+#      reason: Ready
```
martyspiewak: Is there a reason this file is commented out?

pkg/registrar/map_functions.go
```diff
@@ -40,6 +40,53 @@ type Mapper struct {
...
+	if err != nil {
+		mapper.Logger.Error(err, "list ClusterSupplyChains")
+		return nil
+	}
+
+	templateKind := template.GetObjectKind().GroupVersionKind().Kind
+
+	var supplyChains []*v1alpha1.ClusterSupplyChain
+	for scIndex := range list.Items {
+		sc := list.Items[scIndex]
```
martyspiewak:
```suggestion
	for _, sc := range list.Items {
```
nitpick- this would be more consistent with what we do in `ClusterDeliveryToDeliverableRequests` and `ClusterSupplyChainToWorkloadRequests` below and is a bit easier to reason about while reading through the code.

tests/integration/supplychain/policy.yaml
```diff
@@ -15,11 +15,13 @@
 apiVersion: audit.k8s.io/v1 # This is required.
 kind: Policy
 # Don't generate audit events for all requests in RequestReceived stage.
```
martyspiewak:
```suggestion

```
That comment seems to be irrelevant now as it was on the `omitStages` portion that was removed.

martyspiewak (changes requested):

martyspiewak:
Looks great! Left a few comments with some small changes but other than that lgtm!

pkg/registrar/map_functions.go
emmjohnson: That actually causes major problems. See https://play.golang.org/p/GzU80LswSEt

tests/kuttl/supplychain/templates-ytt/01-assert.yaml
emmjohnson: no

pkg/registrar/map_functions.go
squeedee: we also inlined a comment to guide people.

martyspiewak (approved):

PR #305, opened 2021-11-07
Fix 'changed' edge case of ConditionManager

squeedee:
With `PreviousConditions`, no matter what you `Add`, if the resulting set of conditions is the same as `PreviousConditions`, `Changed` should be false.


emmjohnson (approved):

emmjohnson:
I can't remove approval.....

emmjohnson:
Can we remove this @squeedee

PR #310, opened 2021-11-08
Use unique names between kuttl tests.

emmjohnson:
There is a race condition between templates-ytt and templates-consume-output-of-components.


waciumawanjohi:
This isn't addressing the test pollution issue, which is that there are two ClusterConfigTemplates (same for ClusterTemplates) that are being created during kuttl tests with the name "example-config---consume-output-of-resources". That name must be deduplicated.

martyspiewak:
@waciumawanjohi I believe it does fix the pollution because those templates are overwritten when they are applied in the `00-templates.yaml` phase so it doesn't really matter that they have the same name. The real problem was that the templates were both creating a ConfigMap with the same name, so when the second test would run, the expected ConfigMap already existed from a previous test.

I would still be in favor of renaming the templates to match the convention we have in our other tests of using the test name in the template name, but I believe we'd still need the change made in this PR in order to actually get rid of the pollution as the problem is the resulting ConfigMaps having the same name, not the templates themselves. 

(I tried to verify this by renaming the templates but leaving the ConfigMap names as they are and I still saw the flakiness- renaming the ConfigMaps does seem to get rid of the flakiness though).

martyspiewak:
After talking with @waciumawanjohi it looks like the ConfigMap names can stay as they should be created in separate namespaces. The template names and supplychain names/labels are what need to be changed as they exist at the cluster scope.

The flakiness seems to come from the fact that both tests had the same supply chain name/label and so it was non-deterministic which supply chain definition was in operation when the workload in the `templates-ytt` test reconciled. Changing those names seems to get rid of that flakiness.

martyspiewak (approved):

tests/kuttl/supplychain/templates-ytt/01-supply-chain.yaml
```diff
@@ -16,41 +16,41 @@
 apiVersion: carto.run/v1alpha1
 kind: ClusterSupplyChain
 metadata:
-  name: responsible-ops---consume-output-of-resources
+  name: responsible-ops---templates-ytt
 spec:
   selector:
-    integration-test: "templates-consume-output-of-resources"
+    integration-test: "templates-templates-ytt"
```
martyspiewak: nitpick- we can probably drop the first "templates" to have it match the name of the test, like the `templates-consume-output-of-components` test does. not a big deal though!

martyspiewak:
lgtm!

PR #314, opened 2021-11-09
23 informers deliverable

emmjohnson:


squeedee (changes requested):
We shouldn't review this until #304 lands

Also we've got a flake that should have died by now, will investigate:
Screen Shot 2021-11-08 at 11 50 09 PM

squeedee (approved):

pkg/controller/deliverable/reconciler.go
```diff
@@ -31,25 +34,38 @@ import (
 	"github.com/vmware-tanzu/cartographer/pkg/utils"
 )
 
-const reconcileInterval = 5 * time.Second
+type Reconciler interface {
+	Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error)
+	AddTracking(dynamicTracker DynamicTracker)
+}
```
martyspiewak: This interface is declared here, in workload, and in runnable. I think it would make more sense to declare it once where we use it, in the registrar, rather than redeclaring it in each package.

pkg/registrar/registrar.go
```diff
@@ -213,13 +213,18 @@ func registerDeliverableController(mgr manager.Manager) error {
...
+	reconciler := deliverable.NewReconciler(repo, conditions.NewConditionManager, realizerdeliverable.NewRealizer())
 	ctrl, err := pkgcontroller.New("deliverable", mgr, pkgcontroller.Options{
-		Reconciler: deliverable.NewReconciler(repo, conditions.NewConditionManager, realizerdeliverable.NewRealizer()),
+		Reconciler: reconciler,
 	})
 	if err != nil {
 		return fmt.Errorf("controller new: %w", err)
 	}
 
+	reconciler.AddTracking(&external.ObjectTracker{
```
cirocosta: one implicit detail of using `external.ObjectTracker` is that one can actually "pause the watching" by making use of clusterapi's `paused` annotation:

```go
          // PausedAnnotation is an annotation that can be applied to any Cluster API
          // object to prevent a controller from processing a resource.
          //
          // Controllers working with Cluster API objects must check the existence of this annotation
          // on the reconciled object.
          PausedAnnotation = "cluster.x-k8s.io/paused"
```

used as a predicate in

```go
          log.Info("Adding watcher on external object", "GroupVersionKind", gvk.String())
          err := o.Controller.Watch(
                  &source.Kind{Type: u},
                  handler,
                  predicates.ResourceNotPaused(log),
          )
```

(_see https://github.com/kubernetes-sigs/cluster-api/blob/6abfc64139ade5a225ba62d675117c510885f1e2/controllers/external/tracker.go#L56-L60_)

emmjohnson: Interesting...

martyspiewak (approved):

PR #319, opened 2021-11-09
RFC-0015 Supply Chain selection with traits

squeedee:


rfc/rfc-0015-supply-chain-selection-through-traits.md
```diff
@@ -1,3 +1,157 @@
...
+## Summary
+
+A mechanism to allow a cluster to have multiple Supply Chains declared
+1. With similar purposes (or where one extends the behaviour of another)
+2. That will match the developer's requirements without the developer needing to understand naming permutations.
+
+
+```gherkin
+Given Supply Chains that represent permutations of a behaviour tree
+When A developer annotates their workload to select for a permutation
```
jwntrs, line 12: Can we frame this as "When a developer describes characteristics of their workload". They shouldn't be "selecting" a permutation.

rfc/rfc-0015-supply-chain-selection-through-traits.md
```diff
@@ -1,3 +1,157 @@
...
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
+
+```
+
+and `workload/input: image` would have:
+```yaml
+spec:
+  image: https://registry.com/my/image.tgz
```
sclevine:
```suggestion
  image: example.com/my/image
```

rfc/rfc-0015-supply-chain-selection-through-traits.md
```diff
@@ -1,3 +1,157 @@
...
+    * `agent`: an app that monitors other services
+
+* `input`:
+    * `image`: an app that is provided as a OCI image
+    * `source`: an app with source
+
+* `test`:
+    * `tekton`: the devloper has a tekton pipeline that can be used to test the code
+
+* `scan`:
```
sclevine, line 31: Suggest we remove scanning, given that it's more of a policy than a capability. Developers shouldn't be able to enable/disable it.

Eventually, we could create validating/mutating admission webhooks for Workloads that would allow this feature to be used with policies.

waciumawanjohi:
I think it makes sense to start with characteristics approach. I heard others in the community meeting that agreed.

waciumawanjohi:
The features option sounds a bit like the [Switch through fall-back strategy in RFC 9](https://github.com/vmware-tanzu/cartographer/blob/rfc-0009-supply-chain-switches-template-on-flag/rfc/rfc-0009-supply-chain-switches-template-on-flag.md#switch-through-fall-back-strategy), but falling back on whole supply-chain matches rather than on template matches.

waciumawanjohi (changes requested):
Req from IPM: Update RFC to represent what's been done for 1.0 and the other pieces can be moved into another RFC.

scothis:
This may be my history in web development speaking, but I lean hard in favor of the second option. Feature detection is almost always preferable to manual special casing. Doubly so when combined with progressive enhancement. If the Workload specifies source build it; if it specifies an image, skip the build; if it specifies a test pipeline, run it.

rfc/rfc-0015-supply-chain-selection-through-traits.md
squeedee: Definitely not a best practice, but not illegal in any way, and it provided a concrete example of how combinatorial complexity can occur, and be reduced by the RFC

squeedee:
| Req from IPM: Update RFC to represent what's been done for 1.0 and the other pieces can be moved into another RFC.

This is done now @waciumawanjohi ; See last commit. - can I get an approval?

rfc/rfc-0015-supply-chain-selection-through-traits.md
```diff
@@ -1,3 +1,142 @@
...
+* `web-image-test`
+* `web-image-test-scan`
+* etc...
+
+Further, the developer would need to know that `web-image-test-scan` is valid, not `web-image-test-scan`  
+
+## Possible Solutions
+
+
+### 1. Characteristics
```
waciumawanjohi:
```suggestion
## Solution

### Select on characteristics
```

rfc/rfc-0015-supply-chain-selection-through-traits.md
```diff
@@ -1,3 +1,142 @@
...
+    ```
+
+where the rule is:
+
+1. A workload is selected by the supply chain that matches ALL label selectors.
+2. Where a workload matches more than one supply chain, the supply chain with the larger list
+   of matched labels wins
+3. Where there is multiple supply chains with the largest number of matches, an error is reported
+
+### 2. Selecting on fields
```
waciumawanjohi:
```suggestion
## Alternatives/Extensions

### Selecting on fields
```

waciumawanjohi (approved):

PR #333, opened 2021-11-11
hack: use double space in ytt checksum check

cirocosta:
even though standard gnu coreutils accepts taking a single space as the
separator, that's not true for busybox.

by making sure we're passing a double space everywhere we compute a
checksum, we """guarantee""" to be good not only w/ gnu, but also
busybox-based ones (like alpine etc).

```c
				filename_ptr = strstr(line, "  ");          // double space
				/* handle format for binary checksums */
				if (filename_ptr == NULL) {
					filename_ptr = strstr(line, " *");
```

_(see https://github.com/mirror/busybox/blob/15f7d618ea7f8c3a0277c98309268b709e20d77c/coreutils/md5_sha1_sum.c#L296)_

ps.: in case you're wondering why this isn't captured by CI, it's because
we have github running based of ubuntu:

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
```

### DIFF
--- a/hack/ci/install-binaries.sh
+++ b/hack/ci/install-binaries.sh
@@ -54,7 +54,7 @@ install_ko() {
         local fname=ko_${KO_VERSION}_Linux_x86_64.tar.gz
 
         curl -sSOL $url
-        echo "${KO_CHECKSUM} $fname" | sha256sum -c
+        echo "${KO_CHECKSUM}  $fname" | sha256sum -c
         tar xzf $fname
 
         install -m 0755 ./ko /usr/local/bin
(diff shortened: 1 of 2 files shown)


emmjohnson (approved):

emmjohnson:
How to ensure we wont have a regression here?

waciumawanjohi:
> How to ensure we wont have a regression here?

We could [add a self-hosted busybox runner](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) and have our workflows randomly assign the job to the default runner or the self-hosted runner...

Note: Github discourages self-hosted runners for public repos.

cirocosta:
yeah that's a good point - internally we make use of it in GitLab, so there's _some_ form of coverage there that would surface here for sure

PR #341, opened 2021-11-15
Update missed references in Delivery RFC

sclevine:
Signed-off-by: Stephen Levine <slevine@vmware.com>


sclevine:
Note: we should decide if it makes sense to update RFCs once they are merged. Ideally, future changes are captured in documentation/specification, and not by updating past proposals.

waciumawanjohi (approved):

PR #346, opened 2021-11-16
Determine if error is unhandled or handled in the reconciler

emmjohnson:


pkg/controller/deliverable/reconciler.go
```diff
@@ -116,10 +120,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			if err != nil {
 				r.logger.Error(err, "dynamic tracker watch")
 			}
+			err = controller.NewUnhandledError(err)
```
emmjohnson: this isn't quite right in the for loop

pkg/controller/deliverable/reconciler.go
```diff
@@ -131,26 +136,21 @@ func (r *Reconciler) completeReconciliation(deliverable *v1alpha1.Deliverable, e
...
+			return ctrl.Result{}, fmt.Errorf("update deliverable status: %w", updateErr)
 		}
 	}
 
-	if err != nil {
+	if err != nil && controller.IsUnhandledError(err) {
 		return ctrl.Result{}, err
 	}
 
+	r.logger.Info("handled error", "error", err)
```
cirocosta: should this be inside a branch for `if err != nil`?

e.g.

```go
if err != nil {
  if controller.IsUnhandledError(err) {
    return ctrl.Result{}, err
  }
  r.logger.Info("handled error", "error", err)
}
```

otherwise I imagine we'd be printing `handled error: nil` quite a bit

e.g., with a properly working deliverable:

```json
{
  "level": "info",
  "ts": 1637161198.9989965,
  "logger": "controller.deliverable",
  "msg": "handled error",
  "name": "test",
  "namespace": "default",
  "error": null
}
```

pkg/controller/unhandled.go
```diff
@@ -0,0 +1,49 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import (
+	"errors"
+)
+
+func NewUnhandledError(err error) error {
```
cirocosta, line 24: it might be worth leaving a comment that describes what "Unhandled" means in the context of `controller` given how generic the notion of "handled" can be - wdyt?

cirocosta (approved):
lgtm! just left two minor comments

pkg/controller/unhandled.go
emmjohnson: added message

pkg/controller/deliverable/reconciler.go
emmjohnson: good catch

PR #350, opened 2021-11-17
Cartographer accepts flag for log verbosity

waciumawanjohi:


pkg/logger/logger.go
```diff
@@ -0,0 +1,29 @@
+package logger
+
+import (
+	"fmt"
```
cirocosta, line 18: whopsy, missing a newline to separate stdlib vs other imports

```suggestion
	"fmt"

```

pkg/logger/logger.go
```diff
@@ -0,0 +1,29 @@
...
+	"go.uber.org/zap/zapcore"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"strings"
+)
+
+func SetLogLevel(logLevel string) (zap.Opts, error) {
+	var level zapcore.Level
+	switch strings.ToUpper(logLevel) {
+	case "TRACE":
```
emmjohnson: Are we sure we want to support trace? We have only talked about debug and better info

pkg/logger/logger_test.go
```diff
@@ -0,0 +1,170 @@
...
+				Expect(logger.V(loglevel).Enabled()).To(Equal(expected))
+			},
+			Entry("info", 0, true),
+			Entry("debug", 1, false),
+			Entry("trace", 2, false),
+		)
+
+		It("outputs info logs", func() {
+			aNicePhrase := "It's a beautiful day in the neighborhood"
+			logger.V(0).Info(aNicePhrase)
```
emmjohnson: I don't think we should specify a verbosity on the logger itself. logger.Info should log for Info and Debug logs. logger.Debug should log for Debug.

emmjohnson: If this is not available with current interface, I think we should write our own

waciumawanjohi: I think this is a fine idea and one that I've considered as well. I think that's a different story. But yeah, we would just have our own wrapper of the `logr.logger` interface that has two loggers, the original and then one that had the verbosity increased to V1.

pkg/logger/logger.go
waciumawanjohi: Linting!

pkg/logger/logger.go
waciumawanjohi: As per the roman vote, trace has been expurgated from our code!

waciumawanjohi:
Dropping in for context: [Logr prefers V levels to the standard log nouns](https://github.com/go-logr/logr#why-v-levels)

pkg/logger/logger_test.go
waciumawanjohi: The logr README has an example of using `.V(int).Info`: https://github.com/go-logr/logr#but-i-really-want-to-use-a-format-string

cirocosta:
> _@waciumawanjohi Dropping in for context: Logr prefers V levels to the standard log nouns_

one thing to note is that the numbers that we set w/ zap are not necessarily 1:1 with the verbosity we use in the code itself (`zap` makes use of those negative numbers to indicate increased verbosity)

e.g., consider the following:

```go
package main

import (
        "flag"

        "go.uber.org/zap/zapcore"
        ctrl "sigs.k8s.io/controller-runtime"
        "sigs.k8s.io/controller-runtime/pkg/log"
)

var level = flag.Int("level", 0, "log level to pass to zap")

func main() {
        flag.Parse()

        ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zap.Options{
                Level: zapcore.Level(int8(*level)),
        })))

        println("level ", *level)

        l := log.Log.WithName("cartographer")
        l.Info("Info")
        l.V(0).Info("v0 Info")
        l.V(1).Info("v1 Info")
        l.V(2).Info("v2 Info")
}
```

running it with different levels, we get:

```console
$ go run main.go -level=-2
level  -2
{"level":"info","ts":1637162541.85724,"logger":"cartographer","msg":"Info"}
{"level":"info","ts":1637162541.8572812,"logger":"cartographer","msg":"v0 Info"}
{"level":"debug","ts":1637162541.857287,"logger":"cartographer","msg":"v1 Info"}
{"level":"Level(-2)","ts":1637162541.8572912,"logger":"cartographer","msg":"v2 Info"}

$ go run main.go -level=-1
level  -1
{"level":"info","ts":1637162544.5251677,"logger":"cartographer","msg":"Info"}
{"level":"info","ts":1637162544.5252118,"logger":"cartographer","msg":"v0 Info"}
{"level":"debug","ts":1637162544.5252295,"logger":"cartographer","msg":"v1 Info"}

$ go run main.go -level=0
level  0
{"level":"info","ts":1637162548.4194493,"logger":"cartographer","msg":"Info"}
{"level":"info","ts":1637162548.4195154,"logger":"cartographer","msg":"v0 Info"}

$ go run main.go -level=1
level  1
```

https://github.com/uber-go/zap/blob/eaeb0fc72fd23af7969c9a9f39e51b66827507ca/zapcore/level.go#L31-L56

cirocosta:
on a personal note, I'm +1 on going with number-based verbosity levels for allowing folks to filter out based on "desired chattiness" (i.e., sticking to `logger.v(chattiness).info()`) but making that number match what you use in the cli to configure it.

note that `Error` would not use `.V()` as it's always surfaced regardless of the log-level. e.g., modify the example above to have

```go
        err := fmt.Errorf("foo")
        l.Error(err, "Error")
        l.V(0).Error(err, "v0 Error")
        l.V(1).Error(err, "v1 Error")
        l.V(2).Error(err, "v2 Error")
```

and then notice that regardless of the level (e.g., `1` that would log nothing for `Info`) you always get all four error logs

waciumawanjohi:
> running it with different levels, we get:

Yup! This is reflected in the logger tests.

waciumawanjohi:
> one thing to note is that the numbers that we set w/ zap are not necessarily 1:1 with the verbosity we use in the code itself (zap makes use of those negative numbers to indicate increased verbosity)

It's `1 : -1`
😄

waciumawanjohi:
> I'm +1 on going with number-based verbosity levels for allowing folks to filter out based on "desired chattiness" (i.e., sticking to logger.v(chattiness).info()) but making that number match what you use in the cli to configure it.

I would be more inclined to have users specify a V level if we were going to use more than `Info` (V0) and `Debug` (V1). As the examples you posted above demonstrate, the call
```
        logger.V(1).Info("v1 Info")
```

Outputs:
```
{"level":"debug","ts":1637162541.857287,"logger":"cartographer","msg":"v1 Info"}
```
Note the level field `"level":"debug"`

It is only at higher logger verbosity levels that this is translated to `"level":"-INT"`

waciumawanjohi:
> note that Error would not use .V() as it's always surfaced regardless of the log-level. e.g., modify the example above to have

It is possible to set the log level above the Error log level. E.g. If you set the `zapcore.Level(2)` (something akin to `Fatal`), calls to l.Error(err, "Error") will not show up in the log output.

Not something that is very useful, but interesting to note.

cirocosta:
oh, interesting 👍 

so ultimately, what do you think is the best approach to move on with here? is there a particular point that you'd like to get a better idea from more people of what they think?

emmjohnson (approved):

PR #344, opened 2021-11-15
docs/examples: add runnable

cirocosta:
hey folks,

it's been long overdue having Runnable and ClusterRunTemplate documented and
part of the example, benefitting us not only in terms of having docs, but also
ensuring that in an e2e we don't regress as the example is ran on every commit.

please let me know if you believe there's a better way that we could frame those
resources and/or anything else that we could improve here.

thanks!

closes #287

## Changes proposed by this PR

- docs: add two entries to the `reference` page
    - runnable
    - clusterruntemplate

- examples: update the e2e to include a test before building an image
    - now: `source-provider <- source-tester <- image-builder <- deployer`

## Release Note

Add example and documentation of Runnable and ClusterRunTemplate.

## PR Checklist

Note: Please do not remove items. Mark items as done `[x]` or use ~strikethrough~ if you believe they are not relevant

- [x] Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`
- [x] Removed non-atomic or `wip` commits
- [x] Filled in the [Release Note](#Release-Note) section above
- [ ] Modified the docs to match changes

PR #351, opened 2021-11-17
Update docs to latest version

davidmirror-ops:
Signed-off-by: David Espejo <daespejo@vmware.com>

Also in this PR, I created a `main` folder for unreleased version docs

Your comments/thoughts/reviews are welcome


site/config.yaml
```diff
@@ -57,8 +57,10 @@ params:
   docs_search_index_name: index_name
   docs_search_api_key: api_key
   docs_versioning: true
-  docs_latest: v0.0.6
+  docs_latest: v0.0.7
   docs_versions:
+    - main
```
cirocosta: wdyt of `latest`? I think `main` indeed makes sense from the perspective of "versioning is based on git references", but when you look at the website, I think the concept of "latest version" .. "version 0.0.6 .." makes more sense - wdyt? 

---

(regardless, great to now be able to select the most up to date version!)

site/data/docs/main-toc.yml
```diff
@@ -0,0 +1,46 @@
...
+  - title: Examples
+    subfolderitems:
+      - page: Examples
+        url: /examples
+  - title: Documentation Style Guide
+    subfolderitems:
+      - page: Style Guide
+        url: /style-guide
+
+  #     - page: Scope 2
```
cirocosta: looking at the others, it seems like this came straight from what the others had too - should we clean it from this new one right now or do you think we should do some cleanup in a separate PR?

site/config.yaml
cirocosta: feedback from community call: 

- `latest` could instead be used as a pointer to the latest released version (analogous to how https://github.com/vmware-tanzu/cartographer/releases/latest points at whatever latest released tag we have)
- maybe call this .. `preview`? maybe `develpment`?

davidmirror-ops: My vote is for `development` as it is clear for users that it refers to unreleased software on which the team is actively working. For me as user, `latest` refers to the most recently released and stable version, and `preview` sounds like it will eventually see the light but not sure when :)

Please upvote your options, as soon as we have consensus, I can update the PR

site/data/docs/main-toc.yml
davidmirror-ops: yes, it's a copy from the other tocs. Why should we clean this one? Do we want to have a different ToC for the unreleased version?

site/config.yaml
emmjohnson: @squeedee what did we come up with in the docs meeting?

squeedee: "Edge" IIRC.

squeedee: 100% agree "Latest" is latest-known-good.. I'm also ok with "Development"

emmjohnson: Another favorite from community meeting was Development (putting here so people can vote)

cirocosta: +1 on "development", and "latest" for pointing at the latest release (if that's possible)

site/data/docs/main-toc.yml
cirocosta: oh noo, no worries, just trying to figure out whether the comments were something we wanted to clean up in this PR or elsewhere (given that, being comments, they seem like leftovers from experimenting at some point)

davidmirror-ops: I'm kinda used to addressing comments in the same PR but not an expert here so, the idea would be to have a separate PR addressing how to handle unreleased vs latest-stable naming only?

cirocosta: I personally would go for a separate one where we get rid of the commented sections, but no hard opinions on my side 😁  if you feel like deleting here in this one, all good!

site/config.yaml
davidmirror-ops: Updated PR with "development" for unreleased version docs. Using `latest` along the most recent version number would add at least two additional steps to the process whenever we need to update for a new release

cirocosta:
lgtm! there are a couple conflicts at the moment, but that's really it

```
site/content/docs/_index.md
site/content/docs/about.md
site/content/docs/install.md
site/content/docs/reference.md
```

are you familiar with how to resolve those, @davidmirror-ops ? I think those conflicts are coming mostly from #344. if you need any help, just let me know!

thanks!

davidmirror-ops:
Hey @cirocosta! Well not sure how to fix those, but I can try (so I learn) and if I can't, will request help :)

davidmirror-ops:
soo I tried it with a draft PR but ended up with repeated commits in the log so it seems I need help :)

squeedee:
Closing with [#403 as replacement](https://github.com/vmware-tanzu/cartographer/pull/403)

PR #352, opened 2021-11-17
Handle errors workload

emmjohnson:


pkg/controller/workload/reconciler.go
```diff
@@ -64,58 +64,60 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 			r.conditionManager.AddPositive(TemplateObjectRetrievalFailureCondition(typedErr))
+			err = controller.NewUnhandledError(err)
 		case realizer.StampError:
 			r.conditionManager.AddPositive(TemplateStampFailureCondition(typedErr))
 		case realizer.ApplyStampedObjectError:
 			r.conditionManager.AddPositive(TemplateRejectedByAPIServerCondition(typedErr))
+			err = controller.NewUnhandledError(err)
 		case realizer.RetrieveOutputError:
 			r.conditionManager.AddPositive(MissingValueAtPathCondition(typedErr.ResourceName(), typedErr.JsonPathExpression()))
+			//err = controller.NewUnhandledError(err)
```
cirocosta: whoopsy / intentional? 🤔

pkg/controller/workload/reconciler.go
```diff
@@ -124,26 +126,23 @@ func (r *Reconciler) completeReconciliation(ctx context.Context, workload *v1alp
 		workload.Status.ObservedGeneration = workload.Generation
 		updateErr = r.Repo.StatusUpdate(workload)
 		if updateErr != nil {
-			logger.Error(updateErr, "update error")
-			if err == nil {
-				return ctrl.Result{}, fmt.Errorf("update workload status: %w", updateErr)
-			}
+			return ctrl.Result{}, fmt.Errorf("update workload status: %w", updateErr)
```
cirocosta, line 128: it seems like the behavior changed a bit here now - it seems like we'd lose the chance of logging the original `err` in case we had one given that we now return from here regardless of whether `err` was nil or not 🤔  is that right? 

i'm wondering of the following scenario

```
err = somethingBad
updateErr = statusUpdateFailure

if updateErr != nil {
  return .., updateErr
}

// lost the chance of letting people know of "somethingBad"
```

that seems to be the case as opposed to what we had before where we'd end up logging both as we'd do


```
err = somethingBad
updateErr = statusUpdateFailure

if updateErr != nil {
  log updateErr
  if err != nil {
    return .., err
  }
}
```

where we'd end up with both

pkg/controller/workload/reconciler.go
```diff
@@ -162,18 +161,19 @@ func (r *Reconciler) getSupplyChainsForWorkload(workload *v1alpha1.Workload) (*v
 	}
 
 	supplyChains, err := r.Repo.GetSupplyChainsForWorkload(workload)
-	if err != nil || len(supplyChains) == 0 {
+	if err != nil {
+		return nil, controller.NewUnhandledError(fmt.Errorf("get supply chain by label: %w", err))
```
cirocosta: wdyt of

```suggestion
		return nil, controller.NewUnhandledError(fmt.Errorf("get supplychains for workload: %w", err))
```

?

the fact that it uses labels seems to be a detail that would be surfaced by the error message within the method (rather than at this level)

cirocosta (commented):
overall, lgtm! left a couple minor questions

pkg/controller/workload/reconciler.go
emmjohnson: This was intentional. @squeedee and I decided that it would be okay if we lost the original error. We requeue here and assume the system will sort it self out.

pkg/controller/workload/reconciler.go
emmjohnson: We return an unhandled error in one case

pkg/controller/workload/reconciler.go
cirocosta: got it, makes sense 👍  thx!

pkg/controller/workload/reconciler.go
cirocosta: sorry, wdym? I'm thinking more about the content of the messages - rather than `get supply chain by label`, being more high level with `get supplychains for workload` so that it's up to the lower level (actual `repository.GetSupplyChainsForWorkload`) to let us know the details (the fact that it's by label, or that ... say .. it didn't find a "best match" in the future).

emmjohnson: ahhh, I cannot read. I'll make the change. I think the entire API is incorrect, but I'll handle that separately.

pkg/controller/workload/reconciler.go
emmjohnson: WHOOPSY

cirocosta (approved):

PR #355, opened 2021-11-17
Handle errors for runnable

emmjohnson:


pkg/controller/runnable/reconciler.go
```diff
@@ -55,23 +58,63 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
```
cirocosta, line 58: maybe unrelated, but i noticed that in https://github.com/vmware-tanzu/cartographer/pull/358 we moved towards checking for `notfound` in the repository itself rather than at this level - should we include a change for that here too?

pkg/controller/runnable/reconciler.go
```diff
@@ -55,23 +58,63 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	}
 
-	condition, outputs, stampedObject := r.Realizer.Realize(ctx, runnable, logger, r.Repo)
+	if runnable == nil {
```
cirocosta: maybe unrelated, but i noticed that in https://github.com/vmware-tanzu/cartographer/pull/358 we moved towards checking for `notfound` in the repository itself rather than at this level - should we include a change for that here too?

with the current code, I think this check for `nil` wouldn't make sense (unless we do the `if isnotfound()..` there in repository)

pkg/realizer/runnable/errors.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type ApplyStampedObjectError struct {
+	Err           error
+	StampedObject *unstructured.Unstructured
+}
+
+func (e ApplyStampedObjectError) Error() string {
+	return fmt.Errorf("unable to apply stamped object '%s/%s': '%w'",
+		e.StampedObject.GetNamespace(), e.StampedObject.GetName(), e.Err).Error()
```
cirocosta: minor detail - most objects will actually _not_ have `metadata: {name: <>}` set, but instead, `metadata: {generateName: <something>-}`, so this message will actually not be very useful 😁  

e.g.:

> _unable to apply stamped object **'default/':** 'create: admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot_

cirocosta (approved):
lgtm! just to minor comments around:

- apply same pattern of checking for notfound in repository so it's all alike
- error for stamped object using `name` rather than `generateName` (maybe we can go with checking emptiness for `Name` and then, if not found, looking at `GenerateName`?)

pkg/controller/runnable/reconciler.go
emmjohnson: ya this is just a weird diff i believe

Issue #123, opened 2021-09-17
Reorder github workflow


emmjohnson:
Order should be:
- install go
- check copyright
- lint
- everything else

cirocosta:
done @ https://github.com/vmware-tanzu/cartographer/pull/87

PR #359, opened 2021-11-18
repository: use "best match" for supplychain/delivery selection

### LINKED ISSUE #298: Best match multiple labels on workload
### Description of problem

When matching a SupplyChain to a Workload, we currently require that all labels on the supply chain also exist on the workload, but we error if we end up matching more than 1 supply chain.

The labels on a workload should represent characteristics of the workload. When a particular workload gets used across clusters, if a given cluster doesn't have a supply chain matching one of the labels, then the workload should still run.

### Proposed solution

A workload still needs to match all the labels from a supply chain, because if a supply chain runs a "test" step, the workload needs to fulfill the params necessary to run that step. 

However, if we match multiple supply chains, then we should take the supply chain with the most labels. If there is more than one supply chain with the most labels, then we should error.

RFC #319

cirocosta:

1. implement a function (`BestLabelMatches`) that, given something with labels and a set of objects that have selectors, gives the subset of the latter that "best matches" that initial label set according to [RFC-0015].

2. make use of such function for both delivery and workload

[RFC-0015]: https://github.com/vmware-tanzu/cartographer/pull/319

closes #298 

## Release Note

Allow users to have their Workload and Deliverables selecting ClusterSupplyChain and ClusterDelivery (accordingly) based on multiple labels rather than exact match.

## PR Checklist

### DIFF
--- a/pkg/controller/deliverable/reconciler_test.go
+++ b/pkg/controller/deliverable/reconciler_test.go
@@ -197,7 +197,7 @@ var _ = Describe("Reconciler", func() {
 					},
 				},
 			}
-			repo.GetDeliveriesForDeliverableReturns([]v1alpha1.ClusterDelivery{delivery}, nil)
+			repo.GetDeliveriesForDeliverableReturns([]*v1alpha1.ClusterDelivery{&delivery}, nil)
 			stampedObject1 = &unstructured.Unstructured{}
 			stampedObject1.SetGroupVersionKind(schema.GroupVersionKind{
 				Group:   "thing.io",
@@ -270,7 +270,7 @@ var _ = Describe("Reconciler", func() {
 						Message: "some informative message",
 					},
 				}
-				repo.GetDeliveriesForDeliverableReturns([]v1alpha1.ClusterDelivery{delivery}, nil)
+				repo.GetDeliveriesForDeliverableReturns([]*v1alpha1.ClusterDelivery{&delivery}, nil)
 			})
 
 			It("does not return an error", func() {
@@ -640,7 +640,7 @@ var _ = Describe("Reconciler", func() {
 				Version: "alphabeta1",
 				Kind:    "MyThing",
 			})
-			repo.GetDeliveriesForDeliverableReturns([]v1alpha1.ClusterDelivery{delivery, delivery}, nil)
+			repo.GetDeliveriesForDeliverableReturns([]*v1alpha1.ClusterDelivery{&delivery, &delivery}, nil)
 		})
 
 		It("does not return an error", func() {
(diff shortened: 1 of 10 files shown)


pkg/controller/deliverable/reconciler_test.go
```diff
@@ -268,7 +268,7 @@ var _ = Describe("Reconciler", func() {
 						Message: "some informative message",
 					},
 				}
-				repo.GetDeliveriesForDeliverableReturns([]v1alpha1.ClusterDelivery{delivery}, nil)
+				repo.GetDeliveriesForDeliverableReturns([]*v1alpha1.ClusterDelivery{&delivery}, nil)
```
emmjohnson, line 273: indentation

pkg/controller/workload/reconciler_test.go
```diff
@@ -267,7 +267,7 @@ var _ = Describe("Reconciler", func() {
 						Message: "some informative message",
 					},
 				}
-				repo.GetSupplyChainsForWorkloadReturns([]v1alpha1.ClusterSupplyChain{supplyChain}, nil)
+				repo.GetSupplyChainsForWorkloadReturns([]*v1alpha1.ClusterSupplyChain{&supplyChain}, nil)
```
emmjohnson, line 269: indentation

pkg/controller/deliverable/reconciler_test.go
cirocosta: sorry, wdym? is the suggestion to split the args into multiple lines? the change here was solely making `clustedelivery` a pointer

cirocosta:
force push: squashed & rebase

cirocosta:
working on fixing the conflicts

cirocosta:
force-push: rebase + fix conflicts

pkg/controller/deliverable/reconciler_test.go
emmjohnson: it was showing all the way to left before? maybe bad diff?

pkg/repository/repository.go
```diff
@@ -214,23 +214,27 @@ func (r *repository) patchUnstructured(ctx context.Context, existingObj *unstruc
...
 	if err := r.cl.List(ctx, list); err != nil {
 		return nil, fmt.Errorf("list supply chains: %w", err)
 	}
 
-	var clusterSupplyChains []v1alpha1.ClusterSupplyChain
-	for _, supplyChain := range list.Items {
-		if selectorMatchesLabels(supplyChain.Spec.Selector, workload.Labels) {
-			clusterSupplyChains = append(clusterSupplyChains, supplyChain)
-		}
+	selectorGatters := []SelectorGetter{}
```
emmjohnson: selectorGetters instead of Gatters?

emmjohnson (approved):

PR #377, opened 2021-11-19
Enhanced logging for deliverable

emmjohnson:
Co-authored-by: Rasheed Abdul-Aziz<rabdulaziz@vmware.com>


kuttl-test.yaml
```diff
@@ -23,6 +23,6 @@ crdDir: config/crd/bases
 manifestDirs:
   - tests/resources/crds
 commands:
-  - command: build/cartographer
+  - command: build/cartographer --log-level debug
```
waciumawanjohi: 👍 

Separate future chore: I want to figure out how to pass a flag to the make test commands to quiet the output of kuttl tests. Indeed, that might point to a reason to include a `--log-level fatal` so that even the error logs are suppressed.

pkg/controller/deliverable/reconciler.go
```diff
@@ -66,22 +70,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			fmt.Errorf("failed to get object gvk for delivery [%s]: %w", delivery.Name, err)))
 	}
 
 	deliverable.Status.DeliveryRef.Kind = deliveryGVK.Kind
 	deliverable.Status.DeliveryRef.Name = delivery.Name
 
 	if !r.isDeliveryReady(delivery) {
 		r.conditionManager.AddPositive(MissingReadyInDeliveryCondition(getDeliveryReadyCondition(delivery)))
-		return r.completeReconciliation(ctx, deliverable, fmt.Errorf("delivery is not in ready state"))
+		log.Info("delivery is not in ready state")
```
waciumawanjohi, line 88: We pass an error to `completeReconciliation`. Should this be `log.Error`

martyspiewak:
lgtm!

pkg/controller/deliverable/reconciler.go
emmjohnson: No, because this is a handled error

pkg/controller/deliverable/reconciler.go
```diff
@@ -41,21 +42,24 @@ type Reconciler struct {
...
 
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.logger = logr.FromContext(ctx)
-	r.logger.Info("started")
-	defer r.logger.Info("finished")
+	log := logr.FromContext(ctx)
+	log.Info("started")
+	defer log.Info("finished")
+
+	log = log.WithValues("deliverable", req.NamespacedName)
```
cirocosta, line 52: do we need this? controller-runtime is already adding `name` and `namespace` to the object

e.g., with this approach, we have:

```
{
  "level": "info",
  "ts": 1637596812.8353918,
  "logger": "controller.delivery",
  "msg": "started",
  "name": "delivery",
  "namespace": ""
}

{
  "level": "info",
  "ts": 1637596812.8363948,
  "logger": "controller.deliverable",
  "msg": "started",
  "name": "test",
  "namespace": "default"
}

{
  "level": "info",
  "ts": 1637596812.8366206,
  "logger": "controller.deliverable",
  "msg": "delivery is not in ready state",
  "name": "test",
  "namespace": "default",
  "deliverable": "default/test",
  "delivery": "delivery"
}
```

```go
// Reconcile implements reconcile.Reconciler.
func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, err error) {
//...
	log := c.Log.WithValues("name", req.Name, "namespace", req.Namespace)
	ctx = logf.IntoContext(ctx, log)
	return c.Do.Reconcile(ctx, req)
}
```

from https://github.com/kubernetes-sigs/controller-runtime/blob/4d10a0615b11507451ecb58bfd59f0f6ef313a29/pkg/internal/controller/controller.go#L112-L114

kuttl-test.yaml
squeedee: actually we probably dont want to leave this in. It's simple enough to change when you need it. To your question, i think the only ways are to pass a different kuttl-test.yaml OR use env vars. If we have a `CARTO_LOG_LEVEL` set in our shell (or by make) it _should_ pass all the way into the binary.

pkg/controller/deliverable/reconciler.go
```diff
@@ -41,21 +42,24 @@ type Reconciler struct {
...
+	ctx = logr.NewContext(ctx, log)
 
 	deliverable, err := r.Repo.GetDeliverable(ctx, req.Name, req.Namespace)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("get deliverable: %w", err)
+		log.Error(err, "failed to get deliverable")
+		return ctrl.Result{}, fmt.Errorf("failed to get deliverable [%s]: %w", err, req.NamespacedName)
```
cirocosta: I don't think we usually report like this (`[..]`) - is that something we want to adopt? if so, it might be worth communicating the pattern somewhere? perhaps as part of making logging standardized, we could have an entry in CONTRIBUTING where we write down decisions like this. wdyt?

e.g., before this PR we had no matches for this type of error format:

```console
$ ag --stats "\[.*\%s\]" ./pkg
0 matches
0 files contained matches
```

as we used to mostly do like these matches

```
pkg/registrar/map_functions.go
72:             mapper.Logger.Error(err, fmt.Sprintf("could not get GVK for template: %s", templateName))
203:            return fmt.Errorf("missing apiVersion or kind: %s err: %w", obj.GetName(), err)
207:            return fmt.Errorf("unversioned object: %s", obj.GetName())
211:            return fmt.Errorf("unexpected GVK count: %s", obj.GetName())
253:            mapper.Logger.Error(err, fmt.Sprintf("could not get GVK for template: %s", templateName))
```

but now, we have a couple in this new one

```console
$ ag --stats "\[.*\%s\]" ./pkg
22 matches
7 files contained matches
133 files searched
720289 bytes searched
0.006962 seconds
```

pkg/controller/deliverable/reconciler.go
```diff
@@ -109,16 +120,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			err = controller.NewUnhandledError(err)
 		}
 	} else {
+		if log.V(logger.DEBUG).Enabled() {
```
cirocosta, line 123: tbh, this loop is soo cheap that I'd go without the branching .. but happy to hear from other folks too

waciumawanjohi:
On the concern for logging secrets; should we have another log-level that is more verbose and is explicit that it could expose secrets? `--log-level very-verbose-possibly-exposes-secrets` (or something more succinct)

pkg/controller/deliverable/reconciler.go
squeedee: I'll expound a little to share some thoughts. There are many kinds of 'error' (and I'm not happy about it but it's the truth):
* **an exception**:  "This code should not be reached, but no one likes a panic".   **Use**: Log.Error and retry-with-backoff
* **a recoverable error**.. async/external comms usually. **Use**: Log.Error and retry-with-backoff
* **a message** that we consider recoverable by user action:  "I tried to stamp but couldn't" (often these are deeper user validations). **Use**: Status.Conditions (primary form of user comms), Log.Info
* **a message** that we consider recoverable with a retry: "something is not ready yet". **Use**: Status.Conditions, Log.Info

log.Debug is not necessarily tied to errors, but is there to provide noisy context for debugging.

for `retry-with-backoff` we feel that it would also be great to add a Condition for Backoff to communicate with the user. (should be an issue/discussion though)

pkg/controller/deliverable/reconciler.go
squeedee: Our thoughts here were to match log messages with the context eg:

`log.Info("My deliverable is unhappy about widget", "widget", widget)`

The guidance from controller runtime is to not `fmt`  the message, but reference the keys you will find in the structure log. So mixed with the context, the above log line produces:

```
{
  "msg": "My deliverable is unhappy about widget",
  "deliverable": "default/my-deliverable",
  "widget": {
    "apiVersion": "hello.io/widgetctrlr",
    "kind": "Widget",
    "namespace": "my-ns",
    "name": "my-widget"
  }
}
```

if we only did 'namespace' and 'name' it would be less obvious what they refer too.

waciumawanjohi:
Will this fixme be part of this story?
https://github.com/vmware-tanzu/cartographer/blob/5df90e6daf8c73e75a26977bb1dbdfadc053b0c9/pkg/registrar/map_functions.go#L39

pkg/controller/deliverable/reconciler.go
cirocosta: I mean, controller-runtime by default will inject namespace and name, won't it? 🤔  my point was more that we're duplicating that information

pkg/controller/deliverable/reconciler.go
squeedee: Yes this needs intentional discussion. Our idea is to put any key or simple key-value information in `[]` and list/complex data after a `:`

Some people use backticks, but whatever we use it needs to be obvious we are refering to value types.

personally I like Emily's Square Brackets (as they will forever be known, ESB for short) and naturally we want to update contributing with this information (and many other bits) after we ratify it

kuttl-test.yaml
cirocosta: 👍 

https://github.com/peterbourgon/ff

makes that pretty easy for plain flags


```go
          if err := ff.Parse(
                  cmdFlagSet, os.Args[1:],
                  ff.WithEnvVarPrefix("CARTO"),
          ); err != nil {
                  fmt.Fprintln(os.Stderr, err)
                  os.Exit(1)
          }
```

pkg/realizer/deliverable/component.go
```diff
@@ -84,9 +89,13 @@ func (r *resourceRealizer) Do(ctx context.Context, resource *v1alpha1.ClusterDel
 		templatingContext["source"] = inputs.OnlySource()
 	}
 
+	// TODO: Could we be leaking secrets? Should we just print the keys
+	//log.V(logger.DEBUG).Info("using context for stamper", "context", templatingContext)
```
cirocosta: zap does provide a way of providing a custom marshaller for a given type

https://github.com/uber-go/zap/blob/eaeb0fc72fd23af7969c9a9f39e51b66827507ca/zapcore/marshaler.go#L23-L32

but, I'm not really sure we'd want to go that route - I particularly prefer the approach you mentioned (an extra level where you intentionally specify that you want this kind of logging)

cirocosta: apparently kubernetes does provide some sanitization for their own logs? 
 - https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/component-base/logs/sanitization/sanitization.go (sanitizaiton)
 - https://github.com/kubernetes/kubernetes/blob/1d8966f4f92429fc1c1073d121dca1cc3f53974a/staging/src/k8s.io/component-base/logs/options.go#L89-L91 (use of it as a filter)

pkg/controller/deliverable/reconciler.go
squeedee: Yes but it reads like this:
```json
{
  "msg": "My deliverable is unhappy",
  "namespace": "default",
  "name": "my-deliverable"
}
```


this is more verbose and redundant:
```json
{
  "msg": "My deliverable is unhappy",
  "namespace": "default",
  "name": "my-deliverable",
  "deliverable": "default/my-deliverable",
}
```

but now you can easily connect "my deliverable is unhappy" to "a thing"

pkg/controller/deliverable/reconciler.go
squeedee: it's not as cheap as fmt.Sprintf. Objects cost a little to serialize, and a _lot_ to serialize if logging dev_mode is enabled. I'd happily wait for profiling efforts too though

pkg/controller/deliverable/reconciler.go
cirocosta: +100 for making the "we're in a retry backoff" state somehow more explicit (ala when you see a Pod in CrashLoopBackoff, you know the retries will start taking longer)

pkg/controller/deliverable/reconciler.go
cirocosta: aaaaah, I see 👍 

imo there's still duplication though - `controller.deliverable` already lets you know it's a deliverable (it wouldn't reconcile against anything else)

squeedee:
> Will this fixme be part of this story?
> 
> https://github.com/vmware-tanzu/cartographer/blob/5df90e6daf8c73e75a26977bb1dbdfadc053b0c9/pkg/registrar/map_functions.go#L39

Probably not

waciumawanjohi:
From the code here, I surmise the rules of thumbs:
1. Write a debug log at the beginning of method calls the name of a method (and any reasonable context)
2. When checking for an error from a method call, if the error is found write an info log with relevant context
3. When creating an unhandled error, write an error log.

If those aren't correct, what are the rules of thumbs?

waciumawanjohi:
There are times when an error is created and we wait until we get to the reconciler to decide whether to wrap that error in an UnhandledError. If the rule from above holds, doesn't that mean that the error log will miss valuable information (the context that was only available in the method that created the error, which is effectively popped from the stack by the time we return to the reconcile method)?

waciumawanjohi:
I wonder if we should simplify by changing
`log.V(logger.DEBUG).Info("some message")`
to
`log.Debug("some message")`

That would entail creating our own logger interface that wraps logr.logger
It would also tie us more closely to using the debug noun vs V levels.
We would also need a method that was the equivalent of `log.V(logger.DEBUG).Enabled()`

squeedee:
> On the concern for logging secrets; should we have another log-level that is more verbose and is explicit that it could expose secrets? `--log-level very-verbose-possibly-exposes-secrets` (or something more succinct)

That's not the current problem. If we find we do want secrets emitted during debug then a story can be raised for that. 

The exigent problem, is that we do want enough information to aid debug and we don't have a sanitization mechanism in place. The stampingContext contains really useful 'facts' like "this field you're selecting for exists" and "this field you're selecting for contains the information you expected", it would be great to share it. As loathe as I am to sue something as extensive as [datapol](https://pkg.go.dev/k8s.io/component-base/logs/datapol) (I'm suspect of it's performance) it might be a good idea to use it judiciously for the 'stampingContext' today

waciumawanjohi:
> The exigent problem, is that we do want enough information to aid debug and we don't have a sanitization mechanism in place.

I was mooting a possible solution to our lack of sanitization: only log objects that would call for sanitization when the users have explicitly requested insecure logging.

waciumawanjohi:
It feels as if controller-runtime's choice of logr limits our ability to consume k8s work on sanitization.

There's a KEP regarding data sanitization:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-security/1753-logs-sanitization/README.md

And stories in klog for implementation:
https://github.com/kubernetes/klog/pull/183

squeedee:
> 1. Write a debug log at the beginning of method calls the name of a method (and any reasonable context)
> 2. When checking for an error from a method call, if the error is found write an info log with relevant context
> 3. When creating an unhandled error, write an error log.
> 
> If those aren't correct, what are the rules of thumbs?

I don't think these work. If you studiously follow these, you create more output. 
Once all the logging is done, the pair on this work may be able to come up with some guidelines?

IMHO, the hard to read errors and logging we've encountered until now, often with errors logged like "get: get thing: get thing: get ...." comes from a 'rule-of-thumb' approach. The way we tackled these logs is to consider what would happen if an error occurred, and ask what we would like to see logged. We then worked from the lowest levels of our application logic upward, because we realized, for example, that the reconciler did not need to log in depth, all the 'realizer error types', because extensive logging was already done at the 'tips' of the conditional behavior down in component.go.

squeedee:
> There are times when an error is created and we wait until we get to the reconciler to decide whether to wrap that error in an UnhandledError. If the rule from above holds, doesn't that mean that the error log will miss valuable information (the context that was only available in the method that created the error, which is effectively popped from the stack by the time we return to the reconcile method)?

Errors are not 1-1 with logs. At the least, we log useful information closest to the occurance of an error. Whereas the error communicates 'messaging' up the stack. If the information required to create a useful Condition is absent, then that would be a problem, otherwise the error logging that controller-runtime does on backoff is inconsequential (it's covered by log.error at the site of a significant issue)

squeedee:
> I was mooting a possible solution to our lack of sanitization: only log objects that would call for sanitization when the users have explicitly requested insecure logging.

I understand what you're suggesting now, but I feel like having sanitization on the StampContext so we can _always_ display it in Debug is much more desirable.

squeedee:
Rules of thumb:

1. When an error occurs, think about what debug will improve our ability to debug issues related to the error
2. Do not blindly add logs at every call site.
3. With values that will provide context in higher abstractions, add them as soon as you can, for example `"supplychain",supplychain` in the workload reconciler. Do not add them if they're only going to be useful locally (eg stampedObject in our reconcilers will mean nothing to callee's)
4. Place context on errors and describe the error in 'local' terms. Eg: for a get to the api server: "failed to get supply-chain from api server" but for the same error in the reconciler "failed to get supply-chain". There will still be a lot of repetition, but the changing context gives the duplicate messages a reason to exist.

pkg/realizer/deliverable/component.go
emmjohnson: Ive created a card on the board to do this for templatingcontext and outputs

squeedee:
> I wonder if we should simplify by changing `log.V(logger.DEBUG).Info("some message")` to `log.Debug("some message")`
> 
> That would entail creating our own logger interface that wraps logr.logger It would also tie us more closely to using the debug noun vs V levels. We would also need a method that was the equivalent of `log.V(logger.DEBUG).Enabled()`

normally I'd agree except:
1. people familiar with controller-runtime will expect to use log.v().info for higher level messages
2. we would also have to produce our own WithLogger NewContext etc:  https://pkg.go.dev/github.com/go-logr/logr@v0.4.0#FromContext

To me, that seems like a lot of changes to buck a 'barely annoying' convention?

martyspiewak (approved):

CONTRIBUTING.md
```diff
@@ -27,6 +27,33 @@ The Cartographer project team welcomes contributions from the community. If you
...
+- our reconcilers will mean nothing to callee's).
+- Place context on errors and describe the error in 'local' terms. Eg: for a get to the api server: "failed to get
+supply-chain from api server" but for the same error in the reconciler "failed to get supply-chain". There will still 
+be a lot of repetition, but the changing context gives the duplicate messages a reason to exist.
+
+### Error types
+There are many kinds of 'error':
+- an exception: "This code should not be reached, but no one likes a panic". Use: Log.Error and return an unhandled error 
+(retry-with-backoff)
+- a recoverable error: async/external comms usually. Use: Log.Error and return an unhandled erorr (retry-with-backoff)
```
idoru, line 52: Is it worth explicitly naming `NewUnhandledError` here?

PR #380, opened 2021-11-19
354 overridable params

### LINKED ISSUE #354: Overridable params
### Description of problem

Problem outlined here #77 

### Proposed solution

Let's move forward with the approach of having an `overridable: true|false` flag on the SupplyChain/Delivery. So the order of precedence for the params becomes:

```
supply chain component value (overridable: false)
supply chain value (overridable: false)
workload value
supply chain component value (overridable: true)
supply chain value (overridable: true)
template value
```

```
delivery component value (overridable: false)
delivery value (overridable: false)
deliverable value
delivery component value (overridable: true)
delivery value (overridable: true)
template value
```

waciumawanjohi:
Strategy to have overridable params.

closes #354
This is an implementation of #77, which still needs full ratification.

Allows params to be specified for an entire supply-chain as well as for individual resources. Either, both, or neither may be specified. Either may be marked as `overrideable`, indicating that the workload may override the value. If the resource level param is specified, it always clobbers both the supply-chain level param value and overrideable flag.

Introduces the noun `blueprint` as the generic noun for `supply-chain` and `delivery`. (E.g. a supply-chain is a blueprint)
Introduces the noun `order` as the generic noun for `workload` and `deliverable`. (E.g. a supply-chain is an order)

Kuttl params tests consolidated
Unit table test created

Ask: In `component_test.go` and `realizer_test.go` I simply added an empty params object to ensure that the calls matched the new signature. I assume that this is sufficient, but this assumption may warrant inspection.


config/crd/bases/carto.run_clusterdeliveries.yaml
```diff
@@ -48,6 +48,21 @@ spec:
             type: object
           spec:
             properties:
+              params:
```
cirocosta, line 51: we should update the reference docs given that now the CRD has a new field

pkg/realizer/workload/component.go
```diff
@@ -44,7 +44,7 @@ func NewResourceRealizer(workload *v1alpha1.Workload, repo repository.Repository
 	}
 }
 
-func (r *resourceRealizer) Do(ctx context.Context, resource *v1alpha1.SupplyChainResource, supplyChainName string, outputs Outputs) (*unstructured.Unstructured, *templates.Output, error) {
+func (r *resourceRealizer) Do(ctx context.Context, resource *v1alpha1.SupplyChainResource, supplyChainName string, supplyChainParams []v1alpha1.OverridableParam, outputs Outputs) (*unstructured.Unstructured, *templates.Output, error) {
```
waciumawanjohi: The supply-chain params may more properly belong in the instantiation of the resourceRealizer, as the one set is used in every iteration over the different resources.

config/crd/bases/carto.run_clustersupplychains.yaml
```diff
@@ -48,6 +48,21 @@ spec:
             type: object
           spec:
             properties:
+              params:
```
cirocosta, line 51: we should update the reference docs given that now the CRD has a new field (same for the other ones)

config/crd/bases/carto.run_clusterdeliveries.yaml
```diff
@@ -48,6 +48,21 @@ spec:
...
+                  properties:
+                    name:
+                      type: string
+                    overridable:
+                      type: boolean
+                    value:
+                      x-kubernetes-preserve-unknown-fields: true
+                  required:
+                  - name
+                  - overridable
```
cirocosta, line 62: do we want to make these required, or, could we make it possible to omit it?

cirocosta: e.g.,

```yaml
  kind: ClusterSupplyChain
  metadata:
    name: foo
  spec:
    selector:
      foo: "bar"
    resources:
      - name: source-provider
        templateRef:
          kind: ClusterSourceTemplate
          name: source-template
        params:
          - name: foo
            value: bar
```

fails as such:

```
error validating data: 
  ValidationError(ClusterSupplyChain.spec.resources[0].params[0]): 
     missing required field "overridable" in run.carto.v1alpha1.ClusterSupplyChain.spec.resources.params
```

squeedee:
I still suck at Github PR reviews.. The only issues I'm stuck on is:
* using `workload`/`supplychain` in generic situations
* passing the params into `Do` when I think it belongs on the receiver.
* Ciro's comment that we need to update the docs. This should be done before finalizing the merge

Everything else is just discussion.

config/crd/bases/carto.run_clusterdeliveries.yaml
waciumawanjohi: If we end up choosing this implementation, yes, let's have a default to false story.

jwntrs:
Closing in favour of https://github.com/vmware-tanzu/cartographer/pull/387

PR #387, opened 2021-11-23
Blueprint params specify default or value

waciumawanjohi:
Params on supply-chains and deliveries (and on the associated
resources) require either a `default` or `value`.

If the `default` field is specified, the provided value can be overriden
by the param supplied by the workload.

If instead the `value` field is specified, the param value supplied
by the workload is ignored.

If for the same named param, the blueprint specifies a value at
the top level and at an individual resource level, the specification
at the individual resource level will clobber the specification at the
top level.

[#354]


pkg/apis/v1alpha1/common.go
```diff
@@ -33,6 +33,27 @@ type Param struct {
...
 
+type DelegatableParam struct {
+	Name         string                `json:"name"`
+	Value        *apiextensionsv1.JSON `json:"value,omitempty"`
+	DefaultValue *apiextensionsv1.JSON `json:"default,omitempty"`
+}
+
+func (p *DelegatableParam) validateDelegatableParams() error {
+	if p.bothValuesSet() || p.neitherValueSet() {
+		return fmt.Errorf("invalid param: '%s', must set exactly one of value and default", p.Name)
```
emmjohnson, line 44: consider using brackets for param

pkg/templates/params.go
```diff
@@ -22,18 +22,55 @@ import (
 
 type Params map[string]apiextensionsv1.JSON
 
-func ParamsBuilder(defaultParams v1alpha1.DefaultParams, resourceParams []v1alpha1.Param) Params {
+func ParamsBuilder(
+	templateParams []v1alpha1.TemplateParam,
+	blueprintParams []v1alpha1.DelegatableParam,
```
emmjohnson, line 27: I don't know what blueprint is here. We should discuss terminology. I see that it is for supply chain/delivery from caller

emmjohnson (approved):

pkg/templates/params.go
waciumawanjohi: Point well taken. I had a callout about the new phrases in PR #354, but didn't copy that part over. I should have!

Rash opened a discussion on the Kubernetes slack about the naming: https://kubernetes.slack.com/archives/C02HKPSEKV1/p1637620952095700

site/content/docs/reference.md
```diff
@@ -141,6 +139,34 @@ spec:
   selector:
     app.tanzu.vmware.com/workload-type: web
 
+  # parameters to override the defaults from the templates.
+  # if a resource in the supply-chain specifies a parameter
+  # of the same name that resource parameter clobber what is
+  # specified here at the top level (this includes specification
+  # as `value` vs `default`)
+  #
```
rawlingsj, line 147: I wonder if this could be reworded a little differently, assuming this is true something like:

_List of parameters to override supply-chain template parameter defaults.  If a supply-chain contains a default of the same parameter name then it will be replaced.  For reference parameter values in supply-chain templates are mandatory and cannot be overridden._

rawlingsj: or maybe a copy / paste of the description in this PR as that seems clearer

rawlingsj:
Just been having a look at other places in Kubernetes that use a concept of a default value.  `LimitRange` seems like a similar approach where you can add default values for memory and CPU but these can be overridden on per podTemplate basis https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

Just a thought, I wonder if rather than a supply chain value that cannot be overridden perhaps it should be called `mandatory`? Just it feels like mandatory is a little more self explanatory if it cannot be overridden?  Just an idea though.

squeedee:
> Just a thought, I wonder if rather than a supply chain value that cannot be overridden perhaps it should be called mandatory? Just it feels like mandatory is a little more self explanatory if it cannot be overridden? Just an idea though.

This was heavily discussed on Monday in the office hours. We had `overidable` vs `default` behavior and for those present,  default made much more sense. 

**edit**: perhaps I don't know what you mean by `mandatory`? As a flag or as a name for the value field?

pkg/apis/v1alpha1/common.go
squeedee: We'll do a robust search and replace for `'%?'` throughout the code in a later PR.

pkg/templates/params.go
squeedee: deferring for public meeting discussion

jwntrs:
@rawlingsj very interesting, thanks!

> Just been having a look at other places in Kubernetes that use a concept of a default value. LimitRange seems like a similar approach where you can add default values for memory and CPU but these can be overridden on per podTemplate basis https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/


```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: mem-limit-range
spec:
  limits:
  - default:                 # <========= default value
      memory: 512Mi
    defaultRequest:
      memory: 256Mi
    type: Container
```

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: default-mem-demo-2
spec:
  containers:
  - name: default-mem-demo-2-ctr
    image: nginx
    resources:
      limits:
        memory: "1Gi"         # <======== overridden by the pod
```

site/content/docs/reference.md
jwntrs: We still need a fairly comprehensive overhaul of our docs. To me the clearest way to describe the behaviour is simply by enumerating all the possibilities:

supply chain resource entry **value** (highest precedence)
supply chain **value**
workload **value**
supply chain resource entry **default**
supply chain **default**
template **default** (lowest precedence)

waciumawanjohi: I think the most subtle distinction (the one I have difficulty describing succinctly) is the situation where the supply chain specifies a value, but the individual resource specifies a default. In this case, if the workload specifies a value, the workload's value will be respected. That's not evident from the hierarchy represented here. That's the behavior that I'm referencing when I say the resource clobbers the supply chain specification. The resource can not only change the default, it can change a supply chain's _value_ into a _default_.

jwntrs: > The resource can not only change the default, it can change a supply chain's value into a default.

@waciuma In the example you supplied, it's actually the supply overriding itself that causes this behaviour.

PR #390, opened 2021-11-23
Give better message to the use for missing value at path

### LINKED ISSUE #378: MissingValueAtPath status condition
### Description of problem

As a developer, when I see this status condition, its not immediately clear where the problem is:
> Resource 'image-builder' is waiting to read value '.status.latestImage'

If I built my own SupplyChain, I might remember adding an `image-builder` resource, but other users may not have written (and may not even have access to) their SupplyChain.

Even if I looked at the tree of objects created under my workload, there's no resource called `image-builder` that I can immediately point to as the culprit.

### Proposed solution

It would be nice if we could show something like:
> Waiting to read value '.status.latestImage' from resource `Image/some-name`

emmjohnson:


pkg/controller/workload/conditions.go
```diff
@@ -87,12 +89,20 @@ func TemplateObjectRetrievalFailureCondition(err error) metav1.Condition {
...
 	return metav1.Condition{
-		Type:    v1alpha1.WorkloadResourceSubmitted,
-		Status:  metav1.ConditionUnknown,
-		Reason:  v1alpha1.MissingValueAtPathResourcesSubmittedReason,
-		Message: fmt.Sprintf("Resource '%s' is waiting to read value '%s'", resourceName, expression),
+		Type:   v1alpha1.WorkloadResourceSubmitted,
+		Status: metav1.ConditionUnknown,
+		Reason: v1alpha1.MissingValueAtPathResourcesSubmittedReason,
+		Message: fmt.Sprintf("Waiting to read value [%s] from resource [%s/%s] of type [%s]",
+			expression, obj.GetNamespace(), obj.GetName(), fullyQualifiedType),
```
squeedee: I think it's easier for debugging with `kubektl` if this is in the form `[<fully-qualified-type>/name] in ns: <namespace>`

squeedee (approved):
LGTM

PR #413, opened 2021-12-01
Supply chain and delivery can set serviceAccountRef

### LINKED ISSUE #301: SupplyChain/Delivery should have serviceAccountName
### Description of problem

In certain situations, it may not be desirable to have every workload/deliverable supply its own serviceAccountName.

### Proposed solution

Let Operators configure the serviceAccountName on the SupplyChain/Delivery.

We should prefer a Workload/Deliverable service account over a SupplyChain/Delivery service account if both are provided.

martyspiewak:
Closes #301


waciumawanjohi:
This needs documentation.
At least in the [reference doc](https://github.com/vmware-tanzu/cartographer/blob/main/site/content/docs/development/reference.md)

(I don't think the serviceAccount in the workload has been documented there, this would be a good time to add it)

martyspiewak:
> This needs documentation. At least in the [reference doc](https://github.com/vmware-tanzu/cartographer/blob/main/site/content/docs/development/reference.md)
> 
> (I don't think the serviceAccount in the workload has been documented there, this would be a good time to add it)

Makes sense. I added documentation for both in the latest commit.

pkg/controller/deliverable/reconciler_test.go
```diff
@@ -261,19 +261,50 @@ var _ = Describe("Reconciler", func() {
 			Expect(resourceRealizer).To(Equal(builtResourceRealizer))
```
waciumawanjohi: I was wondering why this assert wasn't in the context of the service account specified in the delivery or when using the default service account. Should this be a separate "It" block?

pkg/registrar/map_functions_test.go
```diff
@@ -1248,16 +1278,205 @@ var _ = Describe("MapFunctions", func() {
...
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "some-workload",
+							Namespace: "some-namespace",
+							Labels: map[string]string{
+								"some-label": "some-label-value",
+							},
+						},
+						Spec: v1alpha1.WorkloadSpec{
+							ServiceAccountName: "some-service-account",
+						},
```
waciumawanjohi, line 1375: There are 50 lines of boilerplate in the BeforeEach, but the only thing that should vary between the last test and this one is this. It's difficult to verify that the rest of the boilerplate is the same. That's why I advocate for deduplicating tests. Following how objects are built up from parent contexts is much easier than manually comparing dozens of lines of code for changes.

waciumawanjohi (approved):

pkg/registrar/map_functions_test.go
martyspiewak: I completely agree and I began to refactor the tests to do exactly that but after discussing with others we determined that we would like to refactor the whole map_functions.go and so it didn't seem worth it to refactor the tests right now if we will (hopefully) soon begin another refactor.

pkg/controller/deliverable/reconciler_test.go
martyspiewak: This makes sense to me. Addressed in latest commit.

martyspiewak:
Squashed and rebased on main and force-pushed. 

Re-requesting a review because I switched from using `serviceAccountName` to `serviceAccountRef` per the discussion in #301

pkg/apis/v1alpha1/cluster_supply_chain_test.go
```diff
@@ -55,6 +55,14 @@ var _ = Describe("ClusterSupplyChain", func() {
 			Expect(jsonValue).To(ContainSubstring("selector"))
 			Expect(jsonValue).NotTo(ContainSubstring("omitempty"))
 		})
+
+		It("allows but does not require a service account name", func() {
```
idoru: s/name/ref/

pkg/apis/v1alpha1/common.go
```diff
@@ -84,6 +84,11 @@ type ObjectReference struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 }
 
+type NameAndNamespace struct {
```
idoru: I think it would be more intent revealing if it were named `ServiceAccountRef[erence]`

site/content/docs/development/reference.md
```diff
@@ -148,6 +154,12 @@ spec:
   selector:
     app.tanzu.vmware.com/workload-type: web
 
+  # specifies the service account to be used to create resources if one
+  # is not specified in the workload
+  #
+  # (optional)
+  serviceAccountName: service-account
```
idoru: Should be `serviceAccountRef` instead now?

idoru (commented):
See individual comments. Summary:
- prefer more explicit naming for `serviceAccountRef`'s type
- test prose (name -> ref)
- docs for supplychain should show `serviceAccountRef` not `serviceAccountName`

pkg/apis/v1alpha1/common.go
martyspiewak: I'm definitely open to renaming that. I think it might make sense to keep it generic because the only fields on there are `name` and `namespace` which are not specific to service accounts.

We already have an `ObjectReference` type which has kind, apiversion, name, namespace. My first thought was we could call this new type `SimpleObjectReference` or something to that effect.

What do you think?

idoru: The reason I brought this up was that I was drawing parallels with `ClusterTemplateReference` and `DeliveryClusterTemplateReference`. Additionally, I considered design of core k8s types, say `ClusterRoleBinding`  -- you'll see the role binding has a very specific type for the ref `RoleRef` even though this has very generic fields like `APIGroup` , `Kind` and `Name`.

martyspiewak: Interesting. I think the ClusterTemplateRef is different because we’re doing validation on those to ensure that only certain kinds are used. 

But the RoleRef seems similar enough that I’m convinced we can change to ServiceAccountRef. If we ever have a need to reuse the type we can think about renaming then

idoru (approved):

PR #422, opened 2021-12-02
396 architecture

squeedee:


site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+![Owner](../img/owner.jpg)
+
+| Owner      | Blueprint |
+| ----------- | ----------- |
+| Workload | ClusterSupplyChain |
+| Deliverable | ClusterDelivery |
+
+## Theory of operation
+
+Given an owner that matches a blueprint, cartographer reconciles the resources referenced by the blueprint.
```
martyspiewak: nitpick:
```suggestion
Given an owner that matches a blueprint, Cartographer reconciles the resources referenced by the blueprint.
```

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+The resources are only created when the inputs are satisfied, and a resource is only updated when it's inputs change.
+This results in a system where a new intrinsic result from one resource can cause other resources to change.
```
martyspiewak: question:
What is an `intrinsic result`? I'm not sure I would understand what you mean if I didn't already understand cartographer, but that could just be me. 

I'm struggling to think of a clearer way to say it, though....could we just take out the word "intrinsic"?

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+
+Although Cartographer is not a 'runner of things', a resource can be something as simple as a Job or a CI pipeline.
+However, one advantage of Cartographer's design, is that a resource can also be untriggered. Imagine a Build resource 
+that discovers new base OCI images. If it rebuilds your image, then cartographer will see this new image and update 
+other linked resources.
+
+![Generic Blueprint](../img/generic.jpg)
```
martyspiewak, line 86: These diagrams are 🔥 🔥 🔥

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+| Deliverable | ClusterDelivery |
+
+## Theory of operation
+
+Given an owner that matches a blueprint, cartographer reconciles the resources referenced by the blueprint.
+The resources are only created when the inputs are satisfied, and a resource is only updated when it's inputs change.
+This results in a system where a new intrinsic result from one resource can cause other resources to change.
+
+Although Cartographer is not a 'runner of things', a resource can be something as simple as a Job or a CI pipeline.
+However, one advantage of Cartographer's design, is that a resource can also be untriggered. Imagine a Build resource
```
martyspiewak: question:
> a resource can also be untriggered

I'm not sure I understand what you mean here. Also, I had assumed the following sentence and blueprint was going to explain what this meant, but that doesn't seem to be the case. Am i missing something?

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+
+Although Cartographer is not a 'runner of things', a resource can be something as simple as a Job or a CI pipeline.
+However, one advantage of Cartographer's design, is that a resource can also be untriggered. Imagine a Build resource 
+that discovers new base OCI images. If it rebuilds your image, then cartographer will see this new image and update 
+other linked resources.
+
+![Generic Blueprint](../img/generic.jpg)
+<!-- https://miro.com/app/board/uXjVOeb8u5o=/ -->
+
+When Cartographer reconciles an owner, each resource in the matching blueprint is reconciled:
```
martyspiewak: question:

When you say `each resource in the matching blueprint is reconciled`, does "resource" refer to the template or the kubernetes object? 

Either way, it feels incorrect to say each resource is "reconciled". If it is the kubernetes object, maybe it should be "applied" or "created". If it's the template, perhaps there is a better word we could use. "reconciled" feels like it has too much domain specific meaning that we don't intend in this case.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Theory of operation
+
+Given an owner that matches a blueprint, cartographer reconciles the resources referenced by the blueprint.
+The resources are only created when the inputs are satisfied, and a resource is only updated when it's inputs change.
+This results in a system where a new intrinsic result from one resource can cause other resources to change.
+
+Although Cartographer is not a 'runner of things', a resource can be something as simple as a Job or a CI pipeline.
+However, one advantage of Cartographer's design, is that a resource can also be untriggered. Imagine a Build resource 
+that discovers new base OCI images. If it rebuilds your image, then cartographer will see this new image and update
```
martyspiewak: nitpick:
```suggestion
that discovers new base OCI images. If it rebuilds your image, then Cartographer will see this new image and update 
```

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+### Complete Supply Chain and Delivery with GitOps
+
+![Gitops](../img/gitops.jpg)
+
+## Blueprint Details
+
+### ClusterSupplyChain
+A ClusterSupplyChain blueprint continuously integrates and builds your app.
+
+![ClusterSupplyChain](../img/supplychain.jpg)
```
martyspiewak, line 114: suggestion:
This diagram is a bit blurry and the writing is quite small so it is tough to read. I can tell there is a lot to get in there, but is there a way to make it a bit more readable?

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+
+### ClusterDelivery
+A ClusterDelivery blueprint continuously deploys and validates images to a cluster. A ClusterDelivery has the ability to lock
```
martyspiewak: suggestion:
```suggestion
A ClusterDelivery blueprint continuously deploys and validates Kubernetes configuration to a cluster. A ClusterDelivery has the ability to lock 
```

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,142 @@
...
+![ClusterDelivery](../img/delivery.jpg)
+
+### Selectors
+An owner's labels will determine which blueprint will select for it. The controller will do a "best match" on a blueprint's 
+`spec.selector` with an owner's labels.
+
+A "best match" follows the rules:
+1. If all labels are fully contained in the selector, reconcile the owner with that blueprint
+2. If more than one blueprint has all the labels that the owner has, pick the most identical to the owner
+3. If multiple blueprints match the owner labels, reconcile with the blueprint with the most label matches
```
martyspiewak: question:
I'm a bit confused by these rules. In rule `2`, shouldn't it be the inverse- if an owner has all the labels of more than one blueprint? Also, what does it mean "the most identical to the owner"- does that mean most label matches?

Also I'm not sure what the difference between `2` and `3` is.

martyspiewak (commented):
Left a bunch of comments but this is really awesome! Thanks for all the hard work that went into this!

site/content/docs/development/architecture.md
emmjohnson: In time we will move these to svg and rework them.

site/content/docs/development/architecture.md
emmjohnson: 2. If not all labels match, we choose the blueprint with the most matched labels.
3. If two blueprints match all labels, the blueprint with the more concise match (less non-matching labels) is selected

blueprint a is selected for 

workload
a
b


blueprint a
a
b

blueprint b
a
b
c

site/content/docs/development/architecture.md
emmjohnson: Unfortunately, resource is overloaded. We have a list of resources in the blueprint spec. We changed reconciled to applied.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
+# Architecture and Concepts
+
+## Overview
+
+Cartographer is an open-sourced Supply Chain Choreographer for Kubernetes. Cartographer provides a set of Kubernetes
```
waciumawanjohi, line 5: source

waciumawanjohi: open-source

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+controllers and CRDs that allow a platform operator to create an application platform by specifying Supply Chains and 
+application Delivery workflows.
```
waciumawanjohi: by specifying repeatable, reusable code to production roadmaps. (?)

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Concepts
+
+### Blueprints
+
+Blueprints are a list of templates (called resources) and how they depend upon each other. It forms the dependency graph of
```
waciumawanjohi: A blueprint is a list of templates (called resources) that defines how the templates depend upon each other.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+Blueprints consist of:
+* A **selector** to match owners
+* **Parameters** to pass to all resources 
+* **Resources**: 
+  * A **templateRef** pointing to the template for the resource
+  * **Parameters** to pass to the template
+  * **Inputs**, which specify dependencies for the template
+
+![Blueprint](../img/blueprint.jpg)
```
waciumawanjohi, line 33: Blueprint

All of the image tags go back one folder too many, and so don't display the image.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+<!-- https://miro.com/app/board/uXjVOeb8u5o=/ -->
+
+| Blueprint    | Owner | Valid Templates |
+| ----------- | ----------- | ----------- |
+| [ClusterSupplyChain](reference/#clustersupplychain) | [Workload](reference/#workload) | [ClusterSourceTemplate](reference/#clustersourcetemplate), [ClusterImageTemplate](reference/#clusterimagetemplate), [ClusterConfigTemplate](reference/#clusterconfigtemplate), [ClusterTemplate](reference/#clustertemplate) |
+| [ClusterDelivery](reference/#delivery) | [Deliverable](reference/#deliverable) | [ClusterSourceTemplate](reference/#clustersourcetemplate), [ClusterDeploymentTemplate](reference/#clusterdeploymenttemplate), [ClusterTemplate](reference/#clustertemplate) |
+
```
waciumawanjohi: I would move this to right under the heading concepts. Super useful chart that would help users understand terms that are used in the discussion of blueprint.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+### Templates
+
+Templates create or update resources (i.e. kubectl apply).
```
waciumawanjohi, line 38: Templates create or redefine resources (?)

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+Templates consist of:
+* Parameters to pass to `spec.template` or `spec.ytt`
+* The Kubernetes resource yaml as `spec.template` or `spec.ytt` [see Templating](#tbd)
+* **Output paths** which tell Cartographer where to find the output of the Kubernetes resource
+  * The path field depends upon the specific template kind.
+
+![Template](../img/template.jpg)
+
+Templates are typed by the output they produce.
```
waciumawanjohi, line 48: Templates are typed by the output they expose to the Blueprint.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+| Source | [ClusterSourceTemplate](reference/#clustersourcetemplate) |
+| Deployment | [ClusterDeploymentTemplate](reference/#clusterdeploymenttemplate) |
+| | [ClusterTemplate](reference/#clustertemplate) |
+
+### Owners
+
+Owners represent the workload or deliverable, which in many cases refer to a single application's source or image 
+location.
+
+Owners are the developer provided configuration which cause a blueprint to be reconciled into resources.
```
waciumawanjohi, line 63: Owners (workloads or deliverables) are the developer provided configuration which cause a blueprint to be reconciled into resources. In most cases these refer to a single application's source or image 
location.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+They consist of:
+* **Labels**: blueprints will select based on the labels of an owner, see [selectors](#selectors) 
+* **Params**: parameters supplied to the blueprint, see [Parameter Hierarchy](#parameter-hierarchy)
+* **Source**: The source reference for the input to the Supply Chain or Delivery Blueprints,
```
waciumawanjohi, line 70: As an aside, neither Workload nor Deliverable requires the source field. Maybe this is an indication that they should.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+| Owner      | Blueprint |
+| ----------- | ----------- |
+| Workload | ClusterSupplyChain |
+| Deliverable | ClusterDelivery |
+
+## Theory of operation
+
+Given an owner that matches a blueprint, Cartographer reconciles the resources referenced by the blueprint.
+The resources are only created when the inputs are satisfied, and a resource is only updated when it's inputs change.
```
waciumawanjohi: its

waciumawanjohi: ...and a resource definition is only updated...

(to differentiate updates from the resource's controller, which are happening all the time)

site/content/docs/development/architecture.md
waciumawanjohi: I love the diagrams in general. (And I'm impressed by how polished they look!)

With this one, I'm wary of the implication that the k8s resources that are created in the supply chain are the concern for the developer rather than the operator. I don't think that's a problem to fix today because of the time crunch.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+This results in a system where a new result from one resource can cause other resources to change.
+
+![Generic Blueprint](../img/generic.jpg)
+<!-- https://miro.com/app/board/uXjVOeb8u5o=/ -->
+
+Although Cartographer is not a 'runner of things', a resource can be something as simple as a Job or a CI pipeline.
```
waciumawanjohi: I would omit 'or a CI pipeline'. Or maybe change to a tekton pipeline.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+However, one advantage of Cartographer's design is that resources that self-mutate can cause downstream change.
+
+For example, a Build resource that discovers new base OCI images. If it rebuilds your image, then Cartographer will see
```
waciumawanjohi, line 93: Consider a Build resource that discovers new base OCI images. If it reconciles and creates a new image,

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+For example, a Build resource that discovers new base OCI images. If it rebuilds your image, then Cartographer will see 
+this new image and update downstream resources.
+
+When Cartographer reconciles an owner, each resource in the matching blueprint is applied:
+
+1. **Generate Inputs**: Using the **blueprint resource's** `inputs` as a reference, select outputs from previously applied **Kubernetes resources**
+2. **Generate Params**: Using the [Parameter Hierarchy](architecture.md#parameter-hierarchy), generate parameter values   
+3. **Generate and apply resource spec**: Apply the result of interpolating `spec.template` (or `spec.ytt`), **inputs**, **params** and the **owner spec**. 
+4. **Retrieve Output**: Store the output from the applied resource. The output to use is specified in the **template output path**.
```
waciumawanjohi, line 101: Great summary!

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+this new image and update downstream resources.
+
+When Cartographer reconciles an owner, each resource in the matching blueprint is applied:
+
+1. **Generate Inputs**: Using the **blueprint resource's** `inputs` as a reference, select outputs from previously applied **Kubernetes resources**
+2. **Generate Params**: Using the [Parameter Hierarchy](architecture.md#parameter-hierarchy), generate parameter values   
+3. **Generate and apply resource spec**: Apply the result of interpolating `spec.template` (or `spec.ytt`), **inputs**, **params** and the **owner spec**. 
+4. **Retrieve Output**: Store the output from the applied resource. The output to use is specified in the **template output path**.  
+
+![Realize](../img/realize.jpg)
```
waciumawanjohi, line 103: Dependency arrow always throw me off. That's a me thing.

I think the dependency arrow on the top output is facing the wrong way.

I'm not sure that the bottom output should have a dependency arrow.

site/content/docs/development/architecture.md
squeedee: please try and make suggested edits, then I can just click them

site/content/docs/development/architecture.md
squeedee: They work perfectly when rendered.  Check the preview https://deploy-preview-422--elated-stonebraker-105904.netlify.app/docs/development/architecture

site/content/docs/development/architecture.md
waciumawanjohi: A picture is worth a thousand words (and this is a good picture). This could still benefit from some text. Particularly to call out that the handoff is always by configuring the supply chain to put configuration in some place that the Deliverable can read it (On a single cluster, that could be a ConfigMap). And to call out that in the multi-cluster case we expect users to use GitOps with either a git repo or an image registry.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+## Blueprint Details
+
+### ClusterSupplyChain
+A ClusterSupplyChain blueprint continuously integrates and builds your app.
+
+![ClusterSupplyChain](../img/supplychain.jpg)
+
+### ClusterDelivery
+A ClusterDelivery blueprint continuously deploys and validates Kubernetes configuration to a cluster. A ClusterDelivery 
+has the ability to lock (and unlock) templates which pauses the continuous deploy.
```
waciumawanjohi: > A ClusterDelivery has the ability to lock (and unlock) templates which pauses the continuous deploy.

This behavior will not be in the next release.

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+### ClusterDelivery
+A ClusterDelivery blueprint continuously deploys and validates Kubernetes configuration to a cluster. A ClusterDelivery 
+has the ability to lock (and unlock) templates which pauses the continuous deploy. 
+
+<!--- @TODO MORE ON LOCKING -->
+
+![ClusterDelivery](../img/delivery.jpg)
+
+### Selectors
+An owner's labels will determine which blueprint will select for it. The controller will do a "best match" on a blueprint's
```
waciumawanjohi, line 126: An owner's labels will determine with which blueprint it will pair.
OR
An owner's labels will determine which blueprint it will pair with.

site/content/docs/development/architecture.md
squeedee: I don't know, might skip this for now. We really need to chose the language for "apply" that people will instantly _grok_ as meaning create/update.  I use Update here as it's part of the CRUD idiom, and literally what you see as the verb on the wire to APIServer too.

site/content/docs/development/architecture.md
waciumawanjohi: do a "best match" on blueprints'

site/content/docs/development/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+
+Note: Despite the rules, the controller can still return more than one match. If more than one match is returned, 
+no blueprint will reconcile for the owner.
+
+## Parameter Hierarchy
+
+<!--- @TODO Image of params -->
+
+
+Templates specify the **parameters** they accept in `spec.params`. These can have a default value.
```
waciumawanjohi: > Templates specify the **parameters** they accept in `spec.params`.

This is no longer the case, a template need not specify a param in `spec.params` in order to reference it in the template. Supply chains or workloads can provide that param.

site/content/docs/development/architecture.md
squeedee: repeating the same concept of `(workloads or deliverables)` feels a little spammy to me, we literally lay it out what an owner is 1 para up directly under the heading that clarifies what owners are.

I will submit something about ownership

site/content/docs/development/architecture.md
squeedee: I think you're right

waciumawanjohi (approved):

site/content/docs/development/architecture.md
squeedee: yeah because we needed a data flow arrow at the bottom, we used one at the top, there's no great convention for it other than to create a key and use a decoration.

squeedee: I really wanted a 'data bucket' for in memory 'outputs', which we could then simply 'depend' upon, however the diagram gets very confusing that way too.

site/content/docs/development/architecture.md
squeedee: I'm implying that you pick the type based on the path you would select for the stamped resource. I contend that the choice is forced by the content, not the context you hope it'll end up in?

site/content/docs/development/architecture.md
squeedee: TY!

site/content/docs/development/architecture.md
squeedee: If you want to try your hand at an alternative, this is the board: https://miro.com/app/board/uXjVOeb8u5o=/ 
Just clone the Frame and have at it. If you don't want to but feel strongly that the diagram is too misleading, can you raise a story?

PR #423, opened 2021-12-02
414 informers use best match

### LINKED ISSUE #414: Informers still use exact match to find workloads/deliverables for supply chains/deliveries
### Bug description:
In #359 we changed the behavior of supply chains and deliveries to match workloads and deliverables based on a "best match" of labels. This change was not made in the watchers, though, and so when a change happens to a supply chain or delivery, it still looks for the relevant workloads and deliverables by matching labels exactly (see [here](https://github.com/vmware-tanzu/cartographer/blob/main/pkg/registrar/map_functions.go#L116) and [here](https://github.com/vmware-tanzu/cartographer/blob/main/pkg/registrar/map_functions.go#L148), respectively).

### Steps to reproduce:
1. Apply supply chain A with label X and supply chain B with labels X and Y
2. Apply a workload that has labels X and Y
3. Make a change to supply chain A

### Expected behavior:
The workload should not re-reconcile because it matches supply chain B better

### Actual behavior:
The workload does re-reconcile

martyspiewak:
Closes #414

### DIFF
--- a/pkg/registrar/map_functions.go
+++ b/pkg/registrar/map_functions.go
@@ -19,6 +19,7 @@ package registrar
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -27,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/cartographer/pkg/repository"
 )
 
 //counterfeiter:generate sigs.k8s.io/controller-runtime/pkg/client.Client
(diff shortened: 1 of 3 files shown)


pkg/registrar/map_functions_test.go
```diff
@@ -471,6 +506,26 @@ var _ = Describe("MapFunctions", func() {
...
+							sc := &v1alpha1.ClusterSupplyChain{
+								Spec: v1alpha1.SupplyChainSpec{
+									Selector: map[string]string{
+										"myLabel":      "myLabelsValue",
+										"myOtherLabel": "myOtherLabelsValue",
+									},
+								},
+							}
+
+							clientObjects = []client.Object{workload, sc}
```
waciumawanjohi: Pretty sure this should be
```
clientObjects = []client.Object{workload, sc, clusterSupplyChain}
```

removing `sc` should make the test fail, but in the current setup it does not.

pkg/registrar/map_functions.go
```diff
@@ -127,17 +129,45 @@ func (mapper *Mapper) ClusterSupplyChainToWorkloadRequests(object client.Object)
...
+	scList := &v1alpha1.ClusterSupplyChainList{}
+	err = mapper.Client.List(context.TODO(), scList)
+	if err != nil {
+		mapper.Logger.Error(err, "cluster supply chain to workloads: client list supply chains")
+		return nil, err
+	}
+
+	var selectorGetters []repository.SelectorGetter
+	for _, item := range scList.Items {
+		item := item
```
waciumawanjohi, line 147: Not blocking:
Is this just to create a copy of the object? Would `persistentItem := item.DeepCopy()` suit? Would it be more grokkable?

waciumawanjohi (changes requested):
Just the one change to fix the test. Otherwise, lgtm!

pkg/registrar/map_functions.go
martyspiewak: I think that would work too but I was trying to be consistent with the way the controller finds the best matching SC (see [here](https://github.com/vmware-tanzu/cartographer/blob/main/pkg/repository/repository.go#L286-L290))

pkg/registrar/map_functions_test.go
martyspiewak: Good catch thanks!

waciumawanjohi (approved):

Issue #383, opened 2021-11-22
mapper should take context not logger


emmjohnson:
We also need to make logging and errors consistent in mapper

emmjohnson:
Does not appear to be a way to get context to mapfuncs. Started discussion here - https://kubernetes.slack.com/archives/C02MRBMN00Z/p1639159590098700

PR #424, opened 2021-12-03
Workloads/deliverables reconcile on changes to default service account

### LINKED ISSUE #409: reconcile on changes to default service account when a serviceAccountName is not specified
### Bug description:
When a service account, role, or role binding changes, right now the watchers are configured to only trigger a reconciliation for workloads/deliverables that explicitly reference the service account that has changed (either through a change to the service account itself or to a role/role binding affecting that service account (which is where bug #394 exists)).

### Steps to reproduce:
1. Apply a workload/deliverable that does not specify a service account
2. Modify the default service account, or a role/role binding that affects it

### Expected behavior:
The workload/deliverable should re-reconcile

### Actual behavior:
The workload/deliverable will not re-reconcile

### Additional context:
It's possible we don't want this behavior as it is too much overhead for us to account for and maybe we say if you want the watchers you have to explicitly reference the default service account in your workload/deliverable. I imagine this isn't the case, but might be something worth discussing.

martyspiewak:
Closes #409

### DIFF
--- a/pkg/registrar/map_functions.go
+++ b/pkg/registrar/map_functions.go
@@ -391,7 +391,7 @@ func (mapper *Mapper) ServiceAccountToWorkloadRequests(serviceAccountObject clie
 	for _, sc := range supplyChains {
 		scWorkloads, err := mapper.clusterSupplyChainToWorkloads(sc)
 		if err != nil {
-			mapper.Logger.Error(err, fmt.Sprintf("service account to workload requests: cluster supply chain [%s] to workloads", sc.Name))
+			mapper.Logger.Error(err, "service account to workload requests: cluster supply chain to workloads", "supply chain", sc.Name)
 			return nil
 		}
 		for _, workload := range scWorkloads {
@@ -583,7 +583,7 @@ func (mapper *Mapper) ServiceAccountToDeliverableRequests(serviceAccountObject c
 	for _, d := range deliveries {
 		deliveryDeliverables, err := mapper.clusterDeliveryToDeliverables(d)
 		if err != nil {
-			mapper.Logger.Error(err, fmt.Sprintf("service account to deliverable requests: cluster delivery [%s] to deliverables", d.Name))
+			mapper.Logger.Error(err, "service account to deliverable requests: cluster delivery to deliverables", "delivery", d.Name)
 			return nil
 		}
 		for _, deliverable := range deliveryDeliverables {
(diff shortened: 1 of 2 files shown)


martyspiewak:
~~The work here is done but leaving as a draft to give time to discuss if this is desired behavior~~

We've decided to get this work merged in and then improve the design in #444

pkg/registrar/map_functions.go
```diff
@@ -391,24 +391,22 @@ func (mapper *Mapper) ServiceAccountToWorkloadRequests(serviceAccountObject clie
 	for _, sc := range supplyChains {
 		scWorkloads, err := mapper.clusterSupplyChainToWorkloads(sc)
 		if err != nil {
-			mapper.Logger.Error(err, "service account to workload requests")
+			mapper.Logger.Error(err, fmt.Sprintf("service account to workload requests: cluster supply chain [%s] to workloads", sc.Name))
```
squeedee: Please use [keys and values, not fmt (link)](https://github.com/vmware-tanzu/cartographer/blob/main/CONTRIBUTING.md#best-practices)

pkg/registrar/map_functions.go
```diff
@@ -581,23 +583,22 @@ func (mapper *Mapper) ServiceAccountToDeliverableRequests(serviceAccountObject c
 	for _, d := range deliveries {
 		deliveryDeliverables, err := mapper.clusterDeliveryToDeliverables(d)
 		if err != nil {
-			mapper.Logger.Error(err, "service account to deliverable requests")
+			mapper.Logger.Error(err, fmt.Sprintf("service account to deliverable requests: cluster delivery [%s] to deliverables", d.Name))
```
squeedee: Keys and Values as above

squeedee (changes requested):
Just two log lines and then it's good to go. Ping me in slack when you've done it to hurry me along :D

pkg/registrar/map_functions.go
martyspiewak: I was trying to maintain consistency with how we handle errors in this file elsewhere (see [here](https://github.com/vmware-tanzu/cartographer/blob/79aaec9274dce66f8556d818f6bc191a8775fa08/pkg/registrar/map_functions.go#L76), for example).

There is an open issue (#383) that should address this problem in the entire file so I think it would be best to maintain consistency for now and fix it as part of that issue.

martyspiewak: Address in latest commit

squeedee (approved):

PR #429, opened 2021-12-06
Integration tests: Make audit logging inline with app log output

squeedee:
Make audit logging inline with app log output when running Integration tests

Now you can see the API server activity closer to the time that something happens in the controller

- also dry up the DebugControlPlane code with testEnv Creation


tests/integration/runnable/runnable_suite_test.go
```diff
@@ -56,30 +56,15 @@ var (
 	controllerBuffer *gbytes.Buffer
 )
 
-const DebugControlPlane = false
+const DebugControlPlane = true
```
waciumawanjohi: This looks like it should be removed.

tests/integration/supplychain/supply_chain_suite_test.go
```diff
@@ -65,23 +65,7 @@ var _ = BeforeSuite(func() {
...
-
-		AttachControlPlaneOutput: DebugControlPlane, // Set to true for great debug logging
-	}
-
-	if DebugControlPlane {
-		testEnv.ControlPlane.GetAPIServer().Configure().
-			Append("audit-policy-file", filepath.Join(workingDir, "policy.yaml")).
-			Append("audit-log-path", "-")
-	}
+	testEnv = integration.CreateTestEnv(workingDir, GinkgoWriter)
```
waciumawanjohi, line 66: The `const DebugControlPlane = false` 8 lines prior looks like it should be removed.

tests/integration/runnable/runnable_suite_test.go
squeedee: Oops good catch

waciumawanjohi:
The de-duplication makes sense.

I tried running the integration tests on main and on this branch, inducing a failure in each. I didn't see a difference in behavior. Maybe I wasn't creating the right type of mess? Can you suggest steps I can take to observe the change?

squeedee:
> The de-duplication makes sense.
> 
> I tried running the integration tests on main and on this branch, inducing a failure in each. I didn't see a difference in behavior. Maybe I wasn't creating the right type of mess? Can you suggest steps I can take to observe the change?

@waciumawanjohi 

On my demo branch ([demo-audit-logs](https://github.com/vmware-tanzu/cartographer/tree/demo-audit-logs)):
- I set the constant `DebugControlPlane` to True.
- I added a fail at the end of a runnable integration test
- Observe that audit logs (`"apiVersion":"audit.k8s.io/v1"`) are interpersed between our logs (`"logger":"controller.runnable-service"`):
  ```
  {"level":"info","ts":1638983136.663757,"logger":"controller.runnable-service", ...
  {"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Request","auditID":"36e6 ...
  {"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Request","auditID":"36e6 ... 
  {"level":"info","ts":1638983136.667907,"logger":"controller.runnable-service", ...
  ```

On the "main demo branch" (https://github.com/vmware-tanzu/cartographer/tree/old-audit-log-demo):
- I set the DebugControlPlane to True
- I added the same fail in the runnable integration test
- observe that the audit logs are dumped before the controller logs (and/or after) but not interspersed

squeedee:
I'll just add, that by interspersing the logs, @idoru and I were able to debug reconciler behavior accurately because CRUD operations happened in sequence with our controller logging.

tests/integration/test_helpers.go
```diff
@@ -0,0 +1,48 @@
...
+			filepath.Join(workingDir, "..", "..", "..", "config", "crd", "bases"),
+			filepath.Join(workingDir, "..", "..", "resources", "crds"),
+		},
+		AttachControlPlaneOutput: DebugControlPlane, // Set to true for great debug logging
+	}
+
+	if DebugControlPlane {
+		apiServer := testEnv.ControlPlane.GetAPIServer()
+		apiServer.Out = out
+		apiServer.Err = out
```
squeedee, line 41: This is the change that actually "makes audit logging inline"

waciumawanjohi (approved):

PR #451, opened 2021-12-08
usability: Add aggregated ClusterRoles for admin, view, and edit.

### LINKED ISSUE #450: Support Kubernetes' default aggregated roles
### Description of problem

Users with [the built-in `admin`, `edit`, and `view` roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) cannot manage Cartographer resources

### Proposed solution

Add ClusterRoles with the appropriate annotations to enable default roles to be able to view the resources.

(as cluster-admin)
```shell
$ kubectl create rolebinding dev-editor --clusterrole edit --user dev@company.com
```

(as dev@company.com)
```shell
$ kubectl get workload
NAME      AGE
sockeye   39h
```

evankanderson:

closes #450

Adds ClusterRoles for end-users which [aggregate into the Kubernetes default user-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).

## Release Note

* Users with the Kubernetes default `admin`, `edit`, and `view` roles can now access Cartographer resources.

## PR Checklist


config/rbac/rbac.yaml
```diff
@@ -60,3 +60,90 @@ rules:
...
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+  - apiGroups:
+      - carto.run
+    resources:
+      - clustersupplychains
+      - clustersourcetemplates
+      - clusterimagetemplates
+      - clustertemplates
+      - clusterdeliveries
```
scothis: What does a ClusterRole referencing cluster scoped resources do when applied via a RoleBinding in a namespace? My expectation is that the cluster scoped resources are ignored, but want to verify that. We don't want to give namespace admins access to cluster scoped resources.

martyspiewak:
@evankanderson If you could rebase on main the validation tests should start passing (fixed here: #447)

config/rbac/rbac.yaml
evankanderson:
```
$ kubectl get workloads
NAME      AGE
sockeye   2d10h
$ kubectl get clustersupplychains
Error from server (Forbidden): clustersupplychains.carto.run is forbidden: User "evana@vmware.com" cannot list resource "clustersupplychains" in API group "carto.run" at the cluster scope
```

This is with a rolebinding of `edit` in the `default` namespace.

evankanderson: But if you grant `edit` at the cluster level, you'd be able to _view_ the supplychains.

I could be convinced to drop these.

evankanderson:
Merged; I can rebase / squash if you want.

config/rbac/rbac.yaml
martyspiewak: It does seem from [the k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) and [this CNCF blog post](https://www.cncf.io/blog/2020/08/28/kubernetes-rbac-101-authorization/) that these roles are intended to be used _only_ for namespaced resources. Therefore, I think I'd be in favor of dropping the cluster scoped resources from all of these roles.

(This would make the admin and edit roles identical, so maybe we could use one ClusterRole with both labels for aggregation.)

evankanderson: Done.

martyspiewak (commented):
Sorry if I wasn't so clear in my previous comment- it seems from those articles that these roles are meant for namespace scoped resources and so I was suggesting we remove the cluster scoped resources even from the "view" role.

I also think it'd be inconsistent for them to be in the view role and not the others. But I'm interested to hear your thoughts.

scothis:
While not necessary the focus of this PR, this raises an interesting question as to how developers are expected to get access to the cluster scoped resources (ClusterSupplyChain in particular) to know what workload types are supported in the cluster, or what params are available on a particular supply chain.

martyspiewak:
@scothis I believe that is what [RFC 12](https://github.com/vmware-tanzu/cartographer/blob/main/rfc/rfc-0012-discoverable-input-fields.md) addresses.

scothis:
@martyspiewak regardless of how the contract in expressed on the resource, if the user doesn't have access to view that resource, the contract isn't going to be helpful.

martyspiewak:
@scothis Makes sense. So do we expect most developers to have this `view` role? If we do, I think I'd change my mind and say we should leave the cluster scoped resources on that role.

scothis:
> So do we expect most developers to have this view role?

Not sure, but probably.

> If we do, I think I'd change my mind and say we should leave the cluster scoped resources on that role.

The `view` ClusterRole is typically bound with a namespaced RoleBinding. Based on [Evan's findings above](https://github.com/vmware-tanzu/cartographer/pull/451#discussion_r766047259) the cluster scoped resources won't be visible. If a user was granted access to the view role for the cluster, then they would be able to see all Workloads (among many other resources) in all namespaces.

I'd suggested not depending on the aggregated roles for any cluster scoped resources and figuring out a different way to grant access to them for regular users.

martyspiewak:
Sorry for dropping this for so long over the holidays! 

So @scothis just to recap, from

> I'd suggested not depending on the aggregated roles for any cluster scoped resources and figuring out a different way to grant access to them for regular users.

it sounds like we should drop the cluster scoped resources from the `view` role as well. Am I understanding that correctly?

martyspiewak:
@evankanderson per the discussion above, could you drop the cluster scoped resources from the `view` role? I think it should be good to merge once that's done.

config/rbac/rbac.yaml
```diff
@@ -60,3 +60,53 @@ rules:
...
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+rules:
+  - apiGroups:
+      - carto.run
+    resources:
+      - clustersupplychains
+      - clustersourcetemplates
+      - clusterimagetemplates
+      - clustertemplates
+      - clusterdeliveries
```
martyspiewak:
```suggestion
```

martyspiewak (changes requested):

martyspiewak (approved):

martyspiewak:
Just committed that small change myself instead so we can merge this. Thanks for the PR @evankanderson and for helping with the review @scothis!

PR #483, opened 2021-12-14
ci: add grype vuln scanning for source code & image

cirocosta:

from the commit message:

```
    ci: add grype scanning

    two extra steps to the validation workflow:

    - source scanning: making use of the standard go lock files, grype is
      able to crossreference that against its database (which is updated on
      every run)

    - image scanning: for the image we produced while setting up e2e, grype
      is able to lookup the debian packages inside the container image as
      well as inspect the binary that it founds there (go modules can be
      retrieved from a specific section of the binary).

    it might be seen as redundant to have the source scan, but, I found it
    useful for separating other potential problems we could see from, e.g.,
    finding vulnerabilities in the `ytt` binary that we include in the
    image to support templating w/ it.
```

capturing some thoughts out of band:

> _**scothis** in general, set things like this to alert, but not block_
> _it's easy to lose build reproducibility otherwise_

this PR does essentially that: it can tell us what vulnerabilities have been
found, but it won't block anything based on that (informational only).

## PR Checklist

### DIFF
--- a/.github/workflows/validation.yaml
+++ b/.github/workflows/validation.yaml
@@ -46,7 +46,7 @@ jobs:
       - name: setup base dependencies
         run: |-
           sudo apt update && sudo apt install -y curl tar
-          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree
+          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree grype
 
       - name: setup carvel tooling binaries
         uses: vmware-tanzu/carvel-setup-action@v1
@@ -57,4 +57,16 @@ jobs:
         run: make test
 
       - name: run e2e example
-        run: ./hack/setup.sh cluster cartographer example-dependencies example
+        run: |-
+          sudo ./hack/ci/trust-local-registry.sh
+          ./hack/setup.sh cluster cartographer example-dependencies example
+
+      - name: scan source
+        run: |-
+          grype .
+
+      - name: scan image
+        env:
+          DOCKER_CONFIG: /tmp/cartographer-docker
+        run: |-
+          grype $(cat ./release/cartographer.yaml | grep ' image: ' | awk '{print $NF}')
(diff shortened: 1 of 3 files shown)


.github/workflows/validation.yaml
```diff
@@ -57,4 +57,16 @@ jobs:
         run: make test
 
       - name: run e2e example
-        run: ./hack/setup.sh cluster cartographer example-dependencies example
+        run: |-
+          sudo ./hack/ci/trust-local-registry.sh
```
cirocosta, line 61: this is so that when grype tries to do the equivalent of `docker pull`, it'll not complain about the local registry being insecure (i.e., not serving content through tls w/ a cert it trusts).

we need extra privileges for it as `/etc/docker/daemon.json` is created by root, and we need to send a SIGHUP to `dockerd` (which runs as ... root)

hack/ci/install-binaries.sh
```diff
@@ -27,6 +27,8 @@ readonly GH_VERSION=${GH_VERSION:-2.0.0}
 readonly GH_CHECKSUM=${GH_CHECKSUM:-20c2d1b1915a0ff154df453576d9e97aab709ad4b236ce8313435b8b96d31e5c}
 readonly KUBECTL_TREE_CHECKSUM=${KUBECTL_TREE_CHECKSUM:-f4a3230d46b824889fca56525d051aad70815965a94623388ec0b5dc71781790}
 readonly KUBECTL_TREE_VERSION=${KUBECTL_TREE_VERSION:-0.4.1}
+readonly GRYPE_CHECKSUM=a0aaae28792a70fd465301cef0f3dc4bd09c2e707208f7a576e4085c8ea861d4
+readonly GRYPE_VERSION=0.27.2
```
cirocosta, line 31: note that we _don't_ have to update grype on every grype database update - those are separate things (when you run it, it'll check if it needs to update of not: if so, then it does at runtime).

hack/ci/trust-local-registry.sh
```diff
@@ -0,0 +1,26 @@
...
+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly local_registry=$(ip route get 8.8.8.8 | grep src | awk '{print $7}'):5000
+readonly config_file=/etc/docker/daemon.json
+
+echo "$(jq ". + {\"insecure-registries\": [\"$local_registry\"]}" $config_file)" >$config_file
```
cirocosta, line 24: we can't just write the whole file like

```
echo '{
  "insecure-registries": [...]
}' > /etc/docker/daemon.json
```

because github does tweak the daemon config to have a non-standard cgroups root afair. using `jq` like this we can add the field w/out discarding what github set up.

emmjohnson:
So the only way to know if there is a new cve is to look at the logs? Seems unlikely we will do that. Could we get an alert or something to slack?

cirocosta:
> _So the only way to know if there is a new cve is to look at the logs?_

as is, _yes_

> _Could we get an alert or something to slack?_

wdyt of a comment in the PR? i'm thinking of something like https://github.com/marketplace/actions/create-or-update-comment - that way we don't block on it, but surface the problem

emmjohnson:
We should just be sure to review the logs before we cut new releases.

emmjohnson (approved):

PR #492, opened 2021-12-16
Bump Controller-Runtime to v0.11.0

squeedee:

- Set error levels using zap constants
- Remove 'Error logging enabled' tests because they aren't supported
now.

Note: `--log-level=error` will not print warnings. Previously it would have. However the `logr` interface
doesn't have a `.Warn` method, and `log.V()` no longer takes negative values (warn is -1), so we cannot emit warnings anyway.

## PR Checklist


go.mod
```diff
@@ -202,12 +202,13 @@ require (
...
 	honnef.co/go/tools v0.2.1 // indirect
-	k8s.io/component-base v0.22.4 // indirect
-	k8s.io/klog/v2 v2.9.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20211109043538-20434351676c // indirect
+	k8s.io/component-base v0.23.0 // indirect
+	k8s.io/klog/v2 v2.30.0 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
 	mvdan.cc/gofumpt v0.1.1 // indirect
 	mvdan.cc/interfacer v0.0.0-20180901003855-c20040233aed // indirect
 	mvdan.cc/lint v0.0.0-20170908181259-adc824a0674b // indirect
```
cirocosta, line 210: as @squeedee pointed out in slack, it's not great that we have this dependency coming in

as the author mentions in 

https://github.com/mvdan/mvdan.cc/commit/fc58ea77af1235d8bb87e9b10dcdc8715bf247dc

this has been archived a long time ago and shouldn't be in use.

apparently it's coming from our tools:

```console
 $ go mod why mvdan.cc/lint
# mvdan.cc/lint
github.com/vmware-tanzu/cartographer/tools
github.com/golangci/golangci-lint/cmd/golangci-lint
github.com/golangci/golangci-lint/pkg/commands
github.com/golangci/golangci-lint/pkg/lint/lintersdb
github.com/golangci/golangci-lint/pkg/golinters
mvdan.cc/interfacer/check
mvdan.cc/lint

```

so, might be worth revisiting https://github.com/vmware-tanzu/cartographer/pull/78 where we take the approach of providing a separate go.mod for the tooling deps

cirocosta: nevertheless, not a blocker for this PR at all 😁

cirocosta (approved):
lgtm!

pkg/logger/logger.go
```diff
@@ -32,11 +31,11 @@ func SetLogLevel(logLevel string) (zap.Opts, error) {
...
 	switch strings.ToUpper(logLevel) {
 	case "DEBUG":
-		level = zapcore.Level(-1)
+		level = zapcore.DebugLevel
 	case "INFO":
-		level = zapcore.Level(0)
+		level = zapcore.InfoLevel
 	case "ERROR":
-		level = zapcore.Level(1)
+		level = zapcore.ErrorLevel
```
emmjohnson, line 38: previously Level(1) was warn, do we want to change this?

squeedee: We could use warnlevel but it doesn't change the test or loggers behavior so I figured use the matching level. Zap is more granular than what the logr interface supports,  afaict.

emmjohnson (approved):

PR #493, opened 2021-12-16
Fix create/update bug

### LINKED ISSUE #484: bug: cartographer tries to create when it should update
### Bug description:

(more details to follow - need to investigate)

I haven't reproduced this myself yet, but, we've got two reports about it already:

```
  status:
    conditions:
    - lastTransitionTime: "2021-12-06T10:13:13Z"
      message: ""
      reason: Ready
      status: "True"
      type: SupplyChainReady
    - lastTransitionTime: "2021-12-07T16:34:24Z"
      message: 'unable to apply object [apps/sample-rabbitmq-app]: create: gitrepositories.source.toolkit.fluxcd.io
        "sample-rabbitmq-app" already exists'
      reason: TemplateRejectedByAPIServer
      status: "False"
      type: ResourcesSubmitted
    - lastTransitionTime: "2021-12-07T16:34:24Z"
      message: 'unable to apply object [apps/sample-rabbitmq-app]: create: gitrepositories.source.toolkit.fluxcd.io
        "sample-rabbitmq-app" already exists'
      reason: TemplateRejectedByAPIServer
      status: "False"
      type: Ready
```

even though that GitRepository object already existed. something that the reporter noticed was that the UUID that the child reference up was different, which might explain why we see this.

apparently, this happens when Workloads are recreated (

### Steps to reproduce:

Not really sure yet - to be worked on.

I tried forcing an object to _not_ be deleted by placing a dummy finalizer:

```yaml
  kind: ClusterTemplate
  metadata:
    name: template
  spec:
    template:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        finalizers:
          - finalizers.this.that/delete
      data:
        foo: bar


  ---
  kind: ClusterSupplyChain
  metadata:
    name: supplychain
  spec:
    selector:
      foo: bar
    resources:
      - name: template
        templateRef:
          name: template
          kind: ClusterTemplate

  ---
  kind: Workload
  metadata:
    name: workload
    labels:
      foo: bar
```

and then recreating the Workload, but wasn't able to reproduce the problem.

### Expected behavior:

`update/patch` rather than `create`. 

### Versions:

0.0.7? I'll try to figure out the exact version soon

martyspiewak:

* Workload/deliverable use Get instead of List (by labels) to find existing objects
* Runnable uses List by `carto.run/runnable-name` label only
* Developer service account now requires Get permission

closes #484

## Release Notes
Developer service account now requires Get permission

## PR Checklist


pkg/repository/repository_test.go
```diff
@@ -390,6 +396,43 @@ spec:
...
 		})
 
+		Context("ListUnstructured", func() {
+			//var (
+			//	namespace string
+			//	labels map[string]string
+			//	gvk schema.GroupVersionKind
+			//)
+			//
+			//BeforeEach(func() {
```
cirocosta: wip?

martyspiewak: 🙈  good catch thanks!

emmjohnson (approved):
Test locally. LGTM

squeedee (approved):

PR #494, opened 2021-12-16
go: separate tools go.mod from main go.mod

cirocosta:

- create a separate `go.mod` for tooling (as most of the scripting and
  tooling overall lives under `hack`, I created under `hack/tools`)
- point `go run`'s to it
    - unfortunately `go:generate` directives in code _don't_ use the root
    of the main module as current working directory, so we need to use
    the relative `../../..........` for counterfeiter

while it's true that the final binary will only include the modules that
are actually required in some form (either directly or transitively),
thus, not including the modules required by the tools we use, it might
be that some dependencies from them might end up being picked up by
scanners and reporting that cartographer itself is problematic despite
the dependency being one used by a tool.

by splitting the mods into the main one (go.mod) and a tools one
(hack/tools/go.mod), we're able to completely segregate while still
allowing for dependabot to discover necessary dep bumps (see
.github/dependabot.yaml changes).

note.: yes, this is very very similar to an early attempt from https://github.com/vmware-tanzu/cartographer/pull/78

## Release Note

None - developer-facing only.

## PR Checklist


squeedee:
The build tag should keep tools separated from vulnerability scanners

.github/dependabot.yml
```diff
@@ -20,6 +20,12 @@ updates:
     schedule:
       interval: daily
       time: "00:00"
+  - package-ecosystem: gomod
```
squeedee, line 23: I don't like this. Dependabot keeps deps as close to edge as possible, the tool deps we use generate and result in actual code. Dependabot automation saves us having to remember to do it.

squeedee (approved):

PR #504, opened 2021-12-22
Gitops Example: Supply Chain to Delivery

### LINKED ISSUE #405: Complex Example
Create a complex example which details using multiple environments (source + tekton + kpack + cluster config + delivery + knative)

waciumawanjohi:
---
name: Gitops Example: Supply Chain to Delivery
about: Updating the example

---

closes #405

## Release Note

- Adds example of a supply chain that ends by writing app config to git
- Adds example of a delivery that reads that git repo and deploys the app in cluster
- Adds verification of these examples to the end-to-end test

## PR Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,3 +21,7 @@ path to production. The suggested order is:
   configuration can then be read by a Delivery in another cluster. This allows all
   build processes to be isolated in a cluster completely separate from the
   production cluster exposed to customers and end users.
+- [Delivery: Git => App](basic-delivery/README.md): this example picks up app
+  configuration that has been written to git (it is written to work with the gitwriter
+  example above). The configuration is then applied to the cluster. This minimizes
+  processes occuring in the production cluster exposed to customers and end users.
--- a/examples/basic-delivery/README.md
+++ b/examples/basic-delivery/README.md
@@ -1,0 +1,338 @@
+# Source to App Config In Git
+
+**before you proceed**: the example in this directory illustrates the use of
+the latest components and functionality of Cartographer (including some that
+may not have been included in the latest release yet). Make sure to check out
+the version of this document in a tag that matches the latest version (for
+instance, https://github.com/vmware-tanzu/cartographer/tree/v0.0.7/examples).
+
+---
+
+The [gitwriter-sc] example illustrates how a supply chain can write the configuration
+for a new app into a git repository. This example picks up from that point, using 
+Delivery to read the configuration from git and create the app in the cluster.
+
+```
+
+  git --> knative service
+
+```
+
+As with [basic-sc], the directories here are structured in a way to reflect which Kubernetes
+objects would be set by the different personas in the system. But with Delivery, the
+app developer persona need not be involved; the app operator submits all objects.
+
+
+```
+  '── shared | basic-delivery
+      ├── 00-cluster                         preconfigured cluster-wide objects
+      │                                        to configured systems other than
+      │                                              cartographer (like, kpack)
+      │
+      │
+      └── app-operator                      cartographer-specific configuration
+          ├── supply-chain-templates.yaml            that an app-operator would
+          ├── ...                                                        submit
+          └── supply-chain.yaml
+```
+
+## Prerequisites
+
+1. Kubernetes v1.19+
+
+```bash
+kind create cluster --image kindest/node:v1.21.1
+```
+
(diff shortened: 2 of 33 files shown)


waciumawanjohi:
This PR only addresses the multi-cluster, gitops, supply-chain => delivery flow. There are comments in #405 about including an example of image scanning. If that's desired, we should write a separate story for that.

examples/README.md
```diff
@@ -1,6 +1,23 @@
 # Examples
```
martyspiewak: We should probably add the delivery and gitops examples to this README as well

examples/basic-delivery/README.md
```diff
@@ -0,0 +1,338 @@
...
+NAMESPACE  NAME
+default    Deliverable/gitops
+default    ├─GitRepository/gitops       configuration fetching
+default    └─App/gitops                 configuration application
+```
+
+The App can be inspected for objects created:
+
+```bash
+kgy app gitops | yq -r .status.inspect.stdout
```
martyspiewak: I assume this is an alias that accidentally made it in here?
```suggestion
kubectl get app gitops -o yaml | yq -r .status.inspect.stdout
```

hack/setup.sh
```diff
@@ -248,49 +354,296 @@ install_tekton() {
...
+                --data-value image_prefix="$REGISTRY/example-$test_name-")
 }
 
-teardown_example() {
-        kapp delete --yes -a example
+teardown_example_sc() {
+        export test_name="$1"
+        kapp delete --yes -a "example-$test_name"
+        kapp delete --yes -a "setup-example-$test_name"
+        until [[ -z $(kubectl get pods -l "serving.knative.dev/configuration=$test_name" -o name) ]]; do sleep 1; done
```
martyspiewak: This line has been hanging for me because that pod is not getting deleted and it is stopping the tests from progressing. Is this something wrong on my end (I don't think it is because I don't think the deletes above should remove this pod but I could definitely be wrong)?

hack/setup.sh
```diff
@@ -61,19 +78,36 @@ main() {
...
+
+                        log "all tests passed!!"
                         ;;
 
                 teardown-example)
-                        teardown_example
+                        teardown_runnable_example
+                        teardown_example_sc "basic-sc"
+                        teardown_example_sc "testing-sc"
+                        teardown_gitops_example
```
martyspiewak, line 105: `teardown_gitops_example` is already called in `test_gitops` so I think it'd end up being called twice if we leave this here. Maybe it should only be here and we shouldn't teardown in `test_gitops`

hack/setup.sh
```diff
@@ -46,6 +62,7 @@ main() {
                 case $command in
                 cluster)
                         start_registry
+                        start_repository
```
martyspiewak: I'm not sure I'd want a repository and gitea to be started everything I use this script to start a local cluster with cartographer.

I'd be in favor of making the gitea and gitops setup a separate step (and making the gitops testing separate as well) as they are a bit heavier to test and require more setup that we may not want to go through every time we are running tests locally. 

If not, at the very least I think we should defer this until `example-dependencies` so that the repository isn't setup unless you are setting up the examples. But as I said, I think I'd prefer to have the gitops setup and testing as a separate command. 

What do you think?

martyspiewak (changes requested):

hack/setup.sh
waciumawanjohi: This is true for the other tests as well. If they succeed, they clean up after themselves. If everything has succeeded, teardown-example needn't be called.

So why have any of the tests clean up? Shortcut to avoid test pollution.

hack/setup.sh
waciumawanjohi: This path is exercised in the e2e test. I'm not sure why it isn't working on your machine...

hack/setup.sh
waciumawanjohi: I moved `start_repository` to `example_dependencies`. I've kept all the tests together. My expectation is that the examples are run by the end to end tests, and that devs aren't running that e2e on their local machines.

hack/setup.sh
waciumawanjohi: This succeeds for me on both my personal machine and on a linux VM. But it fails in the github runner. I've commented it out for now.

martyspiewak (approved):
lgtm!

PR #519, opened 2022-01-12
Introduce RFC 18 Workload Report Artifact Provenance

waciumawanjohi:
---
name: RFC 18
about: Workload Report Artifact Provenance
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-0018-workload-report-artifact-provenance/rfc/rfc-0018-workload-report-artifact-tree.md)

---

Opens RFC 18

## Release Note

## PR Checklist

~- [ ] Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`~
~- [ ] Removed non-atomic or `wip` commits~
~- [ ] Filled in the [Release Note](#Release-Note) section above~
~- [ ] Modified the docs to match changes~

### DIFF
--- a/rfc/rfc-0018-workload-report-artifact-tree.md
+++ b/rfc/rfc-0018-workload-report-artifact-tree.md
@@ -17,14 +17,14 @@ status:
   artifacts:
     # oneOf(source,image,config)
     - source:
-        # the sha256 of the ordered JSON of exposed values
+        # the sha256 of the ordered JSON of all other non-from fields
         id: <SHA:string>
         # exposed fields - in this case url and revision
         uri: <:string>
         revision: <:string>
         # an ordered list of resources from which this artifact has been exposed
         passed:
-          - # name of the resource in the supply chain
+            # name of the resource in the supply chain
             resource-name: <:string>
             # GVK of the resource
             kind: <:string>
--- a/rfc/rfc-0018-workload-report-artifact-tree.md
+++ b/rfc/rfc-0018-workload-report-artifact-tree.md
@@ -7,7 +7,27 @@ Those artifacts should report the values exposed, as well as the resources from 
 
 ## Motivation
 
-See RFC 14
+From [RFC 14](https://github.com/paulcwarren/cartographer/blob/rfc-0014-change-tracking/rfc/rfc-0014-change-tracking.md):
+
+### Context
+
+At the core of the architecture of cartographer is the concept that a supply chain is the choreography of objects and their controllers.
+
+A supply chain, as the name suggests, chains together a set of objects defining how the (status) fields of one object feeds into the spec (and sometimes data) fields of another. Thus creating an ordered chain of interacting objects. Because controllers continuously reconcile their objects towards a desired state. A supply chain is, therefore, able to choreograph an otherwise set of independent objects (and their controllers).
+
+A workload acts as input into this supply chain seeding initial values into an object to make the first controller do work. The supply chain manages this workload input data and outputs from each object as it propagates through the supply chain.
+
+It is important to note however, that a workload can have multiple inputs and input into several objects in the supply chain all at the same time. As a result when a workload changes, it may cause several controllers to do work simultaneously.
+
+It is also important to note that a controller may also do work outside of the work that the supply chain is choreographing. Kpack images, for example, are often choreographed as part of a supply chain. But the kpack controller may also do work in response to a base OS image update. Nothing to do with the supply chain but impacting it none-the-less. As a result the supply chain may be triggered part way through by this input.
+
+### Problem
+
+Whilst, on the one hand, the behavior described above is virtuous and generally beneficial to automated outer loop workflows. On the other hand, this behavior can be problematic for more user-centric, imperative workflows, such as those often found in the inner loop. Those of debugging and live update.
+
+### Providing a real world example.
+
+Given a supply chain that has several inputs, let’s say source code image and a debug flag. When a developer applies a workload after changing their source code and turning on debug. Several services in the supply chain may trigger at the same time and these inputs will traverse through the supply chain at different times. But the developer, initiating the debug session will want to wait for that specific source code change to arrive “in cluster” in an image prepped for debugging before attaching their debugger. So, it is important to know when both inputs have fully traversed the supply chain.
 
 ## Detailed Explanation
 
@@ -133,7 +153,7 @@ status:
 
 ## Rationale and Alternatives
 
-See RFC 14 discussion
+See [RFC 14 discussion](https://github.com/vmware-tanzu/cartographer/pull/274)
 
 ## Implementation
 
--- a/rfc/rfc-0018-workload-report-artifact-tree.md
+++ b/rfc/rfc-0018-workload-report-artifact-tree.md
@@ -3,7 +3,7 @@
 ## Summary
 
 Workload/Deliverable status should report the artifacts currently exposed by the resources of the Supply Chain/Delivery.
-Those artifacts should the values exposed, as well as the resources from which that value has been read.
+Those artifacts should report the values exposed, as well as the resources from which that value has been read.
 
 ## Motivation
 
--- a/rfc/rfc-0018-workload-report-artifact-tree.md
+++ b/rfc/rfc-0018-workload-report-artifact-tree.md
@@ -53,35 +53,35 @@ Each workload/Deliverable will include in their status an artifacts field:
 ```yaml
 status:
   artifacts:
-    # oneOf(source,image,config)
+    # oneOf(source,image,config, object)
     - source:
         # the sha256 of the ordered JSON of all other non-from fields
         id: <SHA:string>
         # exposed fields - in this case url and revision
         uri: <:string>
         revision: <:string>
-        # an ordered list of resources from which this artifact has been exposed
+        # the object which produced this artifact
         passed:
-            # name of the resource in the supply chain
-            resource-name: <:string>
-            # GVK of the resource
-            kind: <:string>
-            apiVersion: <:string>
-            # name of the resource on the cluster
-            name: <:string>
-            # namespace of the resource
-            namespace: <:string>
-            # resource version of the object
-            resourceVersion: <:string>
+          # name of the resource in the supply chain
+          resource-name: <:string>
+          # GVK of the resource
+          kind: <:string>
+          apiVersion: <:string>
+          # name of the resource on the cluster
+          name: <:string>
+          # namespace of the resource
+          namespace: <:string>
+          # resource version of the object
+          resourceVersion: <:string>
         from:
-          - # id of the previous resource that was transformed into this one
+          - # id of any artifact(s) that were inputs to the template
             id: <SHA:string>
 ```
 
 example:
-Assume a supply chain that stamps out a GitRepository, Runnable, Image and config in a configmap.
-Assume that a new commit was just made, so that the GitRepository has just changed url/revision
-but the Runnable has not yet updated.
+We'll consider a supply chain that stamps out a GitRepository, Runnable (to test the source code), Image, config in a
+configmap, and finally a Kapp App. We'll assume that a new commit was just made, so that the GitRepository has
+just changed url/revision but the Runnable has not yet updated.
 
 Previous Source output:
 uri: https://www.some-site.com/my-project/my-repo
@@ -94,48 +94,55 @@ revision: b31d09004503e52e84ff633e547f4d5b40503ab3
 ```yaml
 status:
   artifacts:
-    - source:
-        id: e2212e77caf0cb64a25dfa1aca39599b69d72015dbb4ed2ad740f0666af35968
+    - source:   # <--- new revision output by the GitRepository object
+        id: 146c7d74eb956191487236b579e8e4e68462fc7d97c4f1a4677b0ded39e2a3ca
         uri: https://www.some-site.com/my-project/my-repo
         revision: b31d09004503e52e84ff633e547f4d5b40503ab3
         passed:
-          - resource-name: source-provider
-            kind: GitRepository
-            apiVersion: source.toolkit.fluxcd.io/v1beta1
-            name: my-app
-            namespace: my-namespace
-            resourceVersion: "11125094"
-    - source:
-        id: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+          resource-name: source-provider
+          kind: GitRepository
+          apiVersion: source.toolkit.fluxcd.io/v1beta1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11125094"
+    - source:   # <--- old revision from GitRepository. This source is still reported as another artifact still relies on it
+        id: 23156c7ac2170fe95f85a1ad42522c408e67038f8393eea6cd8551e07457c5d7 # <--- the sha256sum of the passed, revision and uri fields
         uri: https://www.some-site.com/my-project/my-repo
         revision: b974272e27c47a01e7a7da07cf8e4415bdb83dae
         passed:
-          - resource-name: source-provider
-            kind: GitRepository
-            apiVersion: source.toolkit.fluxcd.io/v1beta1
-            name: my-app
-            namespace: my-namespace
-            resourceVersion: "11125090"
-          - resource-name: source-tester
-            kind: Runnable
-            apiVersion: carto.run/v1alpha1
-            name: my-app
-            namespace: my-namespace
-            resourceVersion: "11125545"
+          resource-name: source-provider
+          kind: GitRepository
+          apiVersion: source.toolkit.fluxcd.io/v1beta1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11125090"
+    - source:   # <--- source artifact from the Runnable
+        id: 74cb6607d64da1e0324196517340ac7a668042f0e9dfbdd58f8a00a5d0ee9580
+        uri: https://www.some-site.com/my-project/my-repo
+        revision: b974272e27c47a01e7a7da07cf8e4415bdb83dae
+        passed:
+          resource-name: source-tester
+          kind: Runnable
+          apiVersion: carto.run/v1alpha1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11125545"
+        from:
+          - id: 23156c7ac2170fe95f85a1ad42522c408e67038f8393eea6cd8551e07457c5d7 # <--- The id noted above, as the source-tester consumes the source artifact from source-provider
     - image:
-        id: 373c0dc7d3cccd8ef31cd3dcf0f07b6bcc3a9ad1270db8fe43f84e26595af32a
+        id: bde9bc48c3d7a9dd20e94e138422264efca9770c9beb32ac18227eb204315a4a
         image: 10.138.0.2:5000/example-testing-sc-testing-sc@sha256:9aca70a5408b7d5615724bcb8e5eea3bf0765f95eac177433993cf6002311d9b
         passed:
-          - resource-name: image-builder
-            kind: Image
-            apiVersion: kpack.io/v1alpha2
-            name: my-app
-            namespace: my-namespace
-            resourceVersion: "11126348"
+          resource-name: image-builder
+          kind: Image
+          apiVersion: kpack.io/v1alpha2
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11126348"
         from:
-          - id: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+          - id: 74cb6607d64da1e0324196517340ac7a668042f0e9dfbdd58f8a00a5d0ee9580
     - config:
-        id: 22fe69c924b55327ce3e8ca079d7295a947a3641b5ab74bdf7541dc680258c81
+        id: 6e0c84490ecb71a4c5a970383c05a520f4c94546033f48a81c06472fe61f1aad
         config: |
             apiVersion: serving.knative.dev/v1
             kind: Service
@@ -159,14 +166,25 @@ status:
                   imagePullSecrets:
                     - name: registry-credentials
         passed:
-            resource-name: config-provider
-            kind: ConfigMap
-            apiVersion: v1
-            name: my-app
-            namespace: my-namespace
-            resourceVersion: "11181186"
+          resource-name: config-provider
+          kind: ConfigMap
+          apiVersion: v1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11181186"
         from:
-          - id: 373c0dc7d3cccd8ef31cd3dcf0f07b6bcc3a9ad1270db8fe43f84e26595af32a
+          - id: bde9bc48c3d7a9dd20e94e138422264efca9770c9beb32ac18227eb204315a4a
+    - object:
+        id: a03aed19284140c8093fe65a43cb1df5d16ecc12874d76aee63e4da4d7855436
+        passed:
+          resource-name: app-deploy
+          kind: App
+          apiVersion: kappctrl.k14s.io/v1alpha1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "21212378"
+        from:
+          - id: 6e0c84490ecb71a4c5a970383c05a520f4c94546033f48a81c06472fe61f1aad
 ```
 
 ## Rationale and Alternatives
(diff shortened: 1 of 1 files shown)


rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -0,0 +1,142 @@
...
+            kind: Image
+            apiVersion: kpack.io/v1alpha2
+            name: my-app
+            namespace: my-namespace
+            resourceVersion: "11126348"
+        from:
+          - id: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+    - config:
+        id: 22fe69c924b55327ce3e8ca079d7295a947a3641b5ab74bdf7541dc680258c81
+        config: |
```
jwntrs, line 146: This could get pretty big. RFC 14 proposed that this should be `shasum(giant-yaml)`

jwntrs:
Copying over [my comments from RFC14](https://github.com/vmware-tanzu/cartographer/pull/274#issuecomment-1010335810) that are still relevant:

In the office hours this week (Jan 10th, 2022), we talked about how we could extend this type of artifact tracing to track artifacts across clusters.

Assuming you started with a SupplyChain that looked something like the following (from [git-writer example](https://github.com/vmware-tanzu/cartographer/tree/main/examples/gitwriter-sc)): 

[source-provider] -> [image-builder] -> [config] -> [git-writer]

But you made a small change and wrapped the git-writer in the SupplyChain with a `ClusterSourceTemplate`, it would let you emit the revision/url of the committed config. By doing this you would end up with an artifact representing the commit to the config-repo as the last item in the status of your workload.

So your workload status might end up looking like this:
```yaml
artifacts:
  - source:
      id: 000
      url: https://github.com/my-org/my-source
      revision: abc123
      passed:
      - resource-name: source-provider
        kind: GitRepository
  - image:
      id: 111
      image: gcri.io/asdfsdfasdf
      passed:
      - resource-name: image-builder
        apiVersion: kpack.io/v1alpha2
        kind: Image
      from:
      - id: 000
  - config:
      id: 222
      config: shasum(of-some-config)
      passed:
      - resource-name: my-config
        apiVersion: v1
        kind: ConfigMap
      from:
      - id: 111
  - source:
      id: 333
      url: https://github.com/my-org/my-config
      revision: xyz789
      passed:
      - resource-name: config-writer
        kind: Runnable
      from:
      - id: 222
```

Now lets assume you are following the [basic-delivery example](https://github.com/vmware-tanzu/cartographer/tree/main/examples/basic-delivery) on another cluster.

You would end up with a Deliverable with the following status:
```yaml
artifacts:
  - source:
      id: 333
      url: https://github.com/my-org/my-config
      revision: xyz789
      passed:
      - resource-name: config-provider
        kind: GitRepository
 - source: some other-artifact
   ...
```

You can see how we could correlate these two completely independent resources, based solely on their artifacts.

jwntrs:
Copying over [my comments from RFC14](https://github.com/vmware-tanzu/cartographer/pull/274#issuecomment-1011380948) that are still relevant:

@rawlingsj made a really insightful comment in another unrelated thread that we should consider here:
> Customers will likely not be willing to give access from a dev cluster (and developer users) to others such as staging / production.  Instead it will probably be easier to send events out of those locked down clusters to a centralized place for users to get visibility into their environments.

My comment [here](https://github.com/vmware-tanzu/cartographer/pull/519#issuecomment-1011392932) definitely assumes read access to all clusters. So maybe we should start thinking about writing out artifacts as k8s events at the same time as writing them to the workload/deliverable status. That way if someone is already emitting k8s events from their prod clusters, we can integrate into their existing tooling.

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -0,0 +1,142 @@
...
+            resource-name: <:string>
+            # GVK of the resource
+            kind: <:string>
+            apiVersion: <:string>
+            # name of the resource on the cluster
+            name: <:string>
+            # namespace of the resource
+            namespace: <:string>
+            # resource version of the object
+            resourceVersion: <:string>
```
scothis: Resource version is highly volatile. Any change (or even no change for an "empty" update/patch) to the resource will create a new version. This value is typically only used for direct comparisons (are these two resource the same).

Unless you are very careful, it's easy to get into an infinite loop. As updating the Workload status will cause the Workload to be reprocessed, which will cause the templated resource to be updated, even if that update is a noop the resrouceVersion will bump, which need to be reflected on the Workload's status...

rfc/rfc-0018-workload-report-artifact-tree.md
scothis, line 146: should this example be a map[string]string instead of a string?

scothis (commented):
I'm in favor of capturing a reference to the resources created for a Workload on the status, but I'm skeptical that capturing the output values will give users the information they think they are getting. There's little point in merely duplicating state that could instead be looked up on the child resources directly unless the aggregation provides some value. It's not clear what questions you're tying to answer by providing this metadata.

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -0,0 +1,142 @@
+# Draft RFC 0018 Workload Report Artifact Provenance
+
+## Summary
+
+Workload/Deliverable status should report the artifacts currently exposed by the resources of the Supply Chain/Delivery.
+Those artifacts should the values exposed, as well as the resources from which that value has been read.
```
squeedee: English/Grammar?

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Motivation
+
+See RFC 14
```
squeedee: Can we copy it forward?

rfc/rfc-0018-workload-report-artifact-tree.md
squeedee: What's the underlying User Story/Axiom for wanting the RV?

squeedee:
> But you made a small change and wrapped the git-writer in the SupplyChain with a `ClusterSourceTemplate`

@jwntrs I think perhaps this thinking is solid, and should be enforced, that all steps, no matter what, produce an artifact. Even in the case of a kpack deploy, we should have a kpack deploy ID or something of that nature, as an artifact. That way, every Supply Chain Resource has representation in the artifacts list.

jwntrs:
@scothis 

> I'm in favor of capturing a reference to the resources created for a Workload on the status, but I'm skeptical that capturing the output values will give users the information they think they are getting. There's little point in merely duplicating state that could instead be looked up on the child resources directly unless the aggregation provides some value. It's not clear what questions you're tying to answer by providing this metadata.

The problem that we are trying to address is make it easy for the user to be able to identify the artifact graph that gets produced by a supply chain (`commit-123` produces `image-abc`). If I was building a tool on top of Cartographer to show this information, and all I had was the k8s object graph, I would need to reimplement a lot of logic that already exists in Cartographer. For instance, you would need to cross reference each template to know what fields to show from each resource. Cartographer already has this information, so we want to write it out somewhere so it can be consumed more easily.

scothis:
> The problem that we are trying to address is make it easy for the user to be able to identify the artifact graph that gets produced by a supply chain (`commit-123` produces `image-abc`)

You'll never actually know that "`commit-123` produces `image-abc`" unless you also have a deep understanding of how the resource being templated behaves and also have knowledge of the how the template is implemented. At best you can know that a new commit (`commit-123`) triggered a process that emitted an image (`image-abc`).

jwntrs:
Some open questions that I've heard raised:
- Should workload artifacts be wiped out when spec changes? If yes, can we fast forward through the supply chain matching known inputs to outputs and rebuild the artifact tree if nothing has changed?
- Should we write pending artifacts (e.g. waiting on observedGeneration: 5) and move on? If we don't we might starve the supply chain if we're blocking until a specific observedGeneration appears on the status.
- Should we write an empty artifact (only has an id) when using ClusterTemplate

scothis:
> Should workload artifacts be wiped out when spec changes? If yes, can we fast forward through the supply chain matching known inputs to outputs and rebuild the artifact tree if nothing has changed?

No. A resource should reflect the current known state of the world on its status to the best of its ability. There will inherently be latency that's ok and expected. The status should indicate the recency of the status' content. (Generation/ObservedGeneration, LastTransitionTime, etc)

> Should we write pending artifacts (e.g. waiting on observedGeneration: 5) and move on? If we don't we might starve the supply chain if we're blocking until a specific observedGeneration appears on the status.

A controller should almost never wait for an async condition to become true. The general pattern is to reflect the current state and watch "interesting" resources for changes. When one of the watched resources changes, reprocess the whole control loop and pickup the latest "current" values.

rfc/rfc-0018-workload-report-artifact-tree.md
waciumawanjohi: > As updating the Workload status will cause the Workload to be reprocessed, which will cause the templated resource to be updated, even if that update is a noop the resrouceVersion will bump

When the workload reconciles, Cartographer first checks for changes to a stamped object. If there is no proposed change to the spec, Cartographer will not update the object. This should protect us from the infinite loop scenario posited.

waciumawanjohi: > What's the underlying User Story/Axiom for wanting the RV?

Given an artifact, I would like to know the object and spec that created it. The other information can tell us what GVK/Name/Namespace the object was. But that object could have changed in arbitrary ways. The resourceVersion will point to a unique spec for the given object.

Leveraging this knowledge in practice will require additional historical persistence of objects, as etcd is only gauranteed to keep the resourceVersions for 5 minutes.

rfc/rfc-0018-workload-report-artifact-tree.md
waciumawanjohi: The config artifact has one field: `config` which is a string. Similar to the image artifact.

waciumawanjohi:
> I think perhaps this thinking is solid, and should be enforced, that all steps, no matter what, produce an artifact.

I would support such an RFC. @squeedee Do you think this should be a separate RFC, or folded into this one.

---

I'm going to fold it into this one!

waciumawanjohi:
> > The problem that we are trying to address is make it easy for the user to be able to identify the artifact graph that gets produced by a supply chain (commit-123 produces image-abc)

> You'll never actually know that "commit-123 produces image-abc" unless you also have a deep understanding of how the resource being templated behaves and also have knowledge of the how the template is implemented. At best you can know that a new commit (commit-123) triggered a process that emitted an image (image-abc).

@scothis @jwntrs This is an excellent discussion to move to RFC 20, our attempt to ensure that we can tie output N to input M.

jwntrs:
@waciumawanjohi any thoughts on this point?
> Should we write an empty artifact (only has an id) when using ClusterTemplate

waciumawanjohi:
> > Should workload artifacts be wiped out when spec changes? If yes, can we fast forward through the supply chain matching known inputs to outputs and rebuild the artifact tree if nothing has changed?

> No. A resource should reflect the current known state of the world on its status to the best of its ability. There will inherently be latency that's ok and expected. The status should indicate the recency of the status' content. (Generation/ObservedGeneration, LastTransitionTime, etc)

Cartographer's current behavior is to continue promoting artifacts that were stamped with a previous workload spec. As long as that is the behavior, we should not wipe out artifacts from the workload status simply because the workload spec changes.

What would be the implication of no longer promoting artifacts built from previous workloads? Let's consider an example: a workload paired with a simple source (flux gitrepo) -> image (kpack) -> deploy (knative) supply chain. The workload has been doing work and there are valid artifacts at each stage. The user then updates the `.spec.build.env` field on the workload. What are the implications if we don't want old artifacts to be promoted?
1. The source step is unaffected. Cartographer creates a proposed stamp of the object, sees that the spec is the same, and there's a no-op. Cartographer notes the output of the clusterSourceTemplate and moves to the next resource.
2. For the imageTemplate, there is a change. Cartographer creates the proposed stamp, sees that the spec is different from what is on the cluster, and submits the new object definition. But this is a kpack image. It continues to expose `latestImage`. But "we don't want old artifacts to be promoted". Cartographer is unaware that this is an "old" artifact, or that it is connected to a previous workload spec! It would pick up this output and promote it to Knative.

To handle this scenario we would need to do one of two things.
1. When the workload spec updates, before we update objects, delete them.
2. Wait for objects to be in a successful state before promoting them. This begins to touch on [RFC 20](https://github.com/vmware-tanzu/cartographer/pull/556).

What would we gain by adopting one of these strategies? Well one thing that isn't gained is determinism about what is deployed to prod. The old workload spec would have delivered an arbitrary number of revisions through the supply chain. What we gain is clarity about which of those revisions were driven by the old workload spec and which were driven by the new workload spec.

There is a simpler way to achieve that goal: consider the workload version as a field to be captured in the `from` field of each artifact. As we have control over the fact that workload respects generation and observedGeneration (and we will presumably control the persistence layer when we take on historical record keeping), we should leverage those fields. This will protect us from the changes to the workload's resourceVersion that will occur every time we reconcile the workload.

@jwntrs @scothis

waciumawanjohi:
> > Should we write an empty artifact (only has an id) when using ClusterTemplate

Yes. As long as there is a ClusterTemplate we can capture information about what objects were created on the cluster and what inputs led to them. I'll add that to the RFC.

waciumawanjohi:
> > > Should we write an empty artifact (only has an id) when using ClusterTemplate
> 
> Yes. As long as there is a ClusterTemplate we can capture information about what objects were created on the cluster and what inputs led to them. I'll add that to the RFC.

The rfc specifies that artifacts are `oneOf(source,image,config)` and it's not clear what to call the things created by ClusterTemplate. I'm going with `object`, and I'm very open to better suggestions.

rfc/rfc-0018-workload-report-artifact-tree.md
waciumawanjohi: I wouldn't object to encoding the yaml in order to shrink it down. But I wouldn't hash the value. This RFC should make the flow of information through the supply chain observable. Hashing the value makes it possible to see that some change has occurred, but not what that change is.

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -1,3 +1,202 @@
...
+  artifacts:
+    # oneOf(source,image,config, object)
+    - source:
+        # the sha256 of the ordered JSON of all other non-from fields
+        id: <SHA:string>
+        # exposed fields - in this case url and revision
+        uri: <:string>
+        revision: <:string>
+        # the object which produced this artifact
+        passed:
```
jwntrs: Calling this `passed` seems a bit weird after moving away from @sclevine original intent for this field. I don't really have any better suggestions though. It's mostly just a `ref` to the object, but does have the `resource-name` in there.

waciumawanjohi: We could call it `resource`.

waciumawanjohi:
> To handle this scenario we would need to do one of two things...
> Wait for objects to be in a successful state before promoting them. This begins to touch on RFC 20.

RFC 20 now proposes exactly this. Assuming RFC 20 is accepted, we could then either wipe the artifact field for a workload when the spec is updated or note the generation of the workload in the `from` field of each artifact. From an information point of view they would be equivalent. There may be implications for persistence, but without a clearer picture of what we may do there I don't have a preference between these strategies.

scothis:
Borrowing from the intent behind this RFC and 20, I'd like to propose an alternate status shape for Workloads. Not all of the fields are expected to be populatable immediately, but it gives a direction to grow towards while leaving room for additions over time. Other changes to the cartographer model will be required to be able to surface some of these fields.

Key differences:
- focus on resources rather than artifacts. Resources are the nodes of the graph, the artifacts are an output of a node along an edge to another node.
- normalize object type, they are so similar the previous oneOf adds more complexity than it adds clarity. Remember that users are not writing these types, they are reading them.
- keeps extraneous fields out of object references, separates references to template and stamped resources
- resources are linked by resource names rather than by values that happen to be the same

```yaml
...
status:
  observedGeneration: # int64
  conditions: # []metav1.Condition
  supplyChainRef:
    apiVersion: # string
    kind: # string
    namespace: # string (empty for cluster scope)
    name: # string
  resources:
  - name: # string (resource name as defined by the supply chain)
    ref: # corev1.ObjectReference (stamped resource)
      apiVersion: # string
      kind: # string
      namespace: # string (empty for cluster scope)
      name: # string
    templateRef: # corev1.ObjectReference (template resource)
      apiVersion: # string
      kind: # string
      namespace: # string (empty for cluster scope)
      name: # string
    inputs: # (I'm not sure how practical this is today given Cartographer's loose relationship between templates)
    - name: # string (resource name)
      value: # optional json.RawExtension
    outputs:
    - name: # string
      value: # json.RawExtension (may be too large to reasonably include)
      lastTransitionTime: # metav1.Time

    # these fields are modeled after metav1.Condition
    type: # enum 'Source', 'Image', or 'Config' (consider dropping if redundant with templateRef.kind)
    status: # enum 'True', 'False', or 'Unknown' (based on metav1.ConditionStatus)
    observedGeneration: # int64 (alternate to resourceVersion)
    lastTransitionTime: # metav1.Time
    reason: # string
    message: # string
```

cirocosta:
interesting @scothis, trying to make some sense of it making a little more
concrete with an example supplychain (`source-provider -> image-builder`):


```yaml
status:
  resources:
    - name: source-provider
      ref:
        kind: GitRepository ...
      templateRef:
        kind: ClusterSourceTemplate ...
      inputs: []
      outputs:
        - name: ?
          value:  {url: git.io/foo/bar, ref: b4db33f}
          lastTransitionTime: 1
        - name: ?
          value:  {url: git.io/foo/bar, ref: c001b33f}
          lastTransitionTime: 2
      type: Source
      status: True
      lastTransitionTime: 2

    - name: image-builder
      ref:
        kind: Image
      templateRef:
        kind: ClusterImageTemplate
      inputs:
        - name: source-provider
          value: {url: git.io/foo/bar, ref: b4db33f
        - name: source-provider
          value: {url: git.io/foo/bar, ref: c001b33f
      outputs:
        - name: ?                                             # (!!)
          value:  {image: registry.io/foo@sha256:010101}
          lastTransitionTime: 3
      type: Image
      status: True
      lastTransitionTime: 3
```

in the example above, just looking at `status.resources.image-builder.outputs`
one wouldn't be able to tell if that output came from the first input
(`b4db33f` or the second `c001b33f` as the outputs are not linked to the
inputs) - am I thinking about this right?

thx!

scothis:
@cirocosta inputs/outputs were not intended to be a historical record of how a value changed over time, but to capture the "current" output/input for a resource. The definition of current will likely evolve based on other proposals.

So what you sketched as:
```yaml
      outputs:
        - name: ?
          value:  {url: git.io/foo/bar, ref: b4db33f}
          lastTransitionTime: 1
        - name: ?
          value:  {url: git.io/foo/bar, ref: c001b33f}
          lastTransitionTime: 2
```

would actually be
```yaml
      outputs:
        - name: url
          value:  git.io/foo/bar
          lastTransitionTime: 1
        - name: ref
          value:  c001b33f
          lastTransitionTime: 2
```

Note: the transition times are different because the url didn't change, but the ref did in your example.

waciumawanjohi:
@scothis 
What you've proposed is achievable. But I wonder what is the motivation for status that is proposed here. The original user motivation (written originally in RFC 14) stated: "it is important to know when both inputs have fully traversed the supply chain." This proposal doesn't provide information necessary to answer that question from the workload status.

rfc/rfc-0018-workload-report-artifact-tree.md
waciumawanjohi: Changed!

squeedee:
> Yes. As long as there is a ClusterTemplate we can capture information about what objects were created on the cluster and what inputs led to them. I'll add that to the RFC.

Why do we want this to be typed? Aren't these two stipulations enough?
1. A ClusterTemplate's `outputRef` can refer to any node of any type (string, array, deeply nested object)
2. A ClusterTemplate's `output` is never a valid input. For valid inputs, you must use one of the `typed` Template kinds.

waciumawanjohi:
> Why do we want this to be typed?

In the workshopped format, the type is the top level field. Shall we change that to:

```yaml
status:
  artifacts:
    - # the sha256 of the ordered JSON of all other non-from fields
      id: <SHA:string>
      # optional: oneOf(source,image,config)
      source:
        # exposed fields - in this case url and revision
        uri: <:string>
        revision: <:string>
      # the object which produced this artifact
      resource:
        # name of the resource in the supply chain
        resource-name: <:string>
        # GVK of the resource
        kind: <:string>
        apiVersion: <:string>
        # name of the resource on the cluster
        name: <:string>
        # namespace of the resource
        namespace: <:string>
        # resource version of the object
        resourceVersion: <:string>
      from:
        - # id of any artifact(s) that were inputs to the template
          id: <SHA:string>
```

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -1,3 +1,202 @@
...
+          resource-name: <:string>
+          # GVK of the resource
+          kind: <:string>
+          apiVersion: <:string>
+          # name of the resource on the cluster
+          name: <:string>
+          # namespace of the resource
+          namespace: <:string>
+          # resource version of the object
+          resourceVersion: <:string>
```
evankanderson, line 75: Could this be an [`ObjectReference`](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-reference/) or [`TypedLocalObjectReference`](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/typed-local-object-reference)?

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -1,3 +1,202 @@
...
+```yaml
+status:
+  artifacts:
+    # oneOf(source,image,config, object)
+    - source:
+        # the sha256 of the ordered JSON of all other non-from fields
+        id: <SHA:string>
+        # exposed fields - in this case url and revision
+        uri: <:string>
+        revision: <:string>
```
evankanderson, line 62: Should these be top-level fields, or possibly under an `outputs` key?

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -1,3 +1,202 @@
...
+          apiVersion: <:string>
+          # name of the resource on the cluster
+          name: <:string>
+          # namespace of the resource
+          namespace: <:string>
+          # resource version of the object
+          resourceVersion: <:string>
+        from:
+          - # id of any artifact(s) that were inputs to the template
+            id: <SHA:string>
```
evankanderson, line 78: How does Cartographer track the `from` information -- does it require that each resource expose an `observedGeneration` or equivalent in `status`? If not, Cartographer may observe a resource's `status` before it has reconciled a `spec` change and assume that the `status` goes with the current `spec`, when it actually applies to an older `spec`.

rfc/rfc-0018-workload-report-artifact-tree.md
evankanderson: There is also a 1MB limit on the size of a single resource, and the values might be repeated twice due to the addition of `managedFields`. I'd be careful about including large other documents directly for that reason. It seems like it would be better to include a `resourceVersion` or `generation` field referencing the resource's version, rather than the entire contents.

rfc/rfc-0018-workload-report-artifact-tree.md
```diff
@@ -1,3 +1,202 @@
...
+        revision: b974272e27c47a01e7a7da07cf8e4415bdb83dae
+        resource:
+          resource-name: source-tester
+          kind: Runnable
+          apiVersion: carto.run/v1alpha1
+          name: my-app
+          namespace: my-namespace
+          resourceVersion: "11125545"
+        from:
+          - id: 23156c7ac2170fe95f85a1ad42522c408e67038f8393eea6cd8551e07457c5d7 # <--- The id noted above, as the source-tester consumes the source artifact from source-provider
```
evankanderson, line 131: If a component like `kpack` can generate a new build based on external stimulus (i.e. addition of a new base image), what happens to `from` here?

evankanderson (commented):
A few notes, but I'd be super-excited to even start with the list of resources / stages / artifacts managed by the Workload, and add in the "triggered by" separately.

rfc/rfc-0018-workload-report-artifact-tree.md
evankanderson: > > As updating the Workload status will cause the Workload to be reprocessed, which will cause the templated resource to be updated, even if that update is a noop the resourceVersion will bump
> 
> When the workload reconciles, Cartographer first checks for changes to a stamped object. If there is no proposed change to the spec, Cartographer will not update the object. This should protect us from the infinite loop scenario posited.

I think Scott's point is that `resourceVersion` can change even if the underlying resource's `spec` does not. For CRDs, the `metadata.generation` field tracks `spec` and `metadata` changes (but not `status` changes).

PR #524, opened 2022-01-18
Use latest for ytt

emmjohnson:
Co-authored-by: Marty Spiewak <mspiewak@vmare.com>

closes # 

## Release Note

## PR Checklist

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -25,9 +25,6 @@ readonly REGISTRY=${REGISTRY:-"$($ROOT/hack/ip.py):5000"}
 readonly BUNDLE=${BUNDLE:-$REGISTRY/cartographer-bundle}
 readonly RELEASE_DATE=${RELEASE_DATE:-$(TZ=UTC date +"%Y-%m-%dT%H:%M:%SZ")}
 
-readonly YTT_VERSION=0.36.0
-readonly YTT_CHECKSUM=d81ecf6c47209f6ac527e503a6fd85e999c3c2f8369e972794047bddc7e5fbe2
-
 main() {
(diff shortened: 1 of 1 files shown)


scothis (changes requested):
While ytt has strong history of backwards compatible releases, that won't be true forever. Reproducible builds within cartographer is more important than having the latest bits.

martyspiewak:
@scothis The counter-pressure is that old versions of ytt have unpatched CVEs (due to the version of Go they are compiled with) that the later versions address.

As a compromise, do you think it'd be ok to pin to a major version of ytt? Is it fair to assume that if there is a breaking change they would bump major versions? We are talking about moving to using the ytt library instead of the binary (#526) and so we could use dependabot to pin to major versions but still get updates.

scothis:
>The counter-pressure is that old versions of ytt have unpatched CVEs (due to the version of Go they are compiled with) that the later versions address.

I agree with the motivation, I disagree with the implementation

> As a compromise, do you think it'd be ok to pin to a major version of ytt? 

There are a few principles that drive my thinking:
1. every build of Cartographer is pinned to a specific version of ytt, this is fundamental to reproducible builds.
2. changes to the version of ytt are PRed and tested (like every other dependency bump, you can use Dependabot)
3. release branches can stay on an older, more stable version of ytt (if there are CVEs in an incompatible version, we can take that up with the Carvel team)

scothis:
Another option to solve the immediate issue of the go compiler version is to `go build` the ytt source at a specific version and use that as the binary.

As you called out, consuming as a library would also work. My initial spike adding ytt support for templates used ytt as a go module rather than an executable.

PR #570, opened 2022-01-29
RFC Proposal: Introduce Multiple-Paths

squeedee:
Proposal for multiple paths in a supply chain by extending Options.
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-multiple-paths/rfc/rfc-0000-multiple-paths.md)

The whiteboard I've used for visual representations: https://miro.com/app/board/uXjVOTzSDSQ=/


rfc/rfc-0000-multiple-paths.md
```diff
@@ -0,0 +1,241 @@
...
+
+OR
+
+> It matches on labels and fields (if no selector is specified, that is a match)
+
+AND
+
+> It's inputs are fulfilled using these same rules.
+
+Options are a priority list to disambiguate inputs that match in more than one case. It should still be possible
```
waciumawanjohi, line 160: This seems to revert a recent design decision.

In RFC 9 we specify that only 1 option can match. For example, given this workload and supply-chain, an error would be thrown because the conditions for both options are satisfied:

```yaml
kind: Workload
metadata:
  labels:
    is-green: "true"
    is-red: "true"

---
kind: ClusterSupplyChain
spec:
  resources:
    - name: some-step
      templateRef:
        kind: ClusterSourceTemplate
        options:
          - name: option-1
            selector:
              matchFields:
                { key: "metadata.labels.is-green", operation: in, values: ["true"] }
          - name: option-2
            selector:
              matchFields:
                { key: "metadata.labels.is-red", operation: in, values: ["true"] }
```

Making options a priority list reverts that behavior. We see that this same workload paired with a similar supply-chain would now succeed:

```yaml
kind: Workload
metadata:
  labels:
    is-green: "true"
    is-red: "true"

---
kind: ClusterSupplyChain
spec:
  resources:
    - name: some-red-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-red-template
      selector:
        matchFields:
          { key: "metadata.labels.is-red", operation: in, values: ["true"] }

    - name: some-green-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-green-template
      selector:
        matchFields:
          { key: "metadata.labels.is-green", operation: in, values: ["true"] }

    - name: some-final-step
      templateRef:
        kind: ClusterSourceTemplate
        options:
          - name: option-1
            sources:
              - resource: some-red-step
                name: source
          - name: option-2
            sources:
              - resource: some-green-step
                name: source
```

The `match one and only one option` rule makes the supply-chain a little more verbose. For example, the image-builder step would become

```yaml
- name: image-builder
  templateRef:
    kind: ClusterImageTemplate
    options:
      - name: build-image
        images:
          - resource: source-tester
            name: source
      - name: build-image
        selector:
          matchFields:
            { key: "metadata.labels.has-test", operation: NotIn, values: ["true"] }  # <--- new selector necessary
        images:
          - resource: source-provider
            name: source
```

rfc/rfc-0000-multiple-paths.md
```diff
@@ -0,0 +1,241 @@
...
+
+> It's the only option and all other criteria are met.
+
+OR
+
+> It matches on labels and fields (if no selector is specified, that is a match)
+
+AND
+
+> It's inputs are fulfilled using these same rules.
```
waciumawanjohi, line 158: I would call this a transitive condition. It is certainly determinitive. But it is difficult to reason about. It seems a reasonable goal that a user looking at a supply-chain can easily determine how a workload would flow through it. Needing to notate choices at earlier nodes, and use those to reason about what choice will be made at a new node is tricky.

This can be alleviated in a few ways.

1. Remove transitive conditions.
2. Require that there is a visualization tool for supply chains. (Fortunately, Rash has already prototyped one!) Such a tool can make implicit logic in the supply-chain explicit in the visualization.
3. Require a mutating webhook that would make all of the implicit transitive conditions explicit conditions at later steps. For example, taking the `is-green` and `is-red` example:

original:
```yaml
kind: ClusterSupplyChain
spec:
  resources:
    - name: some-red-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-red-template
      selector:
        matchFields:
          { key: "metadata.labels.is-red", operation: in, values: ["true"] }

    - name: some-green-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-green-template
      selector:
        matchFields:
          { key: "metadata.labels.is-green", operation: in, values: ["true"] }

    - name: some-final-step
      templateRef:
        kind: ClusterSourceTemplate
        options:
          - name: option-1
            sources:
              - resource: some-red-step
                name: source
          - name: option-2
            sources:
              - resource: some-green-step
                name: source
```

transformed by a mutating webhook:
```yaml
kind: ClusterSupplyChain
spec:
  resources:
    - name: some-red-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-red-template
      selector:
        matchFields:
          { key: "metadata.labels.is-red", operation: in, values: ["true"] }

    - name: some-green-step
      templateRef:
        kind: ClusterSourceTemplate
        name: some-green-template
      selector:
        matchFields:
          { key: "metadata.labels.is-green", operation: in, values: ["true"] }

    - name: some-final-step
      templateRef:
        kind: ClusterSourceTemplate
        options:
          - name: option-1
            sources:
              - resource: some-red-step
                name: source
            selector:        # <--- This field can be added by a mutating webhook
              matchFields:
                { key: "metadata.labels.is-red", operation: in, values: ["true"] }
          - name: option-2
            sources:
              - resource: some-green-step
                name: source
            selector:        # <--- This field can be added by a mutating webhook
              matchFields:
                { key: "metadata.labels.is-green", operation: in, values: ["true"] }
```

rfc/rfc-0000-multiple-paths.md
squeedee: Image builder being a source tester is obnoxious to me, that's why I prefer to support optional resources.

rfc/rfc-0000-multiple-paths.md
squeedee: For 2., wrt vizualization. We don't even need that, just syntax matching that lets you see/jump to definitions.

I don't think it's a particularily heavy ask to suggest that an input matching upstream is a necessity for an option to play out. Transitive dependencies tend to encapsulate their purpose, eg: I (a resource) depend on the source provider, so clearly I wont run if there is no source. I might need to add an `option` to depend on the image-provider, or I might realize I cannot do anything with images, so I do not depend on it. Somewhat OO in its design, this implies that an upstream resource knows precisely what's needed for it to exist, and you do not need to go prying into its affairs to know wether you need it's output or not. 

Or to state this another way, I (a resource) don't care what the upstream resource selects for, i only care to consume it when it's available)

rfc/rfc-0000-multiple-paths.md
squeedee: See my comment below about the object encapsulation principle I've used throughout this spec. A resource represents a rigid purpose. `<<interface>> IBuildImages`

Options decorate concrete implementations "<<template>> ImageBuilder" with a selection criteria. 

One purpose is to choose different concrete implementations (eg, `<<template>> GitPusher` and `<<template>> ImagePusher`

The other purpose is to choose the inputs (eg as show in build-image in the example)

These two different purposes for options can be teased apart, but it looks kinda silly because it introduces the `or` syntax that I feel is superfluous (it is essentially the option to add an "or" operator as discussed elsewhere, but it get's very messy when matching Inputs and Fields And Labels.. Trying to absorb that much algebra into a yaml spec is unwieldy) 

So for that reason, I left options selecting either *or both* the concrete class (template) and configuring the inputs.

rfc/rfc-0000-multiple-paths.md
waciumawanjohi: I think all of that is correct. I was bringing up the user experience of reading a supply chain and trying to reason through the path a workload would take. I suspect that's a reasonable goal.

squeedee: Which is precisely the picture I'm trying to paint. There are trade offs in complex systems between top down and bottom up comprehension. Being able to look at a single component (bottom up) and know why/how it was selected is simply a matter of "what it selects for", not "also what it's predecessors selected for". I don't look at a flow chart and expect to see each decision diamond to contain all the previous choices made. That's when I switch to top down (I look at the whole chart). 

Top down comprehension **is** degraded by multi-pathing, but not by the nature of this RFC. It's the nature of complex dependencies presented in lists/recordsets, especially those with fan-in behavior. If not for multiple input nodes (fan in) we'd be able to use indentation and nest objects under each other, giving us a wonderful visual clue in the document structure, but we don't have that luxury. 

Denormalizing selectors (as proposed above) does not help with visualizing paths either, it is also a bottom-up solution.

I'd rather we find clear ways of annotating paths through a supply chain. We can (if we uphold determinism) annotate a workload (in the status!) with a reduced list of resources from the chosen path (just like RFC 20's artifact mapping). 

Even without multi-pathing, visualization and 'go to definition' tooling is a valuable addition, so we should *ensure we do that*.

martyspiewak (approved):

emmjohnson (approved):

idoru (approved):

jwntrs (approved):

cirocosta (approved):

squeedee:
Shall we move this to `declined`?

squeedee:
Closing as declined.

This approach empowers supply chain authors to treat Cartographer more like an orchestrator, with a lot of `If-this-then-that` support, vs pushing imperative behavior into specific resources. We'd like to dissuade this kind of thinking.

Issue #436, opened 2021-12-07
Separate Site and Controller repositories


squeedee:
### Description of problem

1. Site content changes at a different rate than the code
2. We often need to update versioned docs separate from the tag for that version (site is always built from a single branch, namely `main`)
3. Waiting for code checks on PRs when the only thing we've changed is documentation, is a significant pain
4. If we do end up creating a multi-product site (Eg, introduce runnable as it's own project) then we'll be out ahead of the infra changes needed (namely, seperating the Site repo :D)

### Plan:

* [X] create cartographer-site repo  [DONE](https://github.com/vmware-tanzu/cartographer-site)
* [x] Place all migration work on a branch [initial-migration](https://github.com/vmware-tanzu/cartographer-site/tree/initial-migration)
  * [x] ensure all branches for site work is merged
  * [x] Move experimental editor over and patch the release code
  * [x] Deal with the CRD generation code so it bases itself off the latest release in the cartographer repository
  * [x] call for a moratorium on site PRs (explain why) in Slack, and put a [Blocked on Migration] tag on all related doc stories.
  * [x] Ensure Linter runs in github actions
  * [x] Ensure dependabot is configured
  * [x] Configure require PR approval
* [x] Make PR for initial-migration and get it merged
* prepare cartographer PR:
  * [x] Clean up any call-through makefile targets in the cartographer makefile (they will no longer be able to reach the site) (dont merge!)
  * [x] remove site and experimental live-editor from cartographer repo (Don't merge!)
* [x] Have netlify use the new repository
* [x] Merge cartographer PR

* [x] unblock stories and let Slack know that the migration is complete

scothis:
An intermediate option between same branch and different repos is to move website publishing to the gh-pages branch. It does mean that PRs need to be more considerate about which branch is the base.

squeedee:
> An intermediate option between same branch and different repos is to move website publishing to the gh-pages branch. It does mean that PRs need to be more considerate about which branch is the base.

Do you propose this because of the pain of obtaining another repository? Otherwise it's pretty slight for me - in fact I'd prefer the separation of concerns and the effort involved over the weirdness of a branch.

Would love for these steps to also help ease the plan, up front, for adopting other subsystems.

scothis:
> Do you propose this because of the pain of obtaining another repository?

Nope, getting a new repo should be easy. Multi-repo vs mono-repo is always a balancing act

squeedee:
[...]

PR #571, opened 2022-02-01
wip: RFC to move away from using current repo as a monorepo

rawlingsj:
This is an initial proposal taking onboard recent discussions on how we build the wider cartographer ecosystem.  It's early stages and aimed to solicit feedback of the general direction going forward.

[RFC Text](https://github.com/vmware-tanzu/cartographer/blob/d884618df7cb94915c1d84eb94122bb76fa81c54/rfc/rfc-0000-move-off-monorepo.md)

### DIFF
--- a/rfc/rfc-0000-move-off-monorepo.md
+++ b/rfc/rfc-0000-move-off-monorepo.md
@@ -1,0 +1,57 @@
+# Draft RFC <0000> move off monorepo
+
+## Summary
+
+Following the [convention proposal](https://github.com/vmware-tanzu/cartographer/pull/514) and [office hours discussion](https://www.youtube.com/watch?v=iCURQqV52Uw&t=1173s) there was a decision to not include conventions in the https://github.com/vmware-tanzu/cartographer repository as using the current cartographer repo as a monorepo would cause challenges around:
+
+- extra support burden on the current maintainers
+- maintaining development and style consistency
+- existing CI, release automation and GitHub tooling doesn't work well with monorepos
+- as we continue to build up the Cartographer ecosystem with integrations, adaptors, platform controllers this will cause a bottleneck for changes proposed on a single repo.
+
+Adopting a microservice approach for Cartographer will aid extensions and help build up the wider ecosystem and community around it.
+This RFC proposes how we will adopt this microservice style approach that will help grow the Cartographer ecosystem, encourage new integrations and extensions is a sustainable and maintainable way that suits the project and its aspirations.
+
+## Motivation
+
+The motivation behind Cartographer as an OSS project aims to build a community and ecosystem to aid supply chain choreography.  Cartographer itself is an "umbrella" for this ecosystem, the first step of this was to build the core components and now we are looking ahead to extend by adding more capabilities.
+ 
+As expressed by maintainers of the https://github.com/vmware-tanzu/cartographer repository it will become hard to manage it if we use it as a monorepo and continue to add more capabilities like conventions etc.
+It is expected that this RFC describes how we split the current repository so it allows the Cartographer OSS project to evolve and scale beyond the concrete capabilities the core aims to provide.
+For example splitting out from the current repo:
+
+-  the __`Runnable`__ feature so it can release independently and install standalone in a target deployment cluster,  avoids deploying all of Cartographer which contains build cluster CRDs not desired in runtime clusters
+- the __examples directory__ so we can build up a wide range of examples that include integrations with wider platform services like conventions.  This also helps encourage contribution of community supply chains for different integrations where we can add automated testing that runs against different setups.
+- the __website__ so we can add documentation for these new integrations and platform capabilities outside of the core repository.  I.e. Documenting services like the conventions or a Jenkins integration should not require a change to the core cartographer repository, this should live outside in its own repository.
+
+## Possible Solutions
+
+There seems to be only two options possible solutions:
+- __monorepo__ we could revisit the [convention proposal](https://github.com/vmware-tanzu/cartographer/pull/514) decision and look at ways to address the monorepo concerns however there seemed a general consensus on the [office hours call](https://www.youtube.com/watch?v=iCURQqV52Uw&t=1173s) that this was not desired.
+- __microservice__ As conventions are now being worked on in a different repository, we should look at splitting out parts that currently reside in https://github.com/vmware-tanzu/cartographer.
+
+Suggestions:
+
+-  __`Runnable`__ feature so it can install standalone in a target deployment cluster and avoids deploying all of Cartographer
+-  __examples directory__ so we can build up a wide range of examples that include integrations with wider platform services like conventions.
+-  __website__ so we can add documentation for these new integrations and platform capabilities outside of the core repository.  I.e. documenting a service like the convention components or a Jenkins integration should not require a change to the core cartographer repository, this should live outside in it's own repository
+- __others?__
+
+## Cross References and Prior Art
+
+Carvel: https://github.com/vmware-tanzu - Corporate sponsored organisation
+Kubernetes: https://github.com/kubernetes - Cloud Native Computing Foundation sponsored organisation
+Tekton: https://github.com/tektoncd - Continuous Delivery Foundation sponsored organisation
+ 
+# Unresolved Questions
+
+[unresolved-questions]: #unresolved-questions
+
+- Should we keep the rfc folder in the cartographer repo as it should relate to only the core components?
+- Should we move the RFC process outside into its own repository too so that it provides governance across the wider project?
+- Should we consider an enhancement process for the project and keep the current repo level RFC process?
+Note examples of OSS projects that use an enhancement process:
+
+- __KEP__ - https://github.com/kubernetes/enhancements/
+- __TEP__ - https://github.com/tektoncd/community/tree/main/teps
+- __JEP__ - https://github.com/jenkinsci/jep


squeedee:
A related issue https://github.com/vmware-tanzu/cartographer/issues/436

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Possible Solutions
+
+There seems to be only two options possible solutions:
+- __monorepo__ we could revisit the [convention proposal](https://github.com/vmware-tanzu/cartographer/pull/514) decision and look at ways to address the monorepo concerns however there seemed a general consensus on the [office hours call](https://www.youtube.com/watch?v=iCURQqV52Uw&t=1173s) that this was not desired.
+- __microservice__ As conventions are now being worked on in a different repository, we should look at splitting out parts that currently reside in https://github.com/vmware-tanzu/cartographer.
+
+Suggestions:
+
+-  __`Runnable`__ feature so it can install standalone in a target deployment cluster and avoids deploying all of Cartographer
```
squeedee, line 35: Not sure this is a high priority.

1. deploying runnable does mean a fatter controller and unnecessary CRDs today, but that can be mitigated in other ways if required.
2. runnable co-opts a lot of the cartographer runtime, so we'd have to libify cartographer as part of this effort.

I feel like adressing extracting runnable as it's own controller (or the same controller with feature flags to only run where the Runnable CRD is installed) is an orthoganal concern, but this RFC opens up our options for such future discussions.

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+-  __examples directory__ so we can build up a wide range of examples that include integrations with wider platform services like conventions.
```
squeedee, line 36: The examples are a part of our e2e as well, we currently kill two birds with one stone and it's an efficiency I think we want to maintain.

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+-  __website__ so we can add documentation for these new integrations and platform capabilities outside of the core repository.  I.e. documenting a service like the convention components or a Jenkins integration should not require a change to the core cartographer repository, this should live outside in it's own repository
```
squeedee, line 37: This one resonates for me, hence the [story](https://github.com/vmware-tanzu/cartographer/issues/436) I created. As a first step, this enables:
1. Treating the docs as an 'umbrella' site, supporting multiple products in the Cartographer group
2. Making it obvious that docs change at a different rate than any one repository, and not confusing people by being inside tagged versions of any given product.
3. Letting us skip unit, integration and e2e tests for our controllers when iterating on docs.

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+- __others?__
```
cirocosta, line 38: I'm particularly not a fan of having the website layout etc as part of the same codebase as the controllers themselves, but I do find useful the benefits that having documentation in the same repo as the code brings, like the ease of having PRs that update code also updating documentation (naturally, not a huge deal in case we do split it - as long as we ensure that we're always keeping the docs in sync etc).

rfc/rfc-0000-move-off-monorepo.md
cirocosta, line 37: I'm particularly not a fan of having the website layout etc as part of the same codebase as the controllers themselves, but I do find useful the benefits that having documentation in the same repo as the code brings, like the ease of having PRs that update code also updating documentation (naturally, not a huge deal in case we do split it - as long as we ensure that we're always keeping the docs in sync etc).

more concretely:

1. I think it's great that if we change something meaningful in the architecture, that we'd have a change coming along in [architecture.md](https://github.com/vmware-tanzu/cartographer/blob/main/site/content/docs/development/runnable/architecture.md) which would fall under the same tag as the one where the code lands, etc
2. not great that we maintain in the same repo things like [header.html](https://github.com/vmware-tanzu/cartographer/blob/ddf2477d50c40d8c7a8b39a1a48ac2c7f6616fc8/site/themes/template/layouts/partials/header.html)

personally, I'd tend towards trying to keep **content** (plain markdown, like https://github.com/vmware-tanzu/carvel-ytt/tree/develop/docs) in this repo, but all the website structure etc in another if possible (e.g, see https://github.com/vmware-tanzu/carvel).

wdyt?

rfc/rfc-0000-move-off-monorepo.md
cirocosta, line 35: _at the moment_, runnable is a controller that is ran as part of the same process that brings up the other core controllers (differently from the example given of Convention service) - if we were to split Runnable into its own repository, do you think we should then make it its own controller? My understanding is that _yes_, it should (given the use of the term "microservices" throughout the proposal).

rfc/rfc-0000-move-off-monorepo.md
squeedee: I have seen no evidence that the docs move at the same pace as our controllers. We often need to work through a range of changes to make updates to most of the docs viable. The one place where doc's should be in lock step with changes, is changes to the API (Reference section). Hence my effort to make sure that API documentation is inline (`pkg/api/....` doc comments). During the docs split, we'll need to rework the code that pulls in api docs into the Reference section (pretty easy) and make sure there is a process to ensure we run that before creating a new Doc Version tag for each controller.

rfc/rfc-0000-move-off-monorepo.md
waciumawanjohi: I agree with Rash. It's not clear that "the juice is worth the squeeze" in terms of separating Runnable into a separate repo. As he mentions, there would be non trivial engineering work because Runnable shares code with the other parts of Cartographer. Do we have evidence that
1. A user would like to use Runnable without using supply-chains/delivery.
2. That user is concerned about the overhead of installing supply-chains/delivery with Runnable.

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+- the __website__ so we can add documentation for these new integrations and platform capabilities outside of the core repository.  I.e. Documenting services like the conventions or a Jenkins integration should not require a change to the core cartographer repository, this should live outside in its own repository.
+
+## Possible Solutions
+
+There seems to be only two options possible solutions:
+- __monorepo__ we could revisit the [convention proposal](https://github.com/vmware-tanzu/cartographer/pull/514) decision and look at ways to address the monorepo concerns however there seemed a general consensus on the [office hours call](https://www.youtube.com/watch?v=iCURQqV52Uw&t=1173s) that this was not desired.
+- __microservice__ As conventions are now being worked on in a different repository, we should look at splitting out parts that currently reside in https://github.com/vmware-tanzu/cartographer.
+
+Suggestions:
```
scothis, line 33: Would it also make sense to split ClusterSupplyChain/Workload from ClusterDelivery/Deliverable? These components are highly likely to be installed into distinct clusters.

rfc/rfc-0000-move-off-monorepo.md
```diff
@@ -0,0 +1,57 @@
...
+
+-  __`Runnable`__ feature so it can install standalone in a target deployment cluster and avoids deploying all of Cartographer
+-  __examples directory__ so we can build up a wide range of examples that include integrations with wider platform services like conventions.
+-  __website__ so we can add documentation for these new integrations and platform capabilities outside of the core repository.  I.e. documenting a service like the convention components or a Jenkins integration should not require a change to the core cartographer repository, this should live outside in it's own repository
+- __others?__
+
+## Cross References and Prior Art
+
+Carvel: https://github.com/vmware-tanzu - Corporate sponsored organisation
+Kubernetes: https://github.com/kubernetes - Cloud Native Computing Foundation sponsored organisation
```
scothis, line 43: Kubernetes is a curious case, as it is developed as a mono repo and consumed as component repos.

rfc/rfc-0000-move-off-monorepo.md
waciumawanjohi: > These components are highly likely to be installed into distinct clusters.

That is true. Do we have a reasonable expectation that users would object to installing both CRDs in a cluster only to leverage one? Would the value delivered match the work of separating the code? (As they do share code)

rfc/rfc-0000-move-off-monorepo.md
squeedee: wouldn't we _have to_? I don't see a choice once it's extracted into another repo.

rfc/rfc-0000-move-off-monorepo.md
squeedee: Please see https://github.com/vmware-tanzu/cartographer/pull/571#discussion_r796704119 wrt priorities for this work. I think ultimately it would be nice, but until we see driving factors as @waciumawanjohi suggests, it's not a _need_.

rfc/rfc-0000-move-off-monorepo.md
rawlingsj: There's been some feedback that a community maintained set of examples that live on their own repo would aid contributions for wider integrations and scenarios.  @vrabbi did you have some thoughts on this?

Once we add conventions and say Jenkins plus a growing number of other integrations it would be good if they existed in a separately maintained repo we can add more automation beyond the scope of the core cartographer repo.

rfc/rfc-0000-move-off-monorepo.md
rawlingsj: I've certainly seen pushback from end users for installing build related components (including CRDs) into a target run cluster.  Having a standalone / microservice installation for this makes sense to me.

rfc/rfc-0000-move-off-monorepo.md
rawlingsj: yeah I guess I was trying to show that there's a wider kubernetes ecosystem that helped evolve the project.  I.e specific areas developed special interest groups that iterate at their own cadence.  That's quite a good model to help encourage wider efforts and involvement from others.

rfc/rfc-0000-move-off-monorepo.md
vrabbi: I think that having an examples repo seperate is of huge value. I domt think e2e needs to be run against all examples so having core functionailty for testing in the main repo makes sense but by having an additional repo only for examples, it can server as a central location for examples of using different integrations like kaniko,jenkins,argo etc.

vrabbi: And having that under the cartographer umbrella is key i believe from a community perspective to show that cartographer truly is pluggable by design. If its in an official cartographer repo it gains more credability.

rfc/rfc-0000-move-off-monorepo.md
davidmirror-ops: A further discussion (probably for a separate RFC) would deal with how splitting into separate repos could trigger changes in Governance, introducing the idea of Working Groups/Feature areas (see [Harbor](https://github.com/goharbor/community/blob/main/MAINTAINERS.md ) as an example). With that approach we could define focus areas for maintainers and potentially onboard contributors for 'microteams' with a clear scope.

rfc/rfc-0000-move-off-monorepo.md
martyspiewak: Would it be that much effort to split out runnable? In terms of the underlying code that's shared and called from the reconciler, it's all libified already, no? The runnable repo can still import `github.com/vmware-tanzu/cartographer/pkg/...` and I would think not much would change.

squeedee: In code, no. In docs? maybe, in tooling, plenty? The argument isn't about the size and shape of the work, the argument is the need to even consider it at this point. Perhaps I'm being too "agile" in wanting to focus on immediate pain points, of which there are enough to accept work to split out docs and be prepared for creating new sub-projects.

rawlingsj: Perhaps we should keep this RFC focussed on what our desired end state is and keep prioritisation a separate discussion?  That would help understand what's involved and encourage community contributions if folks felt it important to them?

PR #588, opened 2022-02-04
github: add scan to release workflow (push to main)

cirocosta:

rather than having the scan of both source and image running for _just_
PRs, have it also as part of commits that get pushed to `main` / tagged commits.

this way older PRs that have been accepted but not yet merged get a
second chance of having their contents scanned as they get into main.

ps.: the workflow is called _release_ because yeah, we _release_ (as a draft)
every commit that lands on `main`:

https://github.com/vmware-tanzu/cartographer/blob/3bf25839b8f726abcef4fc8bb0d5305ec2055e79/.github/workflows/release.yaml#L17-L19

as opposed to validation that's only for PRs

https://github.com/vmware-tanzu/cartographer/blob/3bf25839b8f726abcef4fc8bb0d5305ec2055e79/.github/workflows/validation.yaml#L16-L18

## Release Note

None - internal only.

## PR Checklist

- ~Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`~
- ~Filled in the [Release Note](#Release-Note) section above 
- ~Modified the docs to match changes


.github/workflows/release.yaml
```diff
@@ -95,6 +95,14 @@ jobs:
           ./hack/docker-login.sh
           ./hack/release.sh
 
+      - name: scan source
+        run: |-
+          grype .
+
+      - name: scan image
+        run: |-
```
cirocosta, line 63: differently from the `validation` workflow, here we don't need to point at a `DOCKER_CONFIG` because the images are all pushed to DockerHub, which is public (thus, anon auth is fine)

martyspiewak: Does this mean we can get rid of lines 61-62? Or is this no longer true and that is why it was marked resolved?

martyspiewak (changes requested):

martyspiewak:
I think #718 already added this behavior, could be missing something though.

PR #602, opened 2022-02-10
RFC: Add matchFields and matchLabels to top level of blueprints

jwntrs:
Add matchFields and matchLabels to top level of blueprints
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc/0000-blueprint-match-fields/rfc/rfc-0000-blueprint-match-fields.md)

### DIFF
--- a/rfc/rfc-0000-blueprint-match-fields.md
+++ b/rfc/rfc-0000-blueprint-match-fields.md
@@ -23,7 +23,7 @@ The [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/7
 [what-it-is]: #what-it-is
 
 ```yaml
-apiVersion: carto.run/v1alpha1
+apiVersion: carto.run/v1alpha2                                 # <======== bump this
 kind: ClusterSupplyChain
 metadata:
   name: supply-chain
@@ -82,7 +82,7 @@ This is a breaking change.
 [spec-changes]: #spec-changes
 
 ```yaml
-apiVersion: carto.run/v1alpha1
+apiVersion: carto.run/v1alpha2                                  # <==== bump this
 kind: ClusterSupplyChain
 metadata:
   name: supply-chain
--- a/rfc/rfc-0000-blueprint-match-fields.md
+++ b/rfc/rfc-0000-blueprint-match-fields.md
@@ -1,0 +1,106 @@
+# Meta
+[meta]: #meta
+- Name: Blueprint selector matchFields
+- Start Date: 2022-02-10
+- Author(s): @jwntrs
+- Status: Draft
+- RFC Pull Request:
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) introduced template `options` that included a `matchFields` selector. We should include this same selector at the `blueprint` level
+
+# Motivation
+[motivation]: #motivation
+
+- Allow entire `blueprints` to match an `owner` based on certain fields.
+
+- Ensure that the behaviour of all `selectors` is consistent.
+
+# What it is
+[what-it-is]: #what-it-is
+
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    matchFields:                                                # <=========== add this
+      { key: "workload.labels.pipeline", operation: exists }    # <=========== 
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+The same way as `matchFields` from [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75)
+
+# Migration
+[migration]: #migration
+
+This is most likely a breaking change which would require a bump to the `ClusterSupplyChain` `apiVersion` and introducing a conversion webhook.
+
+```yaml
+  selector:
+    app.tanzu.vmware.com/workload-type: web
+```
+
+needs to become:
+
+```yaml
+  selector:
+    matchFields:
+      - key: workload.metadata.labels."app.tanzu.vmware.com/workload-type"
+        operator: In
+        values: ["web"]
+```
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a breaking change.
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we introduced `matchLabels` at the same time as `matchFields`, when migrating we would simply need to hoist everything under the current `selector` over to the new `matchLabels` field.
+
+```yaml
+  selector:
+    app.tanzu.vmware.com/workload-type: web
+```
+
+becomes
+
+```yaml
+  selector:
+    matchLabels:
+      app.tanzu.vmware.com/workload-type: web
+```
+
+
+# Prior Art
+[prior-art]: #prior-art
+
+[template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) matchFields
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    matchFields:                                                # <=========== add this
+      { key: "workload.labels.pipeline", operation: exists }    # <=========== 
+```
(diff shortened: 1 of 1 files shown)


squeedee:
One of the asks from the OSS slack channel was for clarity in naming. I think the name or at least title for this PR and for the RFC might be better as 
`Add MatchFields to top level of blueprints`?

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+
+# Prior Art
+[prior-art]: #prior-art
+
+[template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) matchFields
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
```
scothis, line 97: Since a workload may only be processed by exactly one supply chain, does this proposal help/hinder the best match logic? On the surface it would seem to make that logic more complicated, but perhaps its a wash.

jwntrs: I think it does make it more complicated, but it's not unmanageable. Right now a best-match is indicated by the number of labels. We could broaden this definition to include the number of labels and number of fields matched.

Good call out, i'll add this as ~~an open question~~ part of how it works.

squeedee (approved):

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,110 @@
...
+
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    matchFields:                                                # <=========== add this
+      { key: "workload.spec.image", operation: exists }         # <===========
```
scothis: since the selector here is a transitional label selector, I'd expect `matchFields` to operate like a traditional field selector. That is, to operate against the content of the Workload resource directly.

The equivalent `matchFields` proposal from RFC 9 in #75 is different since that operates on the templating context. That semantic difference may cause enough user confusing to negate the benefits of offering a field selector at this level.

```suggestion
      { key: "spec.image", operation: exists }                  # <=========== 
```

jwntrs: > RFC 9 in https://github.com/vmware-tanzu/cartographer/pull/75 is different since that operates on the templating context

jwntrs: > RFC 9 in https://github.com/vmware-tanzu/cartographer/pull/75 is different since that operates on the templating context

It doesn't operate on the templating context. It only operates on the workload at the moment. The syntax included the `workload` prefix to leave room for other things like params. However, since the direction from the [RFC for adding params to matchFields](https://github.com/vmware-tanzu/cartographer/pull/618) is moving towards having a top level matchParams selector, perhaps we should amend [RFC9](https://github.com/vmware-tanzu/cartographer/pull/75) to remove the workload prefix.  @emmjohnson @martyspiewak do you have thoughts on this?

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a breaking change.
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we introduced `matchLabels` at the same time as `matchFields`, when migrating we would simply need to hoist everything under the current `selector` over to the new `matchLabels` field.
```
martyspiewak: I would prefer having both `matchFields` and `matchLabels`, rather than forcing users to refer to labels through the `matchFields` key.

There are existing k8s resources (like Deployment) that use `selector.matchLabels` and I would want to keep that convention.

Also, using `matchFields` for labels would be more painful because you'd have to define the operator as well, rather than just using regular regular selection.

emmjohnson: +1

rfc/rfc-0000-blueprint-match-fields.md
martyspiewak: I think I'd be in favor of removing `workload` from the bottom level selectors (i.e. amending/superseding RFC9)

rfc/rfc-0000-blueprint-match-fields.md
jwntrs: updated the RFC to reflect adding both top level selectors

rfc/rfc-0000-blueprint-match-fields.md
jwntrs: removed workload from top level selector syntax

martyspiewak (approved):

squeedee (approved):

sclevine (approved):

waciumawanjohi (approved):

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,95 @@
...
+apiVersion: carto.run/v1alpha2                                 # <======== bump this
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    matchLabels:                                                # <=========== move existing labels under this heading
+      app.tanzu.vmware.com/workload-type: web
+    matchFields:                                                # <=========== add this
+      - { key: ".spec.image", operation: exists }               # <===========
```
cirocosta:
```suggestion
      - { key: "spec.image", operation: exists }               # <=========== 
```


not sure if worth calling out or not, but for template switching we end up going without `.` in the beginning of the jsonpath (e.g., `spec.image` rather than `.spec.image`), which is actually consistent with other core k8s resources ala daemonset:

```yaml
nodeAffinity:
  requiredDuringSchedulingIgnoredDuringExecution:
    nodeSelectorTerms:
    - matchFields:
      - key: metadata.name
        operator: In
        values:
        - target-host-name
```

(see https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#scheduled-by-default-scheduler)

cartographer implementation for template switching: https://github.com/vmware-tanzu/cartographer/blob/3824091dd70ea53679d12ef380bd898cb1ec50d4/pkg/apis/v1alpha1/workload.go#L61-L69

cirocosta (approved):
lgtm!

rfc/rfc-0000-blueprint-match-fields.md
jwntrs: @cirocosta makes sense. FWIW I'm also proposing that we remove the workload prefix: https://github.com/vmware-tanzu/cartographer/pull/653

sclevine (changes requested):
Needs v1alpha1 schema

sclevine (approved):

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,114 @@
...
+apiVersion: carto.run/v1alpha2                                  # <======== v2 spec
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    matchLabels:                                                # <=========== move existing labels under this heading
+      app.tanzu.vmware.com/workload-type: web
+    matchFields:                                                                             # <=========== add this
+      - { key: "spec.image", operation: exists }                                             # <===========
```
cirocosta:
```suggestion
      - { key: "spec.image", operator: Exists }                                             # <=========== 
```

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,114 @@
...
+```yaml
+apiVersion: carto.run/v1alpha1                                  # <======== v1 spec
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    app.tanzu.vmware.com/workload-type: web
+  selectorMatchFields:                                                                       # <=========== add this
+    - { key: "spec.image", operation: exists }                                               # <===========
```
cirocosta:
```suggestion
    - { key: "spec.image", operator: Exists }                                               # <=========== 
```

rfc/rfc-0000-blueprint-match-fields.md
```diff
@@ -0,0 +1,114 @@
...
+```yaml
+apiVersion: carto.run/v1alpha1                                 # <======== v1 spec
+kind: ClusterSupplyChain
+metadata:
+  name: supply-chain
+spec:
+  selector:
+    app.tanzu.vmware.com/workload-type: web
+  selectorMatchFields:                                                                       # <=========== add this
+    - { key: "spec.image", operation: exists }                                               # <===========
```
cirocosta:
```suggestion
    - { key: "spec.image", operator: Exists }                                               # <=========== 
```

PR #607, opened 2022-02-11
Add options for a resource in a supply chain

### LINKED ISSUE #573: Provide multiple templates for a particular resource in my blueprint
### Description of problem

As a supply chain author, I want to provide multiple templates for a particular resource in my blueprint, where each option is selected based on certain characteristics of my resource owner.

### Proposed solution

Implement template `options` as per [RFC 9](https://github.com/vmware-tanzu/cartographer/pull/75)

### Example

```yaml
kind: ClusterSupplyChain
metadata:
  name: responsible-ops
spec:
  selector:
    app: web
  resources:
    - name: source-provider
      templateRef:
        kind: ClusterSourceTemplate
        options:                                  # <--- a list
        - name: git-template
          selector:
            matchFields:
              - key: workload.spec.source.git       # <--- path to field in workload spec
                operator: Exists
        - name: imgpkg-bundle-template
          selector:
            matchFields:
              - key: workload.spec.source.image       # <--- path to field in workload spec
                operator: Exists
```

### Acceptance Criteria

- when exactly one of the selectors matches...then happy path
- when more than one of the selectors matches ...  then things go boom ... with useful message of [....] visible on workload status
- when none of the selectors matches....then things go boom ... with useful message of [....] visible on workload status

- when writing a supply chain that uses conditional options and something is written wrong then......
- ...the criteria is identical
- ...the criteria is a subset
- ....the criteria points to keys that don't exist in spec
- ...the operator is invalid
- ...the yaml is malformed
- ....given an operator type. the appropriate values are not included
- ....options is an empty list
-

emmjohnson:
- A templateRef also accepts options in place of name.
  An option will be chosen based on its selector.
- Clear old tracked resources before re-tracking.

Co-authored-by: Marty Spiewak <mspiewak@vmware.com>

closes #573  

## Release Note

## PR Checklist


pkg/apis/v1alpha1/cluster_supply_chain.go
```diff
@@ -254,27 +144,58 @@ type SupplyChainResource struct {
...
+	// Name of the template to apply
+	// Only one of Name and Options can be specified.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+
+	// Options is a list of template names and Selectors. The templates must all be of type Kind.
+	// A template will be selected if the workload matches the specified Selector.
+	// Only one template can be selected.
+	// Only one of Name and Options can be specified.
+	// +kubebuilder:validation:MinItems=2
```
cirocosta, line 162: Given that the `minitems=2` constraint doesn't show up in `explain`

```
KIND:     ClusterSupplyChain
VERSION:  carto.run/v1alpha1

RESOURCE: options <[]Object>

DESCRIPTION:
     Options is a list of template names and Selectors. The templates must all
     be of type Kind. A template will be selected if the workload matches the
     specified Selector. Only one template can be selected. Only one of Name and
     Options can be specified.

FIELDS:
   name <string> -required-
     Name of the template to apply

   selector     <Object> -required-
     Selector is a field query over a workload resource.
```

**wdyt of including an entry that states such `minitems=2` constraint? e.g.**

```
// A minimum of two options must be specified
```

(yeah, at `k apply -f ...` time we do get the proper message

```scala
The ClusterSupplyChain "sc" is invalid: 
  spec.resources.templateRef.options: 
    Invalid value: 
      0: 
        spec.resources.templateRef.options in body should have at least 2 items
```

but it could be nice for folks to know beforehand as they go through docs / `k explain`)

pkg/apis/v1alpha1/cluster_supply_chain.go
```diff
@@ -254,27 +144,58 @@ type SupplyChainResource struct {
...
-
 type SupplyChainTemplateReference struct {
 	// Kind of the template to apply
 	//+kubebuilder:validation:Enum=ClusterSourceTemplate;ClusterImageTemplate;ClusterTemplate;ClusterConfigTemplate
 	Kind string `json:"kind"`
+
+	// Name of the template to apply
+	// Only one of Name and Options can be specified.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
```
cirocosta, line 155: I noticed that here we're specifying the `MinLength=1` constraint, but in other places like SupplyChainResource we don't

```go
  type SupplyChainResource struct {
          // Name of the resource. Used as a reference for inputs, as well as being
          // the name presented in workload statuses to identify this resource.
          Name string `json:"name"`
```

do you think we should start making use of this more broadly? e.g,. currently it's possible to have a SupplyChainResource with `""` (empty string) set

```yaml
  kind: ClusterSupplyChain
  metadata:
    name: sc
  spec:
    selector:
      foo: bar
    resources:
      - name: ""
        templateRef:
          kind: ClusterSourceTemplate
          name: foo
```

(although it's true that completely not specifying the `name` field is not valid)

might be worth a separate issue? wdyt?

pkg/apis/v1alpha1/workload.go
```diff
@@ -34,11 +36,37 @@ const (
...
-	ServiceAccountErrorResourcesSubmittedReason          = "ServiceAccountError"
 	ServiceAccountSecretErrorResourcesSubmittedReason    = "ServiceAccountSecretError"
 	ResourceRealizerBuilderErrorResourcesSubmittedReason = "ResourceRealizerBuilderError"
+	ResolveTemplateOptionsErrorResourcesSubmittedReason  = "ResolveTemplateOptionsError"
+	TemplateOptionsMatchErrorResourcesSubmittedReason    = "TemplateOptionsMatchError"
 )
 
+// ValidWorkloadPaths Note: this needs to be updated anytime the spec changes
+var ValidWorkloadPaths = map[string]bool{
+	"workload.spec.source":                true,
```
cirocosta, line 47: IIRC, in our templates it's valid to have both paths prefixed with or without `.`, like

```
    urlPath: .status.artifact.url
             ^
```

and

```
    urlPath:  status.artifact.url
             ^
```

do you have any opinions on whether we should take that in consideration for these?

pkg/apis/v1alpha1/cluster_supply_chain.go
```diff
@@ -254,27 +144,58 @@ type SupplyChainResource struct {
...
-	Conditions         []metav1.Condition `json:"conditions,omitempty"`
-	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
+type Selector struct {
+	// MatchFields is a list of field selector requirements. The requirements are ANDed.
+	// +kubebuilder:validation:MinItems=1
+	MatchFields []FieldSelectorRequirement `json:"matchFields"`
+}
+
+type FieldSelectorRequirement struct {
+	// Key is the JSON path in the workload to match against.
```
cirocosta, line 182: Given the definition of `validWorkloadPath`, it seems like we could make more explicit that the possible set of """JSONPath""" (quotes given that it's more like a very small subset) constructs here is restricted to the spec 

```suggestion
	// Key is the JSON path in the Workload spec to match against.
```

pkg/selector/selector_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+var _ = Describe("Selector", func() {
+
+	Context("when the key exists in the context", func() {
+		var context map[string]interface{}
+
+		BeforeEach(func() {
+			context = map[string]interface{}{
+				"hello": "world",
+				"bad":   nil,
```
cirocosta, line 33: maybe a little pedantic, but might be worth adding a case for empty string? e.g., `Exists` would pass for the following

```go
			context = map[string]interface{}{
				"hello": "",
				"bad":   nil,
```
```go
					req := v1alpha1.FieldSelectorRequirement{
						Key:      "hello",
						Operator: v1alpha1.FieldSelectorOpExists,
					}
					ret, err := selector.Matches(req, context)
					Expect(err).ToNot(HaveOccurred())
					Expect(ret).To(BeTrue())
```

tests/kuttl/supplychain/options-with-values/01-supply-chain.yaml
```diff
@@ -0,0 +1,39 @@
...
+      templateRef:
+        kind: ClusterSourceTemplate
+        options:
+        - name: git-template---options-with-values
+          selector:
+            matchFields:
+              - key: workload.spec.source.git.url
+                operator: NotIn
+                values:
+#                  - "https://github.com/spring-projects/spring-petclinic.git"
```
cirocosta: leftovers?

cirocosta (commented):
looking good! I left a couple comments, nothing major 😁  manually tested a couple conditions, pretty cool how through the validations are 👍

pkg/apis/v1alpha1/workload.go
emmjohnson: Hmmmm, no real opinions. The code would work today except for the validations. Open to suggestions.

pkg/apis/v1alpha1/cluster_supply_chain.go
emmjohnson: The term spec is overloaded. We handle more than just workload.spec. workload.metadata is also valid.

pkg/selector/selector_test.go
emmjohnson: Are you suggesting this case should fail?

pkg/apis/v1alpha1/cluster_supply_chain.go
emmjohnson: Oh shoot, forgot to add workload.metadata to prefix only list

cirocosta (approved):

pkg/apis/v1alpha1/cluster_supply_chain.go
squeedee: I dont know how to suggest a suggestion on a suggested edit, so: how about "Key is the JSON path into the Workload Resource to match against"

PR #618, opened 2022-02-14
RFC: Add matchParams selector

jwntrs:
Add matchParams selector
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc/0000-match-params/rfc/rfc-0000-match-params.md)

### DIFF
--- a/rfc/rfc-0000-match-params.md
+++ b/rfc/rfc-0000-match-params.md
@@ -1,0 +1,138 @@
+# Meta
+[meta]: #meta
+- Name: Match params in field selector
+- Start Date: 2022-02-14
+- Author(s): @jwntrs
+- Status: Draft
+- RFC Pull Request:
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) introduced the ability to match templates based on fields in the workload. However matching on params is not well defined. It is currently not possible to match on a field in the supply chain, and matching on fields in the workload is complex. You would need to define something as follows:
+
+```yaml
+  selector:
+    matchFields:
+      - key: workload.spec.params[name == promotion].value    #< ===== does this even work?
+        operator: In
+        values: ["gitops"]
+```
+
+Lets introduce a top level params key:
+
+```yaml
+  selector:
+    matchFields:
+      - key: params.promotion    #< ===== use something like this instead
+        operator: In
+        values: ["gitops"]
+```
+
+
+# Motivation
+[motivation]: #motivation
+
+We should be able to match against supply chain params in our field selectors.
+
+# What it is
+[what-it-is]: #what-it-is
+
+```yaml
+apiVersion: kontinue.io/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: responsible-ops
+spec:
+  selector:
+    app: web
+  params:
+  - name: promotion       #< ===== params can be set here
+    value: (gitops|regops)
+  resources:
+   ...
+    - name: promote
+      templateRef:
+        kind: ClusterTemplate
+        options:
+        - name: git-promotion
+          selector:
+            matchFields:
+              - key: params.promotion           #< ===== introduce top level `params` that relies on our param hierarchy
+                operator: In
+                values: ["gitops"]
+        - name: registry-promotion
+          selector:
+            matchFields:
+              - key: params.promotion
+                operator: In
+                values: ["regops"]
+
+---
+apiVersion: kontinue.io/v1alpha1
+kind: Workload
+metadata:
+  name: petclinic
+  labels:
+    app: web
+spec:
+  params:
+  - name: promotion           #< ===== params can also be set here
+    value: (gitops|regops)
+  source:
+    git:
+      url: https://github.com/spring-projects/spring-petclinic.git
+      ref:
+        branch: main
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+Compute the available params from the workload and supply chain based on the [param hierarchy](https://cartographer.sh/docs/v0.2.0/architecture/#parameter-hierarchy) and pass it to the field matcher as `params`.
+
+# Migration
+[migration]: #migration
+
+N/A
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Ultimately we want our template `selector` and top-level supply chain `selector` to work the same. However, if we introduce params into the matching context, then it would cause some divergence in these two cases. 
+
+The top level selector we only have access to params defined in the workload.
+
+The template selectors would have access to params from the workload as well as the supply chain.
+
+Neither selector would have access to params from the templates.
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+???
+
+# Prior Art
+[prior-art]: #prior-art
+
+???
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+???
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+```yaml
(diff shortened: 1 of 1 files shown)


rfc/rfc-0000-match-params.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Summary
+[summary]: #summary
+
+The [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) introduced the ability to match templates based on fields in the workload. However matching on params is not well defined. It is currently not possible to match on a field in the supply chain, and matching on fields in the workload is complex. You would need to define something as follows:
+
+```yaml
+  selector:
+    matchFields:
+      - key: workload.spec.params[name == promotion].value    #< ===== does this even work?
```
martyspiewak: I think it would look like this:
`workload.spec.params[?(@.name=="promotion")].value`

rfc/rfc-0000-match-params.md
```diff
@@ -0,0 +1,138 @@
...
+- Start Date: 2022-02-14
+- Author(s): @jwntrs
+- Status: Draft
+- RFC Pull Request:
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+The [template switching RFC](https://github.com/vmware-tanzu/cartographer/pull/75) introduced the ability to match templates based on fields in the workload. However matching on params is not well defined. It is currently not possible to match on a field in the supply chain, and matching on fields in the workload is complex. You would need to define something as follows:
```
martyspiewak, line 13: Is the idea here to be able to match on a field (param) in the supply chain that may or may not be overwritten by a workload?

If not, there would be no reason to match on a field in the supply chain, right? Because that field will always be the same.

rfc/rfc-0000-match-params.md
jwntrs: thanks updated!

rfc/rfc-0000-match-params.md
jwntrs: Yeah the idea is to match on the resulting param after we process the [param hierarchy](https://cartographer.sh/docs/v0.2.0/architecture/#parameter-hierarchy). 

This means matching on a supply chain `default` or `value` or the workload `value`.

In the cases where the supply chain sets a value (which cannot be overridden) it may still be useful to be able to switch templates on this static value. You might want to maintain a single supply chain, and use ytt to template out the value during installation. This will let you use first class template switching with very minimal ytt logic.

rfc/rfc-0000-match-params.md
```diff
@@ -0,0 +1,138 @@
...
+
+```yaml
+  selector:
+    matchFields:
+      - key: workload.spec.params[?(@.name=="promotion")].value    #< ===== not so nice
```
scothis, line 18: I asked on [RFC 9 if the key value was JSON Path](https://github.com/vmware-tanzu/cartographer/pull/75/files#r798092768) and never got an answer.

scothis (commented):
Since `matchFields` is generic and able to match any field within the resource, I'd rather not overload that syntax by creating shortcuts for params. Even moreso if we are looking to change the target of the match from the workload/deliverable definition to values that are defined other places like on the ClusterSupplyChain/ClusterDeliverable or at runtime.

If we think that params are important enough to match on, they deserve top level support in a `matchParams` facet of the selector.

jwntrs:
> If we think that params are important enough to match on, they deserve top level support in a matchParams facet of the selector.

@scothis yeah @emmjohnson suggested something similar in our community meeting. I'm happy to take that direction with this RFC. @sclevine I think you mentioned in the meeting that you preferred the RFC as written, any objections to making the above change?

sclevine:
> Since matchFields is generic and able to match any field within the resource, I'd rather not overload that syntax by creating shortcuts for params.

@scothis We already have this shortcut in other places, and `matchFields` already uses non-standard notation for the field selector path (e.g., `workload.spec` instead `spec`). I suspect uses will assume that `params` works with `matchFields`, just as it works in other places. But no strong opinion.

squeedee:
> Since matchFields is generic and able to match any field within the resource, I'd rather not overload that syntax by creating shortcuts for params.

Despite originally thinking that MatchParams was overkill, I think you're right. MatchFields has a clear context, and Params are the resolution of partial-tree. I think drawing a user's attention to this special case with this separate selector is more educational.

squeedee (approved):
I think this rfc is important in it's ability to provide `defaultable fields`. 

The secondary, "supply chain configuration at install time" might actually be an anti-pattern? Creating a hard coded but 'selected' path will present a strange UX for visualiser, whereas templating at install with wholesale replacement of a resoruce makes for clearer supply chains.

scothis:
>I think this rfc is important in it's ability to provide defaultable fields.

That's a compelling distinction as it is not possible with a vanilla field selector operating against the Workload structure directly. It requires a partial templating context to match against.

I'm still on the fence as to whether `matchFields` would be better actually as a vanilla field selector operating against the Workload structure directly and the new behavior be defined within a `matchParam` facet.

rfc/rfc-0000-match-params.md
jwntrs: sorry that comment got missed, it is indeed a jsonpath

waciumawanjohi:
This strikes me as a bit of syntactic sugar. Params can be set in 4 places: 
- In the template
- In the resource field of a supply chain
- At the top level of a supply chain
- In the workload

When specifying a MatchParam, the supply-chain author knows 3 of these values already. Ultimately they are just checking, has the supply chain overwritten the choices made at the other 3 levels. So they need only check if the param exists on the workload and if it has a given value.

I have no objections to the change, but I wonder at its utility.

martyspiewak:
I think I agree with @emmjohnson and @scothis that if we're going to inject the param hierarchy logic in this field, it deserves its own `matchParams` key to be explicit.

But as @waciumawanjohi said, I struggle to understand the practical use case for this.

rfc/rfc-0000-match-params.md
```diff
@@ -0,0 +1,138 @@
...
+apiVersion: kontinue.io/v1alpha1
+kind: Workload
+metadata:
+  name: petclinic
+  labels:
+    app: web
+spec:
+  params:
+  - name: promotion           #< ===== params can also be set here
+    value: (gitops|regops)
```
waciumawanjohi, line 88: In this example, the workload param would be ignored, as the value has been hard coded by the supply-chain author. Switch `value` to `default` to clarify that the workload param is allowed.

```suggestion
    default: (gitops|regops)
```

jwntrs:
> But as @waciumawanjohi said, I struggle to understand the practical use case for this.

jwntrs:
> But as @waciumawanjohi said, I struggle to understand the practical use case for this.

@martyspiewak @waciumawanjohi as @squeedee pointed out this is about providing a path for defaultable fields. I've updated the yaml in the RFC to better illustrate this use case.

sclevine (approved):

emmjohnson (approved):

martyspiewak (approved):

squeedee (changes requested):
> When specifying a MatchParam, the supply-chain author knows 3 of these values already. Ultimately they are just checking, has the supply chain overwritten the choices made at the other 3 levels. So they need only check if the param exists on the workload and if it has a given value.
> 
> I have no objections to the change, but I wonder at its utility.

I agree with @waciumawanjohi and @martyspiewak  here, this is well reasoned. Adding this feature might reduce a "change in two places" complexity, however as it just adds complexity to the code and we've not had any user reports for this, I vote to move this to `declined`, with a note that it should be reopened should strong user need arise.

squeedee (approved):

waciumawanjohi (approved):

Issue #587, opened 2022-02-04
ClusterRunTemplate stamping pod or config map will continuously create new ones


martyspiewak:
### Bug description:
If you have a cluster run template that stamps out a pod or a config map, when you apply a runnable it will continuously stamp out new pods or config maps and garbage collection will not clean anything up.

Outputs are also never added to the runnable status

This seems to be because:
1. we always get a cache miss because we compare specs between the cache and the apiServer and there is no spec on a config map, and the pod's spec gets mutated (at least with the `nodeName`, but possibly with more). Therefore we will always get a cache miss and create a new object.
2. garbage collection is looking for `status.conditions[?(@.type=="Succeeded")].status` ([here](https://github.com/vmware-tanzu/cartographer/blob/29fcadaa57dc7deb25a878abc4a4ff707374bf12/pkg/realizer/runnable/gc/gc.go#L42)) which does not exist on either object. Therefore, it will not clean anything up.

### Steps to reproduce:
Please provide a set of steps to reproduce the behavior (example):
1. Apply attached run template and runnable:
```
apiVersion: carto.run/v1alpha1
kind: ClusterRunTemplate
metadata:
  name: test-template
spec:
  outputs:
    url: .status.hostIP
  template:
    apiVersion: v1
    kind: Pod
    metadata:
      generateName: test-
    spec:
      containers:
      - image: nginx:latest
        imagePullPolicy: Always
        name: webserver
```
```
apiVersion: carto.run/v1alpha1
kind: Runnable
metadata:
  name: test
spec:
  runTemplateRef:
    name: test-template
```
2. Observe the pods created in the namespace

### Expected behavior:
One pod should be created and there should be outputs on the runnable.

### Actual behavior:
Pods are continuously created and not cleaned up and there are no outputs.

### Versions:
 - Infrastructure: kind
 - cartographer version: running off main (currently [this commit](https://github.com/vmware-tanzu/cartographer/commit/ddf2477d50c40d8c7a8b39a1a48ac2c7f6616fc8))
 - k8s version: 1.21

zrob:
Had some discussions with the team on this one. Feels like this is intended behavior. We should document the requirement for the `status.conditions[?(@.type=="Succeeded")].status` and also document that we know configmaps and pods are examples of resources that won't work.

If we get followup demand for any resources that would run into this, let's tackle it as a feature to enable.

waciumawanjohi:
[...]

PR #621, opened 2022-02-14
Run template validate spec present

waciumawanjohi:

Should RunTemplates accept objects that do not have a spec? If an object does not have a `spec` field, there will be a cache miss when Carto examines the object. This will cause a new object to be stamped out.
Consider #587 

## Release Note

## PR Checklist


cirocosta:
> _If an object does not have a spec field, there will be a cache miss when Carto examines the object. This will cause a new object to be stamped out._

I wonder if that means we should try to address this at the cacher rather than individually at the Runnable level (where we can perform inspections like this because there's no `ytt` support) 🤔 

I'm not very familiar with the internals of the caching layer - in the case of a supplychain that stamps out a ConfigMap, what's the current impact of those cache misses?

waciumawanjohi:
>  in the case of a supplychain that stamps out a ConfigMap, what's the current impact of those cache misses?

For a supply chain template, if it deploys a resource that doesn't have a spec, the template will always update the object. Observable behavior would be generation increasing with every reconcile loop.

To fix this at the caching layer (which I think it desirable), we need to have some fields that we expect to be set by cartographer and other fields that we expect to be set by the object's reconciler. (for example: `spec` is Carto's concern and `status` is the other controller's concern). Right now our line in the sand is: Carto is concerned with spec and all other fields are the concern of the other controller. Looking at configmaps, `data` would be a reasonable addition to Carto's concern. There's an exploration that would need to be done to validate the fields that can be assumed to be Carto's concern.

squeedee (approved):

PR #634, opened 2022-02-16
RFC Resource Tracing

scothis:
[Readable](https://github.com/scothis/cartographer/blob/rfc-supply-chain-tracing/rfc/rfc-0000-resource-tracing.md)

closes # 

## Release Note

## PR Checklist

### DIFF
--- a/rfc/rfc-0000-supply-chain-tracing.md
+++ b/rfc/rfc-0000-supply-chain-tracing.md
@@ -65,7 +65,7 @@ status:
     - name:                  # string
     outputs:
     - name:                  # string
-      value:                 # json.RawExtension
+      value:                 # apiextensionsv1.JSON
       lastTransitionTime:    # metav1.Time
 ```
 
--- a/rfc/rfc-0000-supply-chain-tracing.md
+++ b/rfc/rfc-0000-supply-chain-tracing.md
@@ -1,0 +1,182 @@
+# Meta
+[meta]: #meta
+- Name: Supply Chain Tracing
+- Start Date: 2020-02-15
+- Author(s): @scothis
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+Note: Unless stated otherwise, the concepts in this RFC attributed to `Workload`s and `ClusterSupplyChain`s apply equally to `Deliverable`s and `ClusterDelivery`.
+
+# Summary
+[summary]: #summary
+
+Cartographer users create `Workload` and `Deliverable` resources that define work to be processed by `ClusterSupplyChain` and `ClusterDelivery` resources respectively. The work preformed by Cartographer on behalf of the user is largely opaque, and difficult to debug when something goes wrong. This RFC defines a way to trace the activity performed by Cartographer on behalf of the user's `Workload`/`Deliverable` by recording Cartographer's observed state of the world on the status of the resource that triggered the work.
+
+While traceability is an important aspect of artifact provenance, the goal for this RFC is to increase visibility of Cartographers current behavior. This work may serve as a foundation to later attestations of provenance, but is itself not.
+
+# Motivation
+[motivation]: #motivation
+
+For every `Workload` and `Deliverable` we should be able to answer:
+
+- What resources were stamped?
+- Which template was used to stamp out resources?
+- What outputs did each stamped resource produce?
+
+These questions are answerable today so long as you have permission on the cluster to view all the necessary resources, but doing so requires a deep understanding of Cartographer and the templates that define the resources to be stamped out. Instead we can distill the current state on the status of the resource in a form that can be used by client tools and users.
+
+There are a number of use cases for this information, internal and external:
+
+- tooling can visualize the resource graph of a given workload.
+- Cartographer can be more aware of previous resources to aid garbage collection of orphaned resources.
+
+# What it is
+[what-it-is]: #what-it-is
+
+After Cartographer matches a `ClusterSupplyChain` to a `Workload`, the resources defined by the supply chain are resolved to a template and template is applied to stamp out a kubernetes resource on the cluster. After the resource is created, outputs located at paths also defined by the template are read and propagated forward to the next resource in the chain. As Cartographer applies this behavior, it can keep a trace for each resource containing:
+
+1. the name in the ClusterSupplyChain
+2. a reference to the template
+3. a reference to the stamped resource
+4. named input resources as defined by the `ClusterSupplyChain`
+5. the name and value of each output resolved from the stamped resource
+
+The status for a Workload or Deliverable should be enhanced to add:
+
+```yaml
+... existing fields ...
+status:
+  ... existing fields ...
+  resources:
+  - name:                  # string
+    templateRef:           # corev1.ObjectReference
+      apiVersion:            # string
+      kind:                  # string
+      namespace:             # string (empty for cluster scoped resources)
+      name:                  # string
+    stampedRef:            # corev1.ObjectReference
+      apiVersion:            # string
+      kind:                  # string
+      namespace:             # string (empty for cluster scoped resources)
+      name:                  # string
+    inputs:
+    - name:                  # string
+    outputs:
+    - name:                  # string
+      value:                 # json.RawExtension
+      lastTransitionTime:    # metav1.Time
+    observedGeneration:    # int64
+```
+
+Based on the [basic supply chain example](https://github.com/vmware-tanzu/cartographer/tree/53402edb8b8914b4cd36ace82da85f83f3daefc1/examples/basic-sc), the Workload status would have the form:
+
+```yaml
+... existing fields ...
+status:
+  ... existing fields ...
+  resources:
+  - name: source-provider
+    stampedRef:
+      apiVersion: source.toolkit.fluxcd.io/v1beta1
+      kind: GitRepository
+      namespace: default
+      name: my-workload
+    templateRef:
+      apiVersion: carto.run/v1
+      kind: ClusterSourceTemplate
+      name: source
+      outputs:
+      - name: url
+        value: http://source-controller.flux-system.svc.cluster.local./gitrepository/default/my-workload/3d42c19a618bb8fc13f72178b8b5e214a2f989c4.tar.gz
+        lastTransitionTime: "2022-02-16T03:29:52Z"
+      - name: revision
+        value: main/3d42c19a618bb8fc13f72178b8b5e214a2f989c4
+        lastTransitionTime: "2022-02-16T03:29:52Z"
+      observedGeneration: 1
+  - name: image-builder
+    stampedRef:
+      apiVersion: kpack.io/v1alpha2
+      kind: Image
+      namespace: default
+      name: my-workload
+    templateRef:
+      apiVersion: carto.run/v1
+      kind: ClusterImageTemplate
+      name: image
+    inputs:
+    - name: source-provider
+    outputs:
+    - name: image
+      value: registry.example/supply-chain/my-workload@sha256:68f8e8fc6e8ede7a411db9182cd695eac7b3e7e19e4ff9dcb9ba21205c135697
+      lastTransitionTime: "2022-02-16T03:23:37Z"
+      observedGeneration: 14
+  - name: deployer
+    stampedRef:
+      apiVersion: kappctrl.k14s.io/v1alpha1
+      kind: App
+      namespace: default
+      name: my-workload
+    templateRef:
+      apiVersion: carto.run/v1
+      kind: ClusterTemplate
+      name: app-deploy
+    inputs:
+    - name: image-builder
+    observedGeneration: 19
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+- `.status.resources[*]` should match the order and length of `.spec.resource` for a ClusterSupplyChain.
+- `.status.resources[*].name` is the name of the resources as defined within the ClusterSupplyChain resource's array.
+- `.status.resources[*].templateRef` object reference to the template resource referenced by the ClusterSupplyChain.
+- `.status.resources[*].stampedRef` object reference to the Kubernetes resource created by Cartographer for this resource.
+- `.status.resources[*].inputs[*]` inputs model the relationships between resources within the SupplyChain graph. The value of an input can be read from the referenced resource's outputs.
+- `.status.resources[*].inputs[*].name` the name of the resource backing this input. 
+- `.status.resources[*].outputs[*].name` the name of the output. Output names are fixed and defined by the template type.
+- `.status.resources[*].outputs[*].value` raw value of the output. This may be large for config templates.
+- `.status.resources[*].outputs[*].lastTransitionTime` time of the most recent observed change of value. this is not the last time the value observed. This value is helpful to know if a supply chain is progressing. As this field is the result of observation of an edge triggered change, it should not be relied upon when accuracy matters. This is the same behavior as the lastTransitionTime field within a Condition.
+- `.status.resources[*].observedGeneration` the `.metadata.generation` of the stamped resource the outputs are read from. This value is not read from the `.status.observeredGeneration` field that may exist on the stamped resource. A future RFC may choose to take the stamped resource's observed generation under advisement, as that is a change to how Cartographer processes resources, it is inappropriate for this RFC.
+
+In the future, additional information about particular resources may be added. Like an indication of health with an error message when unhealthy. Support is not included in this RFC as it will require additional support within the template to know how to interpret the health of a resource and should be defined in a separate RFC.
+
+# Migration
+[migration]: #migration
+
+All of the data collected by this RFC exists within Cartographer today but is not exposed outwardly by the `Workload`. There are no changes to the runtime behavior required. All changes to existing CRDs are net new and will not break existing clients.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Some users may consider the runtime behavior of a `ClusterSupplyChain` to be an implementation detail that a user creating a `Workload` should not have visibility into. That a particular resource is created by Cartographer for a `Workload` may be viewed as an information disclosure. This change shines light on the internal behavior of a supply chain. The exact definition for each template is not exposed, unless the user also has access to view that resource. The outputs produced for a resource are the most sensitive values captured, but should never contain any confidential data.
+
+If there is significant user concern, a toggle could be developed to partially or fully disable tracing. Such a toggle is not in scope for this RFC.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternative to adding trace information to the Workload status is to continue not adding that information and require clients to reconstruct the values on their own. Other RFCs listed below under prior art have broader scopes that treat tracing as a side effect of deeper semantic changes to how Cartographer processes Workloads.
+
+# Prior Art
+[prior-art]: #prior-art
+
+Several RFC have attempted to make stronger pushes into this space, including:
+
+- [First draft of RFC 0014](https://github.com/vmware-tanzu/cartographer/pull/274)
+- [Introduce RFC 18 Workload Report Artifact Provenance](https://github.com/vmware-tanzu/cartographer/pull/519)
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Are the proposed field names well aligned with existing Cartographer-isms?
+- Are resource level observedGeneration and output level lastTransitionTime values helpful and meaningful?
+- Is all of the information collected and exposed appropriate to expose to all users who can view the workload resource?
+- Is there value in capturing the uuid for referenced resources?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+
+For both the `Workload` and `Deliverable` resources their status is adding a `resources` array, see the [What it is](#what-it-is) section for details.
--- a/rfc/rfc-0000-supply-chain-tracing.md
+++ b/rfc/rfc-0000-supply-chain-tracing.md
@@ -67,7 +67,6 @@ status:
     - name:                  # string
       value:                 # json.RawExtension
       lastTransitionTime:    # metav1.Time
-    observedGeneration:    # int64
 ```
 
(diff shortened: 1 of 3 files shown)


martyspiewak:
I think this is great!

In terms of `observedGeneration`, it feels a bit odd for it to refer to the generation of the stamped object, but not be nested under the `stampedRef`. I also wonder what the intended use of this field is? You mentioned that `lastTransitionTime` could be useful for seeing that the supply chain is progressing but I'm not sure I see what `observedGeneration` is useful for.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,182 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+- `.status.resources[*]` should match the order and length of `.spec.resource` for a ClusterSupplyChain.
+- `.status.resources[*].name` is the name of the resources as defined within the ClusterSupplyChain resource's array.
+- `.status.resources[*].templateRef` object reference to the template resource referenced by the ClusterSupplyChain.
+- `.status.resources[*].stampedRef` object reference to the Kubernetes resource created by Cartographer for this resource.
+- `.status.resources[*].inputs[*]` inputs model the relationships between resources within the SupplyChain graph. The value of an input can be read from the referenced resource's outputs.
+- `.status.resources[*].inputs[*].name` the name of the resource backing this input. 
+- `.status.resources[*].outputs[*].name` the name of the output. Output names are fixed and defined by the template type.
```
cirocosta: made me wonder if it'd be more clear if we instead named the field `resource:` (for inputs) 


~and maybe `type: ` (for output),  so that we're very explicit about it 🤔~ (nvm, for outputs)

cirocosta: ~(or perhaps even `resourceName` to really leave no room for other interpretations?)~

cirocosta: ~or ... what if we even removed the plural here (output**s**) and went solely with `output`? a resource can only export either one (`cluster(source|image|config)template`) or no output (`clustertemplate`)), and the type can already be inferred by looking up the templateref (`.status.resources[*].templateRef.kind`)~

nvm, each entry is a field of the output

martyspiewak:
@emmjohnson and I took a stab at spiking this out to see what it might look like. It's on [this branch](https://github.com/vmware-tanzu/cartographer/tree/spike-resources-in-workload-status), if anyone is interested in taking a look.

scothis:
> In terms of `observedGeneration`, it feels a bit odd for it to refer to the generation of the stamped object, but not be nested under the `stampedRef`. I also wonder what the intended use of this field is? You mentioned that `lastTransitionTime` could be useful for seeing that the supply chain is progressing but I'm not sure I see what `observedGeneration` is useful for.

All fair comments. I was trying to preserve a bit of the behavior from RFC 18 that included the ResourceVersion. I don't have a specific use case in mind. If we don't collectively have a use, I'll remove it.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,182 @@
...
+- `.status.resources[*].outputs[*].lastTransitionTime` time of the most recent observed change of value. this is not the last time the value observed. This value is helpful to know if a supply chain is progressing. As this field is the result of observation of an edge triggered change, it should not be relied upon when accuracy matters. This is the same behavior as the lastTransitionTime field within a Condition.
+- `.status.resources[*].observedGeneration` the `.metadata.generation` of the stamped resource the outputs are read from. This value is not read from the `.status.observeredGeneration` field that may exist on the stamped resource. A future RFC may choose to take the stamped resource's observed generation under advisement, as that is a change to how Cartographer processes resources, it is inappropriate for this RFC.
+
+In the future, additional information about particular resources may be added. Like an indication of health with an error message when unhealthy. Support is not included in this RFC as it will require additional support within the template to know how to interpret the health of a resource and should be defined in a separate RFC.
```
cirocosta: +1 leaving it possible for extension later on

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,182 @@
...
+- `.status.resources[*].inputs[*].name` the name of the resource backing this input. 
+- `.status.resources[*].outputs[*].name` the name of the output. Output names are fixed and defined by the template type.
+- `.status.resources[*].outputs[*].value` raw value of the output. This may be large for config templates.
+- `.status.resources[*].outputs[*].lastTransitionTime` time of the most recent observed change of value. this is not the last time the value observed. This value is helpful to know if a supply chain is progressing. As this field is the result of observation of an edge triggered change, it should not be relied upon when accuracy matters. This is the same behavior as the lastTransitionTime field within a Condition.
+- `.status.resources[*].observedGeneration` the `.metadata.generation` of the stamped resource the outputs are read from. This value is not read from the `.status.observeredGeneration` field that may exist on the stamped resource. A future RFC may choose to take the stamped resource's observed generation under advisement, as that is a change to how Cartographer processes resources, it is inappropriate for this RFC.
```
cirocosta: pretty interesting to have _both_ observedGeneration and lastTransitionTime - e.g., `kpack/Image` having a new build due to the builder being updated (thus, no change in generation _but_ new latestImage produced, thus update in `lastTransitionTime` despite no update in `observedGeneration`)

+1

waciumawanjohi:
I can imagine uses of the lastTransitionTime, think it's a great addition.

I don't think the observedGeneration is actionable. Even more, I think it would be often misinterpreted. I would expect users to infer far too much, e.g. that when the outputs report observedGeneration X then they are the result of the inputs of generation X (which is not necessarily true).

In practice, the observedGeneration on its own just lets me know that the object's controller is 'on'. Even then, without exposing the `.metadata.generation` the observedGeneration is simply a monotonic counter. Vote to remove the field.

waciumawanjohi:
To point out a possible risk:

We have another RFC on tracing that would duplicate much of this information if adopted. The use cases for that RFC remain and I haven't heard progress on alternate work that would need to be done at the level of all of the choreographed components of a supply chain if that RFC were to be rejected. E.g. at the moment I would predict that we're going to move forward with some version of that RFC.

An important note about tracing is that it is artifact centric rather than resource centric. E.g. wherein this proposal reports 1 artifact per resource, there is good reason for the tracing RFC to report more than 1 artifact for each resource.

If we adopt this RFC and later adopt the tracing RFC we'll need to decide:
- if we want to have two longish fields on the workload that have much of the same information, or
- if we want to make a breaking change and alter what's proposed here.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,182 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+Several RFC have attempted to make stronger pushes into this space, including:
+
+- [First draft of RFC 0014](https://github.com/vmware-tanzu/cartographer/pull/274)
+- [Introduce RFC 18 Workload Report Artifact Provenance](https://github.com/vmware-tanzu/cartographer/pull/519)
+
+# Unresolved Questions
```
waciumawanjohi: To point out a possible risk:

We have another RFC on tracing that would duplicate much of this information if adopted. The use cases for that RFC remain and I haven't heard progress on alternate work that would need to be done at the level of all of the choreographed components of a supply chain if that RFC were to be rejected. E.g. at the moment I would predict that we're going to move forward with some version of that RFC.

An important note about tracing is that it is artifact centric rather than resource centric. E.g. wherein this proposal reports 1 artifact per resource, there is good reason for the tracing RFC to report more than 1 artifact for each resource.

If we adopt this RFC and later adopt the tracing RFC we'll need to decide:
- if we want to have two longish fields on the workload that have much of the same information, or
- if we want to make a breaking change and alter what's proposed here.

rfc/rfc-0000-supply-chain-tracing.md
scothis: or change `inputs` to `inputResources`?

I waffled a bit while drafting the RFCs whether `.inputs[*].name` should be `.inputs[*].resource` since the value is from the `resource` field within the ResourceReference struct. I ended up going with `name` here because it's the dominate part of the relationship and I was viewing this as a local object reference like resource.

waciumawanjohi:
Random request:
Can the title of this RFC change? We've been using the term tracing to discuss following artifact A as it propagates through a supply chain and creates downstream changes. This RFC does not attempt to achieve that.

Possible names:
- Resource reporting
- Exposing stamped resources to workload owners
- ...

rfc/rfc-0000-supply-chain-tracing.md
scothis: The generation is always going to be spec-change centric. Outputs via the stamped resource's status can change at any time. I thought it would be interesting to capture the timestamp when the value was first observed. The timestamp is helpful to users as if the timestamp is older than you would have expected a value to be emitted, then that's a strong signal that something is wrong, even if the (future) health says it's working.

rfc/rfc-0000-supply-chain-tracing.md
scothis: Please add a link to the RFC and ideally the parts of the RFC which you feel are relevant

rfc/rfc-0000-supply-chain-tracing.md
scothis: >a resource can only export either one (cluster(source|image|config)template) or no output (clustertemplate))

We certainly could make the output a mux. I didn't because I think it makes it harder for a generic tool to consume as it would need to use a switch to find the active bit of the config. This is a loosely held opinion.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,176 @@
...
+    stampedRef:            # corev1.ObjectReference
+      apiVersion:            # string
+      kind:                  # string
+      namespace:             # string (empty for cluster scoped resources)
+      name:                  # string
+    inputs:
+    - name:                  # string
+    outputs:
+    - name:                  # string
+      value:                 # json.RawExtension
```
martyspiewak: nit:
```suggestion
      value:                 # apiextensionsv1.JSON
```

jwntrs (approved):

martyspiewak (approved):

scothis:
>Can the title of this RFC change? 

sure

>We've been using the term tracing to discuss following artifact A as it propagates through a supply chain and creates downstream changes. This RFC does not attempt to achieve that.

I consider both of these to be tracing, but there's a difference in what is being traced. In this case what we're tracing is the supply chain defined steps for a workload. I agree that it's not artifact level tracing.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,176 @@
...
+      name: my-workload
+    templateRef:
+      apiVersion: carto.run/v1
+      kind: ClusterImageTemplate
+      name: image
+    inputs:
+    - name: source-provider
+    outputs:
+    - name: image
+      value: registry.example/supply-chain/my-workload@sha256:68f8e8fc6e8ede7a411db9182cd695eac7b3e7e19e4ff9dcb9ba21205c135697
```
scothis: On RFC 18 @evankanderson raised a valid concern about the potential size of outputs causing issues.

https://github.com/vmware-tanzu/cartographer/pull/519#discussion_r810103864

>There is also a 1MB limit on the size of a single resource, and the values might be repeated twice due to the addition of managedFields. I'd be careful about including large other documents directly for that reason. It seems like it would be better to include a resourceVersion or generation field referencing the resource's version, rather than the entire contents.

scothis: We could switch including the value to use a hash of the value. Clients would be required to fetch the stamped resource in order to resolve the actual value.

squeedee (approved):

rfc/rfc-0000-supply-chain-tracing.md
scothis: Switched to a `digest` and `path` instead of a `value` for outputs.

rfc/rfc-0000-supply-chain-tracing.md
```diff
@@ -0,0 +1,181 @@
...
+[how-it-works]: #how-it-works
+
+- `.status.resources[*]` should match the order and length of `.spec.resource` for a ClusterSupplyChain.
+- `.status.resources[*].name` is the name of the resources as defined within the ClusterSupplyChain resource's array.
+- `.status.resources[*].templateRef` object reference to the template resource referenced by the ClusterSupplyChain.
+- `.status.resources[*].stampedRef` object reference to the Kubernetes resource created by Cartographer for this resource.
+- `.status.resources[*].inputs[*]` inputs model the relationships between resources within the SupplyChain graph. The value of an input can be read from the stamped resource based on the outputs for the referenced resource.
+- `.status.resources[*].inputs[*].name` the name of the resource backing this input. 
+- `.status.resources[*].outputs[*].name` the name of the output. Output names are fixed and defined by the template type.
+- `.status.resources[*].outputs[*].digest` sha256sum of the raw output value.
```
cirocosta: I wonder if we need to be more specific here given how "sensitive" hashing is to any bit change, e.g., a javascript based client might end up interpreting the raw value differently from the controller due to a different json decoder? wondering that because we're effectively hashing the result of the jsonpath query (rather than straight byte stream)

cirocosta: I know you mentioned that we should try as much as possible to not expose potentially sensitive information in `workload.status`, but storage-wise, made me wonder if it'd be too much of a crazy/bad idea considering a custom apiserver implementation to work around the etcd storage limitation

IIRC, because you can back the endpoints you register against with ... anything, we could back it up with """something else""", although it's clearly a large complexity add

somewhat related: https://github.com/tektoncd/community/pull/606

rfc/rfc-0000-supply-chain-tracing.md
jwntrs: I definitely understand the argument about the config being too big, but I think capturing the value here is important. A lot of our modelling assumes that the workload is the only object that a developer needs to care about, now we're saying that they need to go and make these other associations to other objects.

rfc/rfc-0000-supply-chain-tracing.md
scothis: >I wonder if we need to be more specific here given how "sensitive" hashing is to any bit change, e.g., a javascript based client might end up interpreting the raw value differently from the controller due to a different json decoder?

That's a good point, the digest will only be stable against a string or byte array value, structs are ambiguous. The intent isn't to be cryptographically secure, but to provide a check of the value so a client can be reasonably sure that it has the "correct" value and that the value hasn't changed between when the c

We can drop the `digest` field for now if you're not confident in it, and add it back once we have the semantics nailed.

> made me wonder if it'd be too much of a crazy/bad idea considering a custom apiserver implementation to work around the etcd storage limitation

yes, crazy and bad :)

But on a serious note, there are other concerns that would be nice to have that will never end up on the Workload status. For example, a history of artifacts and how they interact in the supply chain. This kind of data should be captured into a different system that can support temporal (and other types of) queries. I wouldn't use that front an aggregated api, but to be used as a separate store.

rfc/rfc-0000-supply-chain-tracing.md
martyspiewak: I think if we don't have `value` anymore then we'd have to remove `lastTransitionTime` as we won't be storing the old value so we won't know if it's changed. It'd also be misleading as to what `lastTransitionTime`- it could be taken to mean the last time the output path changed.

jwntrs: We link to the template already which has the `path`. If we aren't going to include the output values maybe we just remove outputs altogether. But for the record I think this removes a lot of user value so i'm still very much in favour of keeping the output values.

martyspiewak: I think there is still some utility to including the output path so that developers don't have to go to the templates to find the output path (which may also be hard to decipher as its not necessarily clear that `urlPath` is the output path they're looking for).

I also think there is a lot of value lost by removing `lastTransitionTime`. Is there a way we could keep `value` but just truncate it if it's too long?

mklanjsek: Can we find a reasonable compromise and keep some kind of short summary in there, at least timestamp and status, so we don't have to access stamped resources to collect that info? 

That would be very helpful and definitely improve the performance when displaying the supply chain topology.

scothis: lastTransitionTime is still available. I assume by "a short summary" and `status` you mean something on the order of a metav1.Condition#Status and not the full `$.status` from the stamped resource. 

A simple `True` `False` `Unknown` is something we should drive towards, but to do so requires adding semantics that Cartographer is unaware of today and user facing template configuration, which makes it squarely out of scope for this RFC. 

&lt;sarcasim&gt;
If you want, you can assume that all outputs have a status field that just isn't set (meaning unknown) since that would accurately reflect what Cartographer knows today about the stamped resource.
&lt;/sarcasim&gt;

martyspiewak: > lastTransitionTime is still available

If cartographer isn't storing the values, i'm not sure how it would know if the value has changed.

scothis: > > lastTransitionTime is still available
> 
> If cartographer isn't storing the values, i'm not sure how it would know if the value has changed.

The `digest` represents the raw value. @cirocosta [raised concerns](https://github.com/vmware-tanzu/cartographer/pull/634#discussion_r812314936) about the stability of the digest if we expect others to use it to verify an output, but it's sufficiently stable for our own purposes.

martyspiewak: Ah I thought the discussion was heading towards removing `digest`, but if it's sticking around then ya we can definitely still have `lastTransitionTime`.

emmjohnson (approved):

scothis:
renamed to "Resource Tracing"

rfc/rfc-0000-supply-chain-tracing.md
waciumawanjohi: > On RFC 18 @evankanderson raised a valid concern about the potential size of outputs causing issues.

> We could switch including the value to use a hash of the value. Clients would be required to fetch the stamped resource in order to resolve the actual value.

I also thought we were exploring options to address this possible large status size and therefore remove the digest. I will say, it's not clear to me that we have hypothetical supply chains that would create a 1 MB status (or even 500K, assuming that artifact tracing ends up duplicating much of resource tracing). But if we assume that it is, and that we want to add a hash field and remove the digest, I also wonder about @mklanjsek's question:

> Can we find a reasonable compromise and keep some kind of short summary in there, at least timestamp and status, so we don't have to access stamped resources to collect that info?

Options that I can think of:
1. Include a truncated version of the digest, with a character length that is long enough for most urls, revisions, image digests. This would leave only configs as values that need to be retrieved by end users.
2. Rather than include a digest, construct and include the `kubectl get ... | yq . <path>` command that a user could run to get the value. (Of course, there are 2 popular yq tools, so this may cause more problems than it solves)

scothis: > we want to add a hash field and remove the digest

What do you see as the difference hash and digest? I consider a digest to be the output of a hashing function.

> Include a truncated version of the digest

the key point of a digest is that it's a fixed length (based on the hash algorithm)

> construct and include the `kubectl get ... | yq . <path>` command that a user could run to get the value

We provide the details sufficient to lookup the value with your own set of tools. Generally, I don't expect APIs to make assumptions about the tools that are used to access that API.

squeedee (approved):
I believe folks still expect something like the digest as a part of this but everything else is fine as is.

rfc/rfc-0000-supply-chain-tracing.md
squeedee: I'm not sure how sensible this is, but it is possible to serialize with sorted keys in the go yaml package (unsure about the k8s one). That would at least produce repeatable digests.

rfc/rfc-0000-supply-chain-tracing.md
jwntrs: @sclevine suggested this out-of-band:
> maybe we could add a preview field next to digest that includes, e.g., just the commit SHA, just the image digest, just the first N bytes of the config?

Thoughts on including something like this @scothis?

rfc/rfc-0000-supply-chain-tracing.md
scothis: We can also say that the digest is primarily for Cartographer's internal use (to generate the lastTransitionTime corectly).

rfc/rfc-0000-supply-chain-tracing.md
scothis: @jwntrs it re-raises the same question about the sensitivity of outputs. For a sufficiently small value of N, it should alleviate all practical concerns around overflowing the size limitation for a k8s resource.

cirocosta: @scothis wdyt of something similar to https://github.com/kubernetes/kubernetes/pull/96084 to address the sensitivity case?

scothis: @cirocosta how do we know what value needs to be masked?

scothis: I can update the output `digest`/`value`/`preview` to be whatever people collectively want. 

There are two different dimensions of risk/customer impact to consider with the decision.
1. [confidentiality] "leaking" outputs that an org views as sensitive
2. [availability] a supply chain stops working in production, with an unclear path to recovery

| field  | confidentiality risk | availability risk |
| ------------- | ------------- | -------- |
| `value`  | medium | high |
| `preview`  | medium  | low |
| `digest`  | low | low |

jwntrs: Can I get some thumbs up/thumbs down on moving forward with the following:

Add a `preview` field with which holds the first 200 bytes of each value.

sclevine: @scothis After today's chat, I'm pretty comfortable with adding the `preview` field with <1000 bytes (200 seems fine to me). If you aren't opposed, would you mind adding it?

scothis: added

cirocosta (approved):

sclevine (approved):

idoru (approved):

PR #661, opened 2022-02-24
Write Service object as a YAML fragment in example

pivotaljohn:

This PR illustrates a recommendation to use YAML templating rather than text templating in the "Source to App Config In Git" example.

By:
- capturing the Knative Service object as a [YAML Fragment](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-yaml-fragment), and
- [string-encoding that YAML](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-ytt/#yaml) into the ConfigMap `manifest` entry
the end-user can avoid challenges related to text templating (e.g. string escaping, indentation).

Whenever a YAML structure is being built, the Carvel team recommends using this technique.

Also (minor): remove redundant `load()` call in `serviceaccount.yaml`: the scope of identifiers set from a `load()` is the entire file, not just the document.

## PR Checklist

- [ ] ~Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`~

### DIFF
--- a/examples/gitwriter-sc/README.md
+++ b/examples/gitwriter-sc/README.md
@@ -195,41 +195,49 @@ spec:
   # context is avoided.
   ytt: |
     #@ load("@ytt:data", "data")
-    apiVersion: v1
-    kind: ConfigMap # <=== a configmap is templated
+    #@ load("@ytt:yaml", "yaml")
+
+    #! `service()` returns a YAML fragment that is the Service object
+    #@ def service():
+    apiVersion: serving.knative.dev/v1
+    kind: Service
     metadata:
-      name: #@ data.values.workload.metadata.name # <=== The same templating context is available
-                                                         with info about the workload, params and artifacts
-                                                         created earlier by the supply chain
-    data:
-      #@yaml/text-templated-strings
-      manifest: |
-        apiVersion: serving.knative.dev/v1
-        kind: Service # <=== in the configmap.data field, a Service object is configured
+      name: #@ data.values.workload.metadata.name
+      labels:
+        #! An optional label for kapp-controller is written if specified by the app dev
+        #@ if "labels" in data.values.workload.metadata and \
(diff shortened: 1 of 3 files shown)


pivotaljohn:
troubleshooting the failed build. 👍🏻

pivotaljohn:
🤔  I might need a hand with this.

With the original change-set (edits to the README.md and the example code itself), [the validation failed](https://github.com/vmware-tanzu/cartographer/runs/5312046096?check_suite_focus=true#step:10:1964) for reasons clearly connected to the change-set.
However, I'm unable to determine the root cause — I can't seem to find the clear signal in the CI output.

In an effort to "get back to green," I've since completely walked back any code change (all that's currently sitting in the PR, right now are edits to the README.md).
But that [failed validation for what seems to be an unrelated reason](https://github.com/vmware-tanzu/cartographer/runs/5326411317?check_suite_focus=true#step:9:1956).

Any advice of either something to look at or try?

pivotaljohn:
Thank you, @waciumawanjohi!

Tracking down the root cause of this error:
```
kapp: Error: Applying create service/gitwriter-sc (serving.knative.dev/v1) namespace: default:
      Creating resource service/gitwriter-sc (serving.knative.dev/v1) namespace: default:
        API server says: admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: cannot decode incoming new object: json: cannot unmarshal number into Go struct field ObjectMeta.spec.template.metadata.annotations of type string (reason: BadRequest)
```

Currently quite confused as both versions of the ConfigMap are setting that value of the annotation to a string... somewhere, the value `"1"` is being converted to `1`... 🤔

waciumawanjohi:
Closing this PR as the example currently follows the above advice (to use yaml templating by creating a yaml fragment and encoding it). See: https://github.com/vmware-tanzu/cartographer/blob/912b1521c29c6e64bcbbaeac4a44417375e85131/examples/gitwriter-sc/app-operator/config-service.yaml#L32

PR #675, opened 2022-03-01
Resources in workload status

### LINKED ISSUE #664: Include supply chain resources in workload status
From RFC https://github.com/vmware-tanzu/cartographer/pull/634

Spike on this branch https://github.com/vmware-tanzu/cartographer/tree/spike-resources-in-workload-status

martyspiewak:

Workloads/deliverables list the stamped objects, templates, inputs, and outputs created by the blueprint per #634 

closes #664 

## Release Note

Workloads/deliverables list the stamped objects, templates, inputs, and outputs created by the blueprint per #634 

## PR Checklist


pkg/apis/v1alpha1/common.go
```diff
@@ -191,3 +192,22 @@ type FieldSelectorRequirement struct {
 	// the values array must be empty.
 	Values []string `json:"values,omitempty"`
 }
+
+type RealizedResource struct {
```
cirocosta, line 196: I think it'd be nice to have a description for these , e.g., for the others it's pretty well defined

```console
$ k explain workload.status
KIND:     Workload
VERSION:  carto.run/v1alpha1
...
   observedGeneration   <integer>
     ObservedGeneration refers to the metadata.Generation of the spec that
     resulted in the current `status`.

   resources    <[]Object>
     Resources contain references to the objects created by the Supply Chain and
     the templates used to create them. It also contains Inputs and Outputs that
     were passed between the templates as the Supply Chain was processed.
```

but then lacking at this level

```console
$ k explain workload.status.resources
KIND:     Workload
VERSION:  carto.run/v1alpha1
...
   inputs       <[]Object>

   name <string> -required-

   outputs      <[]Object>
```

pkg/controller/deliverable/reconciler.go
```diff
@@ -154,41 +156,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
-func (r *Reconciler) completeReconciliation(ctx context.Context, deliverable *v1alpha1.Deliverable, err error) (ctrl.Result, error) {
+func (r *Reconciler) completeReconciliation(ctx context.Context, deliverable *v1alpha1.Deliverable, realizedResources []v1alpha1.RealizedResource, err error) (ctrl.Result, error) {
 	log := logr.FromContextOrDiscard(ctx)
 	var changed bool
 	deliverable.Status.Conditions, changed = r.conditionManager.Finalize()
 
 	var updateErr error
-	if changed || (deliverable.Status.ObservedGeneration != deliverable.Generation) {
+	if changed || (deliverable.Status.ObservedGeneration != deliverable.Generation) || !reflect.DeepEqual(deliverable.Status.Resources, realizedResources) {
```
cirocosta, line 197: nit: I usually see folks using [`equality.Semantic.DeepEqual`](https://github.com/kubernetes/apimachinery/blob/2936d3f03931b9c06a641799605d8e806cb2a58b/pkg/api/equality/semantic.go), but, tbh, I'm not sure how significant that is for what we're achieving here

pkg/templates/templates.go
```diff
@@ -24,10 +24,13 @@ import (
 	"github.com/vmware-tanzu/cartographer/pkg/eval"
 )
 
+const PREVIEW_CHARACTER_LIMIT = 1000
```
cirocosta: 1024 👀? 

_(very loosely held opinion 😅 )_

cirocosta:
I was trying out a small supplychain to check some possible edge cases, and noticed that with the current implementation (which is indeed right according to the RFC), we can observe that whenever there's an error that prevents the controller from gathering an output, further resources down the list end up with the `stampedRef` and `outputs` emptied out

e.g., consider the following supplychain:

```yaml
  resources:
    - name: source-provider
      templateRef:
        kind: ClusterSourceTemplate
        name: source

    - name: image-builder
      templateRef:
        kind: ClusterImageTemplate
        name: image
      sources:
        - resource: source-provider
          name: source

    - name: deployer
      templateRef:
        kind: ClusterTemplate
        name: app-deploy
      images:
        - resource: image-builder
          name: image
```

with everything going well, I end up with the following status:

```yaml
  resources:
  - name: source-provider
    outputs:
    - digest: sha256:e6acb13bf8ddf0bf710d1129769ed4b80f60e549c86b908beec37ddd0450dc9c
      lastTransitionTime: "2022-03-01T22:09:45Z"
      name: url
      preview: http://source-controller.gitops-toolkit.svc.cluster.local./gitrepository/default/test-basic/3d42c19a618bb8fc13f72178b8b5e214a2f989c4.tar.gz
    - digest: sha256:76004630fc3135b24df7d411182476d9324e091975f7ef15137fc4c25acd5056
      lastTransitionTime: "2022-03-01T22:09:45Z"
      name: revision
      preview: main/3d42c19a618bb8fc13f72178b8b5e214a2f989c4
    stampedRef:
      kind: GitRepository
      name: test-basic
      namespace: default
    templateRef:
      kind: ClusterSourceTemplate
      name: source
  - inputs:
    - name: source-provider
    name: image-builder
    outputs:
    - digest: sha256:07b7d18426145fd1d71b24fadda0c6bf3568903da9ecd62c85c330ce0316a455
      lastTransitionTime: "2022-03-01T22:10:22Z"
      name: image
      preview: 10.188.0.3:5000/test/test-basic-default@sha256:e28c24a630af33e967b3adb9a1ba3e03cc4180b88eace6ce15f0b15394d84907
    stampedRef:
      apiVersion: kpack.io/v1alpha2
      kind: Image
      name: test-basic
      namespace: default
    templateRef:
      kind: ClusterImageTemplate
      name: image
  - inputs:
    - name: image-builder
    name: deployer
    stampedRef:
      apiVersion: kappctrl.k14s.io/v1alpha1
      kind: App
      name: test-basic
      namespace: default
    templateRef:
      kind: ClusterTemplate
      name: app-deploy
```

but now, consider that the second resource in the list (`image-builder`) for some reason starts being unhappy about the PATCHes that Cartographer is making against it (e.g., maybe the webhook service is down, or .. who knows) - because of such failure, we'd then see that anything after it loses both `stampedRef` _and_ `outputs`:

```diff
--- a/before.yaml
+++ b/now.yaml
@@ -21,16 +21,6 @@
   - inputs:
     - name: source-provider
     name: image-builder
-    outputs:
-    - digest: sha256:07b7d18426145fd1d71b24fadda0c6bf3568903da9ecd62c85c330ce0316a455
-      lastTransitionTime: "2022-03-01T22:10:22Z"
-      name: image
-      preview: 10.188.0.3:5000/test/test-basic-default@sha256:e28c24a630af33e967b3adb9a1ba3e03cc4180b88eace6ce15f0b15394d84907
-    stampedRef:
-      apiVersion: kpack.io/v1alpha2
-      kind: Image
-      name: test-basic
-      namespace: default
     templateRef:
       apiVersion: carto.run/v1alpha1
       kind: ClusterImageTemplate
@@ -38,12 +28,10 @@
   - inputs:
     - name: image-builder
     name: deployer
-    stampedRef:
-      apiVersion: kappctrl.k14s.io/v1alpha1
-      kind: App
-      name: test-basic
-      namespace: default
     templateRef:
       apiVersion: carto.run/v1alpha1
       kind: ClusterTemplate
       name: app-deploy
```

it makes sense to have such behavior given that we're exposing under `.status` what the controller knows at the time as it goes through the DAG, but it's imaginable that one would ultimately want to still be able to tell what's the `image` that is being fed to `deployer` even though PATCHing `image-builder` is currently not working, and it's still probably interesting to know that the latest stampedRef for both `image-builder` and `deployer` (even though in the current reconciliation we're indeed not in a position to stamp anything).

I lean towards "yes, working as intended", but I can see that it might be surprising to some.

wdyt @scothis? 

thx!

jwntrs:
@cirocosta I lean towards "not working as intended". The resources RFC is meant to represent the current state of each resource, independent of other resources. A failure in one resource shouldn't invalidate the status of another. Hmm I think this is doubly true if we are going to use this list of resources as the source of truth for gc.

martyspiewak:
> Hmm I think this is doubly true if we are going to use this list of resources as the source of truth for gc

That's a really good point, we definitely can't rely on it in its current state for GC.

But I do think it could also be misleading if we don't zero out the downstream resources. For example, if resource 1 is an input to resource 2 and then resource 1 begins to fail, if we do not zero out resource 2 then it resource 2 will show resource 1 as an input, even though resource 1 is currently in a bad state and not reporting an output. So you'd end up with something like:
```
status:
  resources:
  - name: resource-1
    templateRef:
      name: some-template-ref
  - name: resource-2
    templateRef:
      name: some-other-template-ref
    stampedRef:
      name: some-stamped-ref
    inputs:
     - name: resource-1
```
This also seems problematic to show to users.

jwntrs:
>  if we do not zero out resource 2 then it resource 2 will show resource 1 as an input, even though resource 1 is currently in a bad state and not reporting an output

This is the big distinction between this resource centric approach vs artifact centric approach (RFC 18). All the resource graph is saying is that this resource is downstream from some other resource. It makes no guarantees about the inputs/outputs themselves. It is meant to represent the current known state of all the resources in the graph.

pkg/realizer/deliverable/realizer.go
```diff
@@ -105,7 +105,34 @@ func generateRealizedResource(resource v1alpha1.DeliveryResource, template templ
...
+			for _, previousResource := range previousResources {
+				if previousResource.Name == resource.Name {
+					outputs = previousResource.Outputs
+					break
+				}
+			}
+		} else {
+			currTime := metav1.NewTime(time.Now())
+			for j, out := range outputs {
+				outputs[j].LastTransitionTime = currTime
```
cirocosta, line 120: do you think we could split some of the following one/more functions? 😅   this code code path is getting tough to follow (same for `workload`)

pkg/apis/v1alpha1/common.go
emmjohnson: We did it. But I'm sold that anyone will understand what we are talking about.

pkg/templates/templates.go
emmjohnson: I'm sad because of kuttl, but ok

pkg/controller/deliverable/reconciler.go
emmjohnson: I'm not sure this is the correct DeepEqual to use? We are comparing two []v1alpha1.RealizedResource. That seems to be intended for full api objects? We use reflect.DeepEqual else where in the code to compare structs/[]structs.

cirocosta: I see, yeah makes sense 👍 👍  thank you!

tests/kuttl/delivery/deliverable-status/02-assert.yaml
```diff
@@ -0,0 +1,117 @@
...
+        apiVersion: v1
+        kind: ConfigMap
+        name: example-first-configmap
+      templateRef:
+        apiVersion: carto.run/v1alpha1
+        kind: ClusterSourceTemplate
+        name: example-first---deliverable-status
+      outputs:
+        - name: url
+          preview: '"three"'
```
cirocosta, line 71: I wasn't really sure about the `"` around the string (as it must being `json`), so I gave a little try using `yaml` instead, and while it's indeed "nicer" for strings, for anything that's an object, `json` indeed looks much better imo

e.g.:

```
  - inputs:
    - name: image-builder
    name: deployer
    outputs:
    - digest: sha256:d6921e73a5877a89cd5c9e4c152a4c23f65bdda57507fd3a6881e7c9a78e6743
      lastTransitionTime: "2022-03-02T21:27:01Z"
      name: config
      preview: '{"deploy":[{"kapp":{}}],"fetch":[{"inline":{"paths":{"kapp-config.yml":"
```

vs

```
  - inputs:
    - name: image-builder
    name: deployer
    outputs:
    - digest: sha256:a97722bdd4787940f98dacd2434dcfc624ae8924bc6ffc960146d2e392f1650b
      lastTransitionTime: "2022-03-02T21:25:37Z"
      name: config
      preview: "deploy:\n- kapp: {}\nfetch:\n- inline:\n    paths:\n      kapp-config.yml:
```

cirocosta (approved):
lgtm! tested with the same case of breaking a resource that's dependent on by a further resource and it preserves the fields 😁  

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-03-02T21:16:59Z"
    message: ""
    reason: Ready
    status: "True"
    type: SupplyChainReady
  - lastTransitionTime: "2022-03-02T21:28:39Z"
    message: |-
      unable to apply object [default/test-basic] for resource [image-builder] in supply chain [supply-chain]: patch: admission webhook "validation.webhook.kpack.io" denied
the request: validation failed: Immutable field changed: spec.tag
      got: foo, want: 10.188.0.3:5000/test/test-basic-default
    reason: TemplateRejectedByAPIServer
    status: "False"
    type: ResourcesSubmitted
```

still keeping

```yaml
  - inputs:
    - name: source-provider
    name: image-builder
    outputs:                                                                                                                                                                        - digest: sha256:f360f4f9c5e231a9f7bc989684d18cffaa528d60040b9ca38c4c812ce6e65257
      lastTransitionTime: "2022-03-02T21:27:01Z"
      name: image
      preview: '"10.188.0.3:5000/test/test-basic-default@sha256:b89fdb67be6ffbbbec3d89e410bcba74d757f0b547c1edd9ead617d8b773dbd5"'
    stampedRef:
      apiVersion: kpack.io/v1alpha2
      kind: Image
      name: test-basic                                                                                                                                                                namespace: default
    templateRef:
      kind: ClusterImageTemplate
      name: image
  - inputs:
    - name: image-builder
    name: deployer
    outputs:
    - digest: sha256:d6921e73a5877a89cd5c9e4c152a4c23f65bdda57507fd3a6881e7c9a78e6743
      lastTransitionTime: "2022-03-02T21:27:01Z"
      name: config
      preview: '{"deploy":[{"kapp":{}}],"fetch":[{"inline":{"paths":{"kapp-config.yml":"---\napiVersion:
        kapp.k14s.io/v1alpha1\nkind: Config\nrebaseRules:\n  - path:\n      - metadata\n      -
        annotations\n      - serving.knative.dev/creator\n    type: copy\n    sources:
        [new, existing]\n    resourceMatchers: \u0026matchers\n      - apiVersionKindMatcher:\n          apiVersion:
        serving.knative.dev/v1\n          kind: Service\n  - path:\n      - metadata\n      -
        ...
    stampedRef:
      apiVersion: kappctrl.k14s.io/v1alpha1
      kind: App
      name: test-basic
      namespace: default
    templateRef:
      kind: ClusterConfigTemplate
      name: app-deploy

```

pkg/apis/v1alpha1/common.go
```diff
@@ -191,3 +192,39 @@ type FieldSelectorRequirement struct {
...
+type Input struct {
+	// Name is the name of the resource in the blueprint whose output the resource consumes as an input
+	Name string `json:"name"`
+}
+
+type Output struct {
+	// Name is the output type generated from the resource [url, revision, image or config]
+	Name string `json:"name"`
+
+	// Preview is a preview of the value of the output
```
cirocosta, line 222: given that the value one finds here is a string whose contents are json-encoded,I think we should call that out here in the comments for anyone willing to interpret that somehow (ala `JSON.parse()` in javascript ..)

emmjohnson: since it is a preview, we could be cutting it off and then you would not be able to interpret with JSON.parse

cirocosta: uh yes, that's a good point indeed 👍

PR #678, opened 2022-03-01
Cleanup orphaned objects when blueprint template changes

### LINKED ISSUE #592: When a resource with multiple templates select a different template, then there are not orphan objects
As a follow on to #573 

```
Given: A supply chain that has two templates for a resource, templateA and templateB
AND: a workload that results in templateA being selected
When: A workload is modified to result in templateB being selected
Then: The objects created by templateA are cleaned up
```

This can resolve both ytt templates and options that generate orphans.

martyspiewak:
Co-authored-by: Marty Spiewak <mspiewak@vmware.com>

Cleanup orphaned objects when blueprint template changes

closes #592

## Release Note

Workloads/deliverables cleanup orphaned objects when blueprint template changes

## PR Checklist


pkg/controller/deliverable/reconciler.go
```diff
@@ -260,6 +265,40 @@ func (r *Reconciler) trackDependencies(deliverable *v1alpha1.Deliverable, realiz
...
 
+func (r *Reconciler) cleanupOrphanedObjects(ctx context.Context, previousResources, realizedResources []v1alpha1.RealizedResource) error {
+	log := logr.FromContextOrDiscard(ctx)
+
+	var orphanedObjs []*corev1.ObjectReference
+	for _, prevResource := range previousResources {
+		orphaned := true
+		for _, realizedResource := range realizedResources {
+			if realizedResource.StampedRef.Kind == prevResource.StampedRef.Kind &&
+				realizedResource.StampedRef.Name == prevResource.StampedRef.Name {
```
cirocosta, line 280: wdyt of considering the whole `apiVersion` here? e.g., to account for version bumps (like how we bumped `kpack/v1alpha1` to `kpack/v1alpha2` a couple months ago), or even group changes

pkg/controller/deliverable/reconciler.go
```diff
@@ -260,6 +265,40 @@ func (r *Reconciler) trackDependencies(deliverable *v1alpha1.Deliverable, realiz
...
+	for _, orphanedObj := range orphanedObjs {
+		obj := &unstructured.Unstructured{}
+		obj.SetNamespace(orphanedObj.Namespace)
+		obj.SetName(orphanedObj.Name)
+		obj.SetGroupVersionKind(orphanedObj.GroupVersionKind())
+
+		log.V(logger.DEBUG).Info("deleting orphaned object", "object", orphanedObj)
+		err := r.Repo.Delete(ctx, obj)
+		if err != nil {
+			return err
```
cirocosta, line 299: should we wrap this err? iirc, that's the pattern we use at least in most places

cirocosta (commented):
looking good! left two comments on deliverable, but should be for workloads too 😁

cirocosta:
I didn't get to do proper acceptance on it yet, but I wonder if the GC would suffer from something like I mentioned in https://github.com/vmware-tanzu/cartographer/pull/675#issuecomment-1055915660 where we lose `stampedRef`s (I _think_ **no** because we're always looking at `.status`?)

pkg/controller/deliverable/reconciler.go
emmjohnson: No, because we end up ignoring it

emmjohnson:
> I didn't get to do proper acceptance on it yet, but I wonder if the GC would suffer from something like I mentioned in [#675 (comment)](https://github.com/vmware-tanzu/cartographer/pull/675#issuecomment-1055915660) where we lose `stampedRef`s (I _think_ **no** because we're always looking at `.status`?)

It would have suffered... but we resolved that comment in the other PR 🥇

cirocosta:
it looks really good! I was trying to get it to "break", and one scenario that I've found is the following:

1. create a supplychain that has a resource whose template we want to switch on the fly (i.e., make it flip between say, ConfigMap and Secret)

2. for that ConfigMap, put a finalizer on it that won't ever be acted upon (thus, DELETEs will not really delete)

3. notice that the new object is created, a DELETE is indeed attempted (`metadata.deletionTimestamp` is indeed set), but the object stays there and there will be no further tries (as at that point we already lost that reference from `status.resources`)

tbf, I don't think there's a workaround here - as long as we're able to get a 200OK back from the `DELETE` call we're making, I'd say that indeed, consider it deleted 👍

cirocosta (approved):
lgtm 😁

PR #698, opened 2022-03-08
Validate blueprint selectors

### LINKED ISSUE #692: Validate selectors in webhook so they don't break workloads and deliverables
Currently:
When a supply chain contains a Top Level Selector (Selector, SelectorMatchExpressions, SelectorMatchFields) and any of
them are invalid, then all workloads will stop matching, with a status of Unknown, until the supply chain is repaired.

Preferably:
If the selectors are invalid, then reject the template in a webhook.

Based on this comment from Scott:

> Can that be detected and rejected at admission time? That way it will never be a problem at run time. A value that can never be valid should be rejected via a validating webhook (or similar), while resources that could possibly become valid should be admitted.

_Originally posted by @scothis in https://github.com/vmware-tanzu/cartographer/discussions/686#discussioncomment-2290215_

emmjohnson:

closes #692 

## Release Note

## PR Checklist

### DIFF
--- a/pkg/apis/v1alpha1/common_validations.go
+++ b/pkg/apis/v1alpha1/common_validations.go
@@ -16,10 +16,11 @@ package v1alpha1
 
 import (
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/jsonpath"
 	"reflect"
 	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/jsonpath"
 )
 
 func validateResourceOptions(options []TemplateOption, validPaths map[string]bool, validPrefixes []string) error {
--- a/pkg/apis/v1alpha1/common_validations.go
+++ b/pkg/apis/v1alpha1/common_validations.go
@@ -16,6 +16,7 @@ package v1alpha1
 
 import (
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/jsonpath"
 	"reflect"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/apis/v1alpha1/common_validations.go
```diff
@@ -77,3 +90,26 @@ func validPath(path string, validPaths map[string]bool, validPrefixes []string)
...
+	var err error
+
+	labelSelector := &metav1.LabelSelector{
+		MatchLabels:      selectors.Selector,
+		MatchExpressions: selectors.SelectorMatchExpressions,
+	}
+
+	_, err = metav1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		return fmt.Errorf("selectorMatchExpressions or selectors are not valid: %w", err)
```
cirocosta: uh I think this can be deceiving in case the validation fails for `selector` but _not_ `selectorMatchExpressions`, e.g.:

```yaml
kind: ClusterSupplyChain
metadata:
  name: basic-image-to-url
spec:
  selector:
    foo-: web

  selectorMatchFields:
    - key: spec.image
      operator: Exists
```

has a problem in `selector` but the result from the webhook is:

```
kapp: Error: Applying create clustersupplychain/basic-image-to-url (carto.run/v1alpha1) cluster:
  Creating resource clustersupplychain/basic-image-to-url (carto.run/v1alpha1) cluster:
    API server says: admission webhook "supply-chain-validator.cartographer.com" denied the request: error validating clustersupplychain [basic-image-to-url]: 

 \/ \/ \/ \/ \/ \/ \/
selectorMatchExpressions or selectors are not valid: key:
      Invalid value: "foo-": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') (reason: error validating clustersupplychain [basic-image-to-url]: selectorMatchExpressions or selectors are not valid: key: Invalid value: "foo-": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]'))

```

pkg/apis/v1alpha1/common_validations.go
```diff
@@ -60,10 +63,20 @@ func validateFieldSelectorRequirements(reqs []FieldSelectorRequirement, validPat
 		if !validPath(req.Key, validPaths, validPrefixes) {
 			return fmt.Errorf("requirement key [%s] is not a valid path", req.Key)
 		}
+
+		if err := validJsonpath(req.Key); err != nil {
```
cirocosta, line 67: pretty nice - I wonder if this is something we should do for the `Cluster*Template` outputs (like `imagePath`, etc) given that one could end up messing up their jsonpaths there, thus, having some early validation could be helpful

wdyt? (if 👍 , I think it'd better fit in a different PR though)

pkg/apis/v1alpha1/common_validations.go
emmjohnson: Ya, we chatted about that. We can make the call twice for a better message

pkg/apis/v1alpha1/common_validations.go
emmjohnson: write up an issue!

cirocosta (approved):
nice! lgtm 😁

PR #700, opened 2022-03-08
Markdown linter

squeedee:
Lint markdown in the docs for consistency and editability.


emmjohnson (approved):

cirocosta (approved):
nice! do you think we should also cover other areas of the repo? e.g., README.md .. examples, etc? I think `./examples` is the one that could make the most use of it (aside from `./site`)

PR #705, opened 2022-03-09
Introduce Simplify Runnable Usage RFC

waciumawanjohi:

[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-simplify-runnable-usage/rfc/rfc-0000-simplify-runnable-use.md)

## Release Note

## PR Checklist

### DIFF
--- a/rfc/rfc-0000-simplify-runnable-use.md
+++ b/rfc/rfc-0000-simplify-runnable-use.md
@@ -322,10 +322,12 @@ Runnables will be able to do so.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-This RFC exchanges ease of usage for the app operator for complexity in the
+- This RFC exchanges ease of usage for the app operator for complexity in the
 controller. If we decide that app operators have no issue with creating
 ClusterRunTemplates and templating Runnables, there is no need to undertake
 this work.
+- With this RFC, the workload (supply-chain) controller would have a
+  dependency on the Runnable CRD.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/rfc/rfc-0000-simplify-runnable-use.md
+++ b/rfc/rfc-0000-simplify-runnable-use.md
@@ -275,7 +275,7 @@ directly manage the creation and reporting of objects
 # Prior Art
 [prior-art]: #prior-art
 
-Discuss prior art, both the good and bad.
+Runnable and ClusterRunTemplate are the prior art.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
@@ -312,7 +312,7 @@ Is there need for this behavior on ClusterDeploymentTemplates?
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes
 
-This RFC proposes adding a boolean field named `runnable` to the .spec of:
+This RFC proposes adding an enum field named `lifecycle` to the .spec of:
 - ClusterSourceTemplate
 - ClusterImageTemplate
 - ClusterConfigTemplate
--- a/rfc/rfc-0000-simplify-runnable-use.md
+++ b/rfc/rfc-0000-simplify-runnable-use.md
@@ -10,75 +10,75 @@
 # Summary
 [summary]: #summary
 
-Users will be able to create a Supply Chain Template and use a flag to 
-indicate that the object being stamped out is an immutable object. If they 
-do so, they will need to provide a success criteria. Given these inputs, 
-Cartographer will create a Runnable for the user and ensure that an 
+Users will be able to create a Supply Chain Template and use a flag to
+indicate that the object being stamped out is an immutable object. If they
+do so, they will need to provide a success criteria. Given these inputs,
+Cartographer will create a Runnable for the user and ensure that an
 appropriate ClusterRunTemplate exists.
 
 # Motivation
 [motivation]: #motivation
 
-Runnable exists in Cartographer to allow users to create and leverage 
-resources that are immutable, one time use objects. The prime example of 
-this is Tekton, which uses immutable pipelinerun and taskrun objects to 
-accomplish work. Users can leverage Tekton for testing and more, prime use 
+Runnable exists in Cartographer to allow users to create and leverage
+resources that are immutable, one time use objects. The prime example of
+this is Tekton, which uses immutable pipelinerun and taskrun objects to
+accomplish work. Users can leverage Tekton for testing and more, prime use
 cases Cartographer supports.
 
-While useful, Runnable represents the greatest complexity in Cartographer's 
-user interface. To use an immutable object in a supply chain users must 
-first template the object in a ClusterRunTemplate. Then they must template a 
-Runnable in one of the Cartographer template types used by supply 
+While useful, Runnable represents the greatest complexity in Cartographer's
+user interface. To use an immutable object in a supply chain users must
+first template the object in a ClusterRunTemplate. Then they must template a
+Runnable in one of the Cartographer template types used by supply
 chains/deliveries (e.g. ClusterSourceTemplate, ClusterImageTemplate, etc).
 
 The extra burden of this complication can observed in a number of ways:
 
-1. In office hour conversations, the Cartographer team has been given 
-   feedback that the layering of objects presents a conceptual barrier for 
-   users. In that conversation it was suggested that Cartographer not add 
+1. In office hour conversations, the Cartographer team has been given
+   feedback that the layering of objects presents a conceptual barrier for
+   users. In that conversation it was suggested that Cartographer not add
    more such layers.
-2. I've presented to teams of internal VMware engineers and VMware customers 
-   considering Tanzu Application Platform (for which Cartographer is a core 
-   component). Weeks later I've been asked, to clarify only a single part of 
+2. I've presented to teams of internal VMware engineers and VMware customers
+   considering Tanzu Application Platform (for which Cartographer is a core
+   component). Weeks later I've been asked, to clarify only a single part of
    Cartographer, why Tekton is embedded as part of Runnable.
-3. In our examples we demonstrate how users can create an application 
-   platform from Cartographer. It takes only one example to demonstrate a 
-   simple supply chain that goes from source code to deployment. It takes 
-   two examples to demonstrate how to add testing to a supply chain, one 
-   example focused on using Runnables on their own and another example 
+3. In our examples we demonstrate how users can create an application
+   platform from Cartographer. It takes only one example to demonstrate a
+   simple supply chain that goes from source code to deployment. It takes
+   two examples to demonstrate how to add testing to a supply chain, one
+   example focused on using Runnables on their own and another example
    focused on using Runnable in the context of a supply chain.
-4. In writing tutorials, the same pattern has been necessary: first to 
-   explain Runnable on its own and then to demonstrate how to use Runnable 
+4. In writing tutorials, the same pattern has been necessary: first to
+   explain Runnable on its own and then to demonstrate how to use Runnable
    with Cartographer.
 
-This complication is a needless barrier to entry. Cartographer should 
-be simple to use for our prime use cases, of which testing with Tekton is 
+This complication is a needless barrier to entry. Cartographer should
+be simple to use for our prime use cases, of which testing with Tekton is
 one.
 
-The pattern of using Runnable in a supply chain is predictable and 
-programmable. Rather than asking all app operators to learn this pattern, 
-Cartographer should create Runnables when supply chain templates indicate 
+The pattern of using Runnable in a supply chain is predictable and
+programmable. Rather than asking all app operators to learn this pattern,
+Cartographer should create Runnables when supply chain templates indicate
 their need.
 
 # What it is
 [what-it-is]: #what-it-is
 
-Supply Chain/Delivery templates will add a field that will indicate that the 
-object templated needs to be wrapped by runnable. This will be an enum 
-field, with possible values `mutable` and `immutable`. The field will default to 
-mutable and users will not be required to specify it. The field will be 
+Supply Chain/Delivery templates will add a field that will indicate that the
+object templated needs to be wrapped by runnable. This will be an enum
+field, with possible values `mutable` and `immutable`. The field will default to
+mutable and users will not be required to specify it. The field will be
 called `lifecycle`.
 
-When a template specifies `lifecyle:immutable`, rather than creating the object 
-templated in the supply chain template, Cartographer will create a Runnable 
-and ensure that there is an appropriate ClusterRunTemplate. The 
-ClusterRunTemplate will template out all of the fields that vary and the 
-Runnable will provide all of these fields as inputs. The ClusterRunTemplate 
-will provide outputs that match the outputs of the supply-chain template in 
+When a template specifies `lifecyle:immutable`, rather than creating the object
+templated in the supply chain template, Cartographer will create a Runnable
+and ensure that there is an appropriate ClusterRunTemplate. The
+ClusterRunTemplate will template out all of the fields that vary and the
+Runnable will provide all of these fields as inputs. The ClusterRunTemplate
+will provide outputs that match the outputs of the supply-chain template in
 question (ClusterSourceTemplate, ClusterImageTemplate, etc).
 
 ## Current example
-This is an example of what an app operator would need to do today to use 
+This is an example of what an app operator would need to do today to use
 Runnable in a supply chain (taken from
 https://github.com/vmware-tanzu/cartographer/tree/main/examples/testing-sc):
 
@@ -138,7 +138,7 @@ spec:
 ```
 
 ## Proposed example
-Under this proposal, users would only specify one object with a new immutable 
+Under this proposal, users would only specify one object with a new immutable
 flag.
 
 ```yaml
@@ -177,21 +177,21 @@ spec:
 # How it Works
 [how-it-works]: #how-it-works
 
-When Cartographer processes a template with the immutable flag, rather than 
-creating/updating the templated object it will create/update a Runnable and 
+When Cartographer processes a template with the immutable flag, rather than
+creating/updating the templated object it will create/update a Runnable and
 ensure that a matching ClusterRunTemplate exists.
 
-What is meant by "ensure that a matching ClusterRunTemplate exists"? For any 
-given supply chain template (a cluster scoped object) there will only need 
-to be one ClusterRunTemplate. All of the N Runnables created by N workloads 
-will reference this one ClusterRunTemplate. Only if the supply chain 
-template is updated will this ClusterRunTemplate object need updating. 
+What is meant by "ensure that a matching ClusterRunTemplate exists"? For any
+given supply chain template (a cluster scoped object) there will only need
+to be one ClusterRunTemplate. All of the N Runnables created by N workloads
+will reference this one ClusterRunTemplate. Only if the supply chain
+template is updated will this ClusterRunTemplate object need updating.
 
 ## Inputs
 The templated object will have two types of fields: hard-coded and templated.
-Hard coded fields can simply be copied directly into the ClusterRunTemplate. 
-Templated fields will be translated. Rather than reading from the specified 
-location, these will read from the runnable's inputs field. This can be done 
+Hard coded fields can simply be copied directly into the ClusterRunTemplate.
+Templated fields will be translated. Rather than reading from the specified
+location, these will read from the runnable's inputs field. This can be done
 simply by prepending the template paths with runnable.spec.inputs
 
 Example:
@@ -223,15 +223,15 @@ spec:
 ```
 
 ## Names
-The runnable will be created with the same name assigned to the templated 
+The runnable will be created with the same name assigned to the templated
 object with the object kind appended.
 
-Why append the object kind? We can imagine a supply chain that had one 
-template to create a tekton pipelinerun and another to create a tekton 
-taskrun. The template authors would not worry about name collisions, as they 
-are stamping out different objects. But when they specify that these are 
-immutable objects they may be surprised to see that two objects of the same 
-type are created, which now can have a name collision. To comply with 
+Why append the object kind? We can imagine a supply chain that had one
+template to create a tekton pipelinerun and another to create a tekton
+taskrun. The template authors would not worry about name collisions, as they
+are stamping out different objects. But when they specify that these are
+immutable objects they may be surprised to see that two objects of the same
+type are created, which now can have a name collision. To comply with
 kubernetes name restrictions, slashes in the name must become dashes.
 
 From the example above, assuming a workload with the name `app`:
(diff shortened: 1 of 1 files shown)


rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+## Resource reporting on workload
+The workload reports the resources created from each step. Should this point 
+to the Runnable object or to the most recently successful immutable object 
+created?
```
waciumawanjohi: I think pointing to the most recent successful object should be the behavior. This will match the object that the user has templated out. It will also save the user from going to the runnable and then inspecting the runnable to determine the object which is providing the values reflected there.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+  information exposed to the user at all currently? Perhaps Runnable should 
+  be updated to report this information in its status (similar to workload 
+  reporting the resources).
+
+## Simplify runnables that use 'selected'
+ClusterRunTemplates are able to read values from one object that is 
+pre-existing in the cluster. They do this by specifying values to read from 
+'selected'. The selected object is the object that satisfies the Runnable's
+.spec.selector field. Should we enable users that are leveraging this 
+behavior to use the simplification specified in this RFC?
```
waciumawanjohi: > Should we enable users that are leveraging this behavior to use the simplification specified in this RFC?

Not until regular supply chain templates can also select and read values from arbitrarily selected resources.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+   revisionPath: spec.params[?(@.name=="revision")].value
+   runnable: true                                         # <== Flag
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When Cartographer processes a template with the runnable flag, rather than 
+creating/updating the templated object it will create/update a Runnable and 
+a ClusterRunTemplate.
```
scothis: A namespaces resource creating a cluster scoped resource that other users can see/consume feels off. It would be better to allow the Runnable to inline the template and only create the runnable resource.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+
+and a Runnable with
+```yaml
+kind: Runnable
+spec:
+   inputs:
+      workload.spec.source.git.url: *some-value-from-workload.spec.source.git.url*
+```
+
+The runnable and ClusterRunTemplate will be created with matching names and
```
scothis: This will all but guarantee name collisions when multiple namespaces try to use the same cluster scoped resource name.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+config it will read .status.outputs.config.
+
+# Migration
+[migration]: #migration
+
+This is an additive change and does not present any breaking changes. Users 
+who wish to continue to hand create ClusterRunTemplates and template out 
+Runnables will be able to do so.
+
+# Drawbacks
```
scothis, line 322: This model creates a strong coupling between the blueprint and the runnable, logically it's no longer an optional, semi-independent component.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+              volumeClaimTemplate:
+                 spec:
+                    accessModes:
+                       - ReadWriteOnce
+                    resources:
+                       requests:
+                          storage: 256Mi
+   urlPath: spec.params[?(@.name=="repository")].value    # <== Refers to path on the immutable object that will be created
+   revisionPath: spec.params[?(@.name=="revision")].value
+   runnable: true                                         # <== Flag
```
scothis: favor enums to booleans for fields that change behavior. If/when we want to add a third option, we won't be stuck.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,318 @@
...
+## Simplify runnables that use 'selected'
+ClusterRunTemplates are able to read values from one object that is 
+pre-existing in the cluster. They do this by specifying values to read from 
+'selected'. The selected object is the object that satisfies the Runnable's
+.spec.selector field. Should we enable users that are leveraging this 
+behavior to use the simplification specified in this RFC?
+
+## Success criteria
+Runnable currently hard codes success to be a .status.condition with
+Type:Succeeded and Status:True. E.g. we hard code Runnable to work with
```
scothis: why does a `Runnable` have a `Succeeded` condition? The runnable itself is a living, level-triggered resource for which a `Ready` condition is more approrpriate. A `Succeeded` condition is more appropriate for batch-style resources that run to completion, resources which the `Runnable` is stamping out.

rfc/rfc-0000-simplify-runnable-use.md
scothis: how is "successful" defined?

rfc/rfc-0000-simplify-runnable-use.md
squeedee: I think i prefer `lifecycle: [immutable|mutable]` with mutable the default.

+1 to Scott's sentiments, there's already been discussions about other forms of imprinting/stamping we might do, especially to support "most recently successful" factories for artifacts (not suggesting they're good ideas, but if they do turn out to be, then it'd be nice to have some leeway).

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Runnable/ClusterRunTemplates are currently hard-coded to treat Tekton's success condition as the evidence of success. When the templated object's `status.conditions[?(@.type=="Succeeded")].status == True` the templated object is considered successful.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: That's a very good point... Will update the proposal.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: > logically it's no longer an optional, semi-independent component.

How so? The proposal does not change any existing behavior. Users are free to continue writing Runnables/ClusterRunTemplates directly and even to use them outside the context of a supply chain.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Happy to make this change!

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: > why does a Runnable have a Succeeded condition? ...A Succeeded condition is more appropriate for... resources which the Runnable is stamping out.

My language was not clear here. The success criteria is not for the Runnable object itself, but for the object created by the Runnable. When Runnable stamps out resources, it reflects values from the most recently submitted successful object. A successful object is one with a .status.condition with Type:Succeeded and Status:True.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Proposal updated to clarify that only a single ClusterRunTemplate should be created from any instance of a supply chain template (e.g. multiple supply chains/workloads referring to a ClusterSourceTemplate of an immutable object will result in one ClusterRunTemplate and multiple Runnables).

Good old singleton pattern!

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Addressed here: https://github.com/vmware-tanzu/cartographer/pull/705#discussion_r823296071

rfc/rfc-0000-simplify-runnable-use.md
scothis: The workload/supply chain reconciler now has direct knowledge of Runnable. Previously, it had no more knowledge of Runnable than any other resource defined by a template.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,389 @@
...
+programmable. Rather than asking all app operators to learn this pattern,
+Cartographer should create Runnables when supply chain templates indicate
+their need.
+
+# What it is
+[what-it-is]: #what-it-is
+
+Supply Chain/Delivery templates will add a field that will indicate that the
+object templated needs to be wrapped by runnable. This will be an enum
+field, with possible values `mutable` and `immutable`. The field will default to
```
scothis: enum values are typically capitalized.

```suggestion
field, with possible values `Mutable` and `Immutable`. The field will default to
```

scothis: Another way to think about this is `LevelTriggered` vs `EdgeTriggered`.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,389 @@
...
+              volumeClaimTemplate:
+                 spec:
+                    accessModes:
+                       - ReadWriteOnce
+                    resources:
+                       requests:
+                          storage: 256Mi
+   urlPath: spec.params[?(@.name=="repository")].value    # <== Refers to path on the immutable object that will be created
+   revisionPath: spec.params[?(@.name=="revision")].value
+   lifecycle: immutable                                         # <== Flag
```
scothis, line 174: what other fields from the Runnable need to be included here? The retention policy in particular jumps out as something that might need to be configurable.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,389 @@
...
+
+When Cartographer processes a template with the immutable flag, rather than
+creating/updating the templated object it will create/update a Runnable and
+ensure that a matching ClusterRunTemplate exists.
+
+What is meant by "ensure that a matching ClusterRunTemplate exists"? For any
+given supply chain template (a cluster scoped object) there will only need
+to be one ClusterRunTemplate. All of the N Runnables created by N workloads
+will reference this one ClusterRunTemplate. Only if the supply chain
+template is updated will this ClusterRunTemplate object need updating.
```
scothis, line 188: I'd rather see a Runnable work without requiring a ClusterRunTemplate. The CRT is creating a layer of indirection whose value is negative. The abstraction adds no value in this scenario while we have to hack around its existence.

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Inputs
```
scothis, line 190: if we don't require a CRT, can get rid of managing inputs to the template?

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: The workload/supply chain reconciler now has direct knowledge of *some* Runnables. But Runnable remains a component that can be used outside of the context of Cartographer. And a user could still choose to use the current workflow to use Runnable in a supply chain (e.g. the user creates a ClusterRunTemplate and then templates a Runnable within a supply chain template). If the user does so, the workload/supply chain reconciler will have no special knowledge of that Runnable.

scothis: We have different definitions of what "direct knowledge" means.

Today, we could take the Runnable resource and split it off into an optional extension. The Cartographer core would run just fine without it. After this change, if the Runnable resource was removed, Cartographer itself would break. Cartographer never defines and creates a Runnable resource today, after this change it will.

That's what I mean by "direct knowledge". There's no such thing as partial direct knowledge, either Cartographer knows about the type and how to create it or it doesn't. A template defining how to stamp out a Runnable is not direct knowledge from the perspective of Cartographer, its knowledge from the template.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: 1. I have incredibly lightly held opinions about how the field is named.
2. Runnable stops checking an object's state when it reports a success condition. Is that edge triggered? I tend to reduce edge triggered to "Listen to events" and level triggered to "Watch status".

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: I'm proposing this mostly as syntactic sugar to enable user to get started using resources like Tekton in supply chains. I don't think that it needs to cover every knob and feature of Runnable, (see for example [the question about selector](https://github.com/vmware-tanzu/cartographer/blob/rfc-simplify-runnable-usage/rfc/rfc-0000-simplify-runnable-use.md#simplify-runnables-that-use-selected)). If users get to the level of sophistication that they want to access more advanced Runnable capabilities, then they can write Runnables directly.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: An advantage of this RFC (so far) is that it does not entail any change to the Runnable controller. To be clear, I am quite willing to change the Runnable controller. But the design of Runnable was contentious in the first place, and I'm wary of picking fights that can be avoided.

That said, let's talk about building the right thing, rather than the institutionally easy thing!

## Must we require a ClusterRunTemplate (CRT)
What would a Runnable look like with an embedded template? Let's transform [this example](https://github.com/vmware-tanzu/cartographer/blob/rfc-simplify-runnable-usage/rfc/rfc-0000-simplify-runnable-use.md#current-example)

We would drop the `templateRef` field from the Runnable and add a `template` field. As you note, we could get rid of the `inputs` field from the Runnable. We would need to add the CRT's `outputs` field to the Runnable. The new object would look like this:
```yaml
kind: ClusterSourceTemplate
metadata:
  name: source-linter
spec:
  template:
    kind: Runnable
    metadata:
      name: $(workload.metadata.name)$-linter
    spec:
      serviceAccountName: pipeline-run-management-sa
      template:
        kind: PipelineRun
        metadata:
          generateName: $(workload.metadata.name)$-pipeline-run-
        spec:
          pipelineRef:
            name: linter-pipeline
          params:
            - name: repository
              value: $(workload.spec.source.git.url)$
            - name: revision
              value: $(workload.spec.source.git.revision)$
          workspaces:
            - name: shared-workspace
              volumeClaimTemplate:
                spec:
                  accessModes:
                    - ReadWriteOnce
                  resources:
                    requests:
                      storage: 256Mi
      outputs:
        url: spec.params[?(@.name=="repository")].value
        revision: spec.params[?(@.name=="revision")].value
  urlPath: .status.outputs.url
  revisionPath: .status.outputs.revision
```

I'll point out for clarity, this is not using the lifecycle flag to create a Runnable behind the scenes. This is an example of the classic method of creating a Runnable, with the proposed removal of the CRT.

As I mentioned, this would be a change to the Runnable controller. Specifically, users would be able to write a Runnable with a templateRef and inputs, and then need to write a CRT. Or users could write a Runnable with a template and outputs.

### Side Question: Could we avoid adding the `outputs` field?
The only alternative that I can see here is to mirror the entirety of the most recently successful stamped object in the status of the Runnable. Perhaps in a .status.mostRecentlySuccessful field (to give users a clue as to what they are reading). In that case the Runnable's `outputs` field could be removed. And the outer `urlPath` would point to `status.mostRecentlySuccessful.spec.params[?(@.name=="repository")].value`. The revisionPath would be similar.

### Result
- Runnable user tutorials becomes more complicated. We have to explain the choice before users. Either 1) use CRT, and Runnable with templateRef and inputs; or 2) use Runnable with template and outputs. Failing that we'd at least have to tell them to ignore the template and output fields they see in the Runnable CRD.
- The logic of the Carto's Runnable controller becomes more complicated. The controller has to handle these two different ways of specifying a Runnable.
- The change to the workload/supplychain controller becomes simpler. Rather than ensuring a single instance of a CRT exists, the workload/supplychain controller can always write Runnables that don't reference CRTs.

Is that tradeoff worth it?

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Answered here: https://github.com/vmware-tanzu/cartographer/pull/705#discussion_r825114968

Resolving to keep the conversation streamlined.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Ah! I see what you mean.

Runnable would be able to work separate from Supply Chains, but Supply Chains would be forever coupled to Runnable. I'm happy to write this into the drawbacks. I'm also happy to take on this dependency. As I stated in the RFC, enabling testing is a prime use case for Cartographer. While a Cartographer user could theoretically depend on some other Tekton wrapper instead of Runnable, in practice Carto supply chain users that want to leverage Tekton have to depend on Runnable.

rfc/rfc-0000-simplify-runnable-use.md
scothis: I don't know how meaningful this example is, since a user wouldn't create this resource assuming the RFC is otherwise accepted, they'd use the new syntax. A user could create a Runnable independent of the the template resource. In general, I'm not as concerned about the user experience for the Runnable resource as with this change users need to be less aware of it. Logically, the Runnable resource becomes more of an internal implementation detail than a first class.

What I like most about this approach is that a ClusterRunTemplate resource is no longer required. Without this change, the ClusterRunTempaltes would mix user and system resources.

waciumawanjohi: I've created a separate RFC for adding a "template" field to the Runnable. See here: https://github.com/vmware-tanzu/cartographer/pull/787

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,391 @@
...
+The templated object will have two types of fields: hard-coded and templated.
+Hard coded fields can simply be copied directly into the ClusterRunTemplate.
+Templated fields will be translated. Rather than reading from the specified
```
cirocosta, line 193: In order for this to take place, I imagine we have to limit the scope to jsonpath-based templating as we couldn't do that for `ytt`, is that right?

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,391 @@
...
+```yaml
+kind: Runnable
+spec:
+   inputs:
+      workload.spec.source.git.url: *some-value-from-workload.spec.source.git.url*
+```
+
+## Names
+The runnable will be created with the same name assigned to the templated
+object with the object kind appended.
```
cirocosta, line 227: pretty interesting - I _think_ this has been discussed elsewhere, but, wdyt of using `metadata.generateName`? that way we can be sure that the name we got is not conflicting with anything in that namespace

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,391 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+When Cartographer processes a template with the immutable flag, rather than
+creating/updating the templated object it will create/update a Runnable and
+ensure that a matching ClusterRunTemplate exists.
+
+What is meant by "ensure that a matching ClusterRunTemplate exists"? For any
+given supply chain template (a cluster scoped object) there will only need
+to be one ClusterRunTemplate. All of the N Runnables created by N workloads
```
cirocosta, line 186: I wonder if we're getting into a non crucial optimization here - as ClusterRunTemplate objects just carry data, I wonder if it'd be enough to just always create one for a given runnable (then be able to rely on `generateName` for these). wdyt?

rfc/rfc-0000-simplify-runnable-use.md
```diff
@@ -0,0 +1,391 @@
...
+   urlPath: spec.params[?(@.name=="repository")].value    # <== Refers to path on the immutable object that will be created
+   revisionPath: spec.params[?(@.name=="revision")].value
+   lifecycle: immutable                                         # <== Flag
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+When Cartographer processes a template with the immutable flag, rather than
+creating/updating the templated object it will create/update a Runnable and
```
cirocosta, line 181: have you thought about how this would impact `workload.status.resources[*].stampedRef`? I wonder if it'd be strange for someone to see in their templates that they're stamping out `kind: PipelineRun` and then they see `Runnable` there.

I particularly really like the fact that this proposal keeps that approach of the direct children being all level triggered (via the use of Runnable) though

waciumawanjohi: I discuss the .status.resources here:
https://github.com/vmware-tanzu/cartographer/blob/rfc-simplify-runnable-usage/rfc/rfc-0000-simplify-runnable-use.md#resource-reporting-on-workload

I'll note that in Office Hours Rash advocated for not creating a Runnable, and rather moving this functionality into the workload controller's code:
https://youtube.com/clip/Ugkx0s1Qy_CEr7aG2QWaiMQK1N01PRjCXNAu

I am fully amenable to that change if there is a member of the technical oversight committee that will approve that change.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: For every ClusterSourceTemplate that stamps out an immutable object, there is one logical ClusterRunTemplate. That ClusterRunTemplate is determinative (it uses no information from the workload). I don't think we should create N ClusterRunTemplates which are all identical except for their name, when instead we could create just one.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: I'm not opposed to using generate name for the runnable. I expect it will require slightly more processing by our workload, as we would need to list all the runnables in the namespace and filter for the one that we want.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Ensuring that an expected object exists is also standard Carto behavior. For every workload, the controller ensures that the expected objects have been created.

waciumawanjohi: In some discussions the concern for a race condition has been raised. I do not believe that is a concern. In the worst case scenario a user defines a ClusterSourceTemplate with an immutable lifecycle. This template is then referenced by both a delivery and a supply chain. A reconciliation is kicked off for both at the same time. Both check for the existence of the expected ClusterRunTemplate. Neither see it. Both submit the same ClusterRunTemplate with the same name (because it is based on the name of the ClusterSourceTemplate). One submission is handled first (and creates the object), the other second (and updates the object, with no changes). No conflict exists.

I welcome further poking of this, but I do not see a concern about name collisions or race conditions.

rfc/rfc-0000-simplify-runnable-use.md
waciumawanjohi: Yes, that's a good call out. I think that is fine, as the goal of this RFC is to provide a convenience method for easy cases, not to replace the direct usage of Runnable and CRT in all cases.

cirocosta:
marking as blocked as we'd like to tackle https://github.com/vmware-tanzu/cartographer/pull/787 first

PR #718, opened 2022-03-16
Run validation before releasing

martyspiewak:

* Updates release workflow to only create releases off tags from commits on main.
* Also only creates release if tag is formatted correctly

## PR Checklist


idoru (approved):
LGTM, thanks!

scothis (changes requested):
Do we never expect to create a release from a release branch? Or back port a bug fix or security patch to a previous release that isn't on main?

emmjohnson (approved):

idoru:
@scothis  As of #717 the release is a separately a triggered action and doesn't run tests which now run for every commit on `main`.

If that's the model going forward, when we have release branches then I imagine we'll also run tests for all commits on those too, and expand this restriction.

cirocosta:
@idoru so you're saying that we could in `validation.yaml` remove

https://github.com/vmware-tanzu/cartographer/blob/fa8800dba24ff1ea01256ae0537ac06670274fb6/.github/workflows/validation.yaml#L15-L20

by something like 

```yaml
name: validation
on:
  push:
```

so we run validation on every commit anywhere, and then drop the check this PR is adding? so that we could, before pushing a tag (from `main` or a release branch), make sure that the validation did take place, everything went well, and we can release from it?

I think keeping the ability to release from a tag regardless of the branch is very important

martyspiewak:
@cirocosta I'd recommend something more like:
```
name: validation
on:
  push:
    branches:
      - main
      - release-*
```
or something to that effect, rather than running validations on every commit to any spike branch we create.

But are you also saying you'd prefer that the check introduced in this PR be removed?

scothis:
>or something to that effect, rather than running validations on every commit to any spike branch we create.

That's what forks are for.

> so we run validation on every commit anywhere

It's not every commit, it's every **push**, just because a commit is on a branch doesn't mean that the workflow will have run for that commit. And certainly doesn't mean that the workflow passed.

If there's one commit I want to makes sure passes all validation it's a tag. Assuming the tagged commit had all the tests pass, I'd still want the release artifacts to undergone the same testing. Two builds of the same source will be very similar, but not identical.

For example, [this is a workflow I've found to work well](https://github.com/vmware-tanzu/cartographer-conventions/actions/runs/1935647268). It builds an artifact once, runs a battery of acceptance tests and then if the run is for a tag, creates a draft release with those artifacts. The released bits are the bits the tests ran against.

.github/workflows/release.yaml
```diff
@@ -17,45 +17,86 @@ name: release
 on:
   push:
     tags:
-      - 'v*'
+      - 'v[0-9]+\.[0-9]+\.[0-9]+(-build\.[0-9]+)?'
 
 jobs:
-  release:
+  check-if-main:
```
idoru: I feel that the `check-if-main` probably isn't necessary now that we're running tests on release again.

idoru (changes requested):

idoru (approved):
LGTM 👍

PR #727, opened 2022-03-18
Return an error if a stamped resource tries to override the namespace

### LINKED ISSUE #614: Stamped resources should not be allowed to specify a namespace
### Bug description:

When specifying a resource to be stamped using ytt, a user is able to specify a destination namespace.

### Steps to reproduce:

1. Write any template using ytt
2. Specify a namespace for the stamped object `namespace-a`
3. Create a workload in `namespace-b`
4. Notice that the stamped object is indeed placed in `namespace-a`

### Expected behavior:

All stamped objects are placed in the same namespace as the workload.

### Actual behavior:

Stamped objects can be placed in any namespace.

### Additional Context

Stamping using conventional templating works as expected.

emmjohnson:

closes #614 

## Release Note

## PR Checklist


pkg/apis/v1alpha1/cluster_template.go
```diff
@@ -56,6 +58,8 @@ type TemplateSpec struct {
 	// interpolation using the $()$ marker format. For more
 	// information, see: https://cartographer.sh/docs/latest/templating/
 	// You cannot define both Template and Ytt at the same time.
+	// You cannot define the namespace for the resource. The resource will
+	// be created in the owner namespace.
```
martyspiewak: This sounds a bit like we'll ignore the namespace and just create in the owner namespace, rather than fail (which is the behavior this PR is introducing). Is there a way we can make that a bit clearer? i.e. something like:
```
You should not define the namespace for the resource- it will automatically be created in the owner namespace. If the namespace is specified and is not the owner namespace, the resource will fail to be created.
```

martyspiewak (changes requested):
Looks great! Just left a comment about the documentation that seems to be a bit misleading. Let me know what you think.

pkg/apis/v1alpha1/cluster_run_template.go
```diff
@@ -46,6 +46,9 @@ type RunTemplateSpec struct {
 	// the blueprint is applied. Templates support simple value
 	// interpolation using the $()$ marker format. For more
 	// information, see: https://cartographer.sh/docs/latest/templating/
+	// You should not define the namespace for the resource - it will automatically
+	// be created in the owner namespace. If the namespace is specified and is not
+	// the owner namespace, the resource will fail to be created.
```
martyspiewak, line 51: Shoulda mentioned this originally sorry- this part is only true for the YTT section. For the template, validation will stop users from specifying a namespace so the docs should reflect that. i.e. something like what you had before:
```
You cannot define the namespace for the resource. The resource will be created in the owner namespace.
```

pkg/apis/v1alpha1/cluster_template.go
```diff
@@ -47,6 +47,9 @@ type TemplateSpec struct {
 	// interpolation using the $()$ marker format. For more
 	// information, see: https://cartographer.sh/docs/latest/templating/
 	// You cannot define both Template and Ytt at the same time.
+	// You should not define the namespace for the resource - it will automatically
+	// be created in the owner namespace. If the namespace is specified and is not
+	// the owner namespace, the resource will fail to be created.
```
martyspiewak, line 52: Same as above- this is only true fro ytt. For the template, it should be:
```
You cannot define the namespace for the resource. The resource will be created in the owner namespace.
```

martyspiewak (changes requested):

martyspiewak (approved):
Lgtm!

PR #728, opened 2022-03-18
Add upgrade test script and workflow

### LINKED ISSUE #417: Upgrade testing
### Description of problem

We want to guarantee an upgrade path for users.

### Proposed solution

Add a workflow that deploys the previous release, and tests an upgrade to the current candidate.

martyspiewak:

* Adds script and workflow to test upgrades
* Workflow runs of every PR and push to main, but will not block merging if it fails.

closes #417 

## PR Checklist

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -1,0 +1,45 @@
+# Copyright 2021 VMware
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: upgrade test
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  upgrade-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+        with: {fetch-depth: 0}
+
+      - name: set up Go
+        uses: actions/setup-go@v2
+        with: {go-version: '^1.17'}
+
+      - name: setup base dependencies
+        run: |-
+          sudo apt update && sudo apt install -y curl tar
+          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree grype
+
+      - name: setup carvel tooling binaries
+        uses: vmware-tanzu/carvel-setup-action@v1
+
+      - name: run upgrade test
+        run: |-
+          sudo ./hack/ci/trust-local-registry.sh
+          ./hack/upgrade-test.sh
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: setup base dependencies
         run: |-
           sudo apt update && sudo apt install -y curl tar
-          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree grype
+          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree
 
       - name: setup carvel tooling binaries
         uses: vmware-tanzu/carvel-setup-action@v1
--- a/hack/upgrade-test.sh
+++ b/hack/upgrade-test.sh
@@ -1,0 +1,282 @@
+#!/usr/bin/env bash
+# Copyright 2021 VMware
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# shellcheck disable=SC2155
+readonly DIR="$(cd "$(dirname "$0")" && pwd)"
+readonly HOST_ADDR=${HOST_ADDR:-$("$DIR"/ip.py)}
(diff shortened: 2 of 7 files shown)


.github/workflows/upgrade-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+        with: {fetch-depth: 0}
+
+      - name: set up Go
+        uses: actions/setup-go@v2
+        with: {go-version: '^1.17'}
+
+      - name: setup base dependencies
+        run: |-
+          sudo apt update && sudo apt install -y curl tar
+          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree grype
```
cirocosta:
```suggestion
          ./hack/ci/install-binaries.sh ko kubebuilder kuttl tree
```

hack/upgrade-test.sh
```diff
@@ -0,0 +1,282 @@
...
+readonly TEST_NAME="gitwriter-sc"
+readonly GIT_SERVER="git-server.default.svc.cluster.local:80"
+readonly SOURCE_REPO="hello-world"
+readonly SOURCE_BRANCH="main"
+readonly CONFIG_REPO="gitops-test"
+readonly CONFIG_BRANCH="main"
+readonly CONFIG_COMMIT_MESSAGE="Update config"
+
+main() {
+  ./hack/setup.sh cluster cartographer-latest example-dependencies
```
cirocosta:
```suggestion
  $DIR/setup.sh cluster cartographer-latest example-dependencies
```

or have a `cd $DIR` as the very first thing in `main`

---

in the scripts we usually either:

- in `main`, change directory to an absolute path (usually the root of the repository, like `cd $ROOT`
- in functions, refer to files with absolute paths (like, `ytt $DIR/overlays/...` where `$DIR` is an absolute path to a directory)

such that it doesn't matter where we execute the scripts from

wdyt?

hack/upgrade-test.sh
```diff
@@ -0,0 +1,282 @@
...
+    kubectl tree deliverable gitops
+  done
+
+  log "testing gitops-to-app FAILED :("
+  exit 1
+}
+
+verify_new_carto_deployed() {
+  until [[ $(kubectl -n cartographer-system get deployment cartographer-controller -o json | jq '.metadata.labels."app.kubernetes.io/version"') == '"v0.0.0-dev"' &&
+    $(kubectl -n cartographer-system get deployment cartographer-controller -o json | jq '.status.availableReplicas') != 0 ]]; do
```
cirocosta, line 234: I'm not tooootally sure but I think `k rollout status deployment --watch --timeout <something>` _should_ do it, but no strong opinion on moving from what we have here to what I suggested as `rollout` ends up not giving us that nice progress os _"hey, I'm trying ..."_ from time to time

hack/upgrade-test.sh
```diff
@@ -0,0 +1,282 @@
...
+  log "setting up git server"
+
+  kubectl apply -f ./hack/git-server.yaml
+
+  until [[ $(kubectl get deployment git-server -o json | jq '.status.readyReplicas') == 1 ]]; do
+    log "waiting for git-server deployment"
+    sleep 10
+  done
+
+  (kubectl port-forward service/git-server $port:80 &) >/dev/null
```
cirocosta: I think this will leak the port forwarding process behind 🤔  

when doing something similar, I remember going for the following

```
                kubectl -n $namespace port-forward service/$name $port:80 &
                trap "kill $! || true" EXIT
```

that `trap` there gives us the ability of having that `kill <pid>` going on regardless of how we exit the process (sure, a SIGKILL on bash would prevent it 😅  but regular terminations would all get the kill going)

cirocosta:
ran it locally, working good!

on CI, sounds like we could make use of the overlay that gets rid of resource reservations/limits:

```
11:20:20PM: ---- waiting on 1 changes [14/15 done] ----
11:20:20PM:  L ongoing: waiting on pod/cartographer-controller-674c998659-w722h (v1) namespace: cartographer-system
11:20:20PM:     ^ Pending: Unschedulable (message: 0/1 nodes are available: 1 Insufficient cpu.)
                                                                                   ^^^^


11:21:15PM: fail: reconcile deployment/cartographer-controller (apps/v1) namespace: cartographer-system
11:21:15PM:  ^ Deployment is not progressing: ProgressDeadlineExceeded (message: ReplicaSet "cartographer-controller-674c998659" has timed out progressing.)
```

e.g., how we do

```bash
install_cert_manager() {
        ytt --ignore-unknown-comments \
                -f "$DIR/overlays/remove-resource-requests-from-deployments.yaml" \       # (!)  this 
                -f https://github.com/jetstack/cert-manager/releases/download/v$CERT_MANAGER_VERSION/cert-manager.yaml |
                kapp deploy --yes -a cert-manager -f-
}
```

hack/upgrade-test.sh
cirocosta: tbh, we _might_ actually not even need the `sleep` and waiting given that `kapp` does the waiting for us I believe (although, still, nice to make sure we're pointing at the right version - the 0.0.0-dev one)

hack/upgrade-test.sh
```diff
@@ -0,0 +1,282 @@
...
+      --data-value git_commit_message="$CONFIG_COMMIT_MESSAGE")
+}
+
+test_source_to_gitops() {
+  local config_dir=$1
+  local port=$2
+
+  log "testing source to gitops"
+
+  SUCCESS=false
```
cirocosta: I think these upper case variables could be not only lowercase, but also marked `local` as they're supposed to only be available inside the function - iirc, we've been reserving upper case ones for those that have definitions at the global scope usually with some default that can be overwritten via environment variables

hack/upgrade-test.sh
```diff
@@ -0,0 +1,282 @@
...
+
+  pushd "$config_dir"
+    git clone "http://localhost:$port/$CONFIG_REPO.git"
+    pushd "$CONFIG_REPO"
+      for sleep_duration in {20..1}; do
+        git fetch --all --prune
+
+        git checkout "$CONFIG_BRANCH" >/dev/null 2>/dev/null || { echo "- waiting $sleep_duration seconds" && sleep "$sleep_duration" && continue; }
```
cirocosta: I believe this could be split like

```bash
                git checkout "$CONFIG_BRANCH" >/dev/null 2>/dev/null || {
                        echo "- waiting $sleep_duration seconds"
                        sleep "$sleep_duration" 
                        continue
                }
```

hack/upgrade-test.sh
martyspiewak: Ah interesting. Felt it was good to explicitly make sure we're pointing to the right version but I can switch it if you prefer.

cirocosta: no worries - I  think being explicit there is good 👍 👍  thank you!

cirocosta (approved):

PR #738, opened 2022-03-23
Introduce RFC for allowing resource to report status

### LINKED ISSUE #694: R4RFC: Allow resources to report status
### Description of problem

Currently a `SupplyChain` has no knowledge of the current status of any of its underlying resources. There have been many attempts to document how we could let resources report this information, but most of these solutions also try to use this information to gate downstream behaviour.  We are requesting an RFC that proposes a solution for how resources can report (and categorize) this information strictly for informational purposes. The goal would be to let templates indicate what state the resource might be in (ready, failing, etc) so that we can relay that information to the workload status.

### Prior art
https://github.com/vmware-tanzu/cartographer/pull/556
https://github.com/vmware-tanzu/cartographer/issues/28
https://github.com/vmware-tanzu/cartographer/issues/162
https://github.com/vmware-tanzu/cartographer/issues/247

martyspiewak:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-resources-report-status/rfc/rfc-0000-allow-resources-to-report-status.md)

closes: #694 

## PR Checklist


rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+3. 'Unknown'
+
+It is up to each template author to determine what "success" and "failure" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not be able to determine success or failure.
+
+A `ClusterResourceStatusRule` requires defining matchers for both the `Succeeded` state and the `Failed` state.
+The matchers in the `Succeeded` state are ANDed together, whereas the matchers in the `Failed` state are ORed. That is, a resource is considered to have `Succeeded` only if _all of_ the matchers are fulfilled, but is considered to have `Failed` if _any one of_ the matchers are fulfilled.
+
+The supported matchers are `matchConditions` and `matchFields`. Each is an array, and an author can define either or both of these matchers for a given state.
+
+For `matchConditions`, authors need only define the `type` and `status` of the condition for Cartographer to read. In the owner status, the full condition will be reflected (with the `reason`, `message`, etc.)
```
cirocosta: if the `matchConditions` array is populated with more than one item (say, 2), would we then bubble up to the owner status both of the conditions that matched?

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+  succeeded:
+    matchConditions:
+      - type: ReconcileSucceeded
+        status: 'True'
+  failed:
+    matchFields:
+      - key: 'status.conditions[?(@.type=="ReconcileFailed")].status'
+        operator: 'In'
+        values: ['True']
+        messagePath: '.status.usefulErrorMessage'
```
cirocosta: neat!

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+
+The supported matchers are `matchConditions` and `matchFields`. Each is an array, and an author can define either or both of these matchers for a given state.
+
+For `matchConditions`, authors need only define the `type` and `status` of the condition for Cartographer to read. In the owner status, the full condition will be reflected (with the `reason`, `message`, etc.)
+For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
+
+Each template will now have a new field in the spec `resourceStatusRuleRef` where authors can specify the `ClusterResourceStatusRule` to use for that template. If no `ClusterResourceStatusRule` is defined, Cartographer will default to listing the resource as `Succeeded` once it has been successfully applied to the cluster and any relevant outputs have been read off the resource.
```
cirocosta, line 251: `Succeed` by default indeed makes sense for some resources like `configmap`, but I'm not very sure that's a good default overall - I feel like we should default to `Unknown` as per the description of it

> _"Unknown" strictly represents that Cartographer has not be able to determine success or failure._

it sounds quite fit for that 🤔

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the two states be `Succeeded` and `Failed`?
+- Is it right for Cartographer to default to `Succeeded`?
+
+# Spec. Changes
+[spec-changes]: #spec-changes
+- All templates will now have an optional `resourceStatusRuleRef` which will take a `name`. In the future, it could also take a type and namespace if we introduce namespace-scoped resourceStatusRules.
```
cirocosta: made me wonder if, similar to ClusterSupplyChain/ClusterDelivery where we make use of `.status` to report whether the supply chain / delivery is well formed, the templates should now include a status that would reveal whether they're ready or not as it could be that one creates a template that points `resourceStatusRuleRef` at an object that doesn't exist 🤔

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+Why should we *not* do this?
+- The owner status will potentially grow significantly, based on how many matchers are in a `ClusterResourceStatusRule`.
+- We only allow for reflecting three states, though this design allows for creating more states in the future.
+- This design does not allow authors to define the states themselves.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - The `ClusterResourceStatusRule` could instead be inlined in the templates.
```
cirocosta: I think another alternative here is allowing both options:

- inline, if you'd like a one-off that is very specific and unlikely to be shared (e.g., for a configmap)
- ref, if something like `ready: true | ready: false` where it'd be reused a lot

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,320 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the two states be `Succeeded` and `Failed`?
+- Is it right for Cartographer to default to `Succeeded`?
```
cirocosta: `Succeed` by default indeed makes sense for some resources like `configmap`, but I'm not very sure that's a good default overall - I feel like we should default to `Unknown` as per the description of it

> _"Unknown" strictly represents that Cartographer has not be able to determine success or failure._

it sounds quite fit for that 🤔  

There'd definitely be the implication of making this a a breaking change though.

---

any thoughts from folks around defaulting to Unknown instead?

rfc/rfc-0000-allow-resources-to-report-status.md
emmjohnson: Yes

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Ah that's a good point. That would probably be worthwhile to add. Will add that to the RFC.

rfc/rfc-0000-allow-resources-to-report-status.md
squeedee: If there is more than one, which is the canonical top level status. That is, looking at a particular resource in the workload, whats the source of truth for the status of the resource?

martyspiewak: We will bubble up anything the template author provides and that will be the source of truth. That is not for Cartographer to determine.

squeedee: In the in person design discussion, we were talking about being able to provide one entry to refer to (including programatically) for the state of an underlying (stamped) resource. I thought the idea here was to help developers rapidly identify where their issue is, including in visualisation tools?

squeedee: I think i'm just being slow. The source of truth is stampedStatus.status ?

martyspiewak: `stampedStatus.status` is the overall status of the resource from cartographer's perspective. That status is determined by reconciling the matchers defined (so it will be `Succeeded` if all matchers under `succeeded` match, `Failed` if any matched under `failed` matches, or otherwise `Unknown`).

matt-royal:
This would be very helpful for the uses of Cartographer that we're considering in https://github.com/cloudfoundry/cf-k8s-controllers. We need a SupplyChain to yield a built image, but we don't want to be coupled to the specific resources that the SupplyChain creates. This would allow us to determine if the SupplyChain has failed for a given Workload (i.e if any of the resources created failed).

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+- The owner status will potentially grow significantly, based on how many matchers are in a `ClusterResourceStatusRule`.
+- We only allow for reflecting three states, though this design allows for creating more states in the future.
+- This design does not allow authors to define the states themselves.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - The `ClusterResourceStatusRule` could instead be inlined in the templates.
+    - Separating these rules in a separate CRD allows for re-usability of those rules.
```
scothis: Is the configuration sufficiently reusable to justify an independent resource?

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+  - We could also not require the templates to specify the `ClusterResourceStatusRule` and instead have those rules automatically select templates based on the `kind` and `apiVersion` of the resource they are stamping out.
+    - This could be a bit confusing because of ytt. A template could end up stamping out different types of resources and so Cartographer would only be able to inform users which `ClusterResourceStatusRule` is used for a given template after an owner is applied, and it could then be reflected in the owner status.
```
scothis: This isn't a limitation created by ytt, while maybe not the most useful thing, the apiVersion/kind can be templated without ytt.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+- We only allow for reflecting three states, though this design allows for creating more states in the future.
+- This design does not allow authors to define the states themselves.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  - The `ClusterResourceStatusRule` could instead be inlined in the templates.
+    - Separating these rules in a separate CRD allows for re-usability of those rules.
+  - We could also not require the templates to specify the `ClusterResourceStatusRule` and instead have those rules automatically select templates based on the `kind` and `apiVersion` of the resource they are stamping out.
```
scothis: This is interesting, we implemented this pattern within Service Bindings to know how to write into a workload resource.

The gotcha is that you need to define a rule for each type exactly once, and with kapp deployments, only a single app can define a given resource. So there will be contention between different authors about who gets to define the rules for a particular resource. If there are legitimately different interpretations of the status for the same resource, you're stuck.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+
+# Prior Art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the two states be `Succeeded` and `Failed`?
```
scothis: "Succeeded" and "Failed" are terms associated with edge triggered processes that run to completion. The resources the state represents are not edge triggered, nor can they complete. One of the fundamental misunderstandings of the role of Cartographer is that is orchestrates an execution pipeline. Using terms associated with those concepts perpetuated that misunderstanding.

Much of this RFC is written in past tense, "Succeeded" and "Failed" are past tense. We're not concerned with the past state, it's the present known state. (setting aside the speed of light and that everything we can perceive is from the near past).

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+- Owner `status.resources` will now have `stampedStatus` on each resource.
+- A new CRD, `ClusterResourceStatusRule` will be created with the following spec:
+```yaml
+---
+apiVersion: carto.run/v1alpha1
+kind: ClusterResourceStatusRule
+metadata: {}
+spec:
+
+  succeeded:
```
scothis: Some resources will report details about why something is unknown that could be useful to propagate. For example, if a resource goes into "Unknown" while it performs a long lived task, that can be useful to distinguish from other types of "Unknown".

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the two states be `Succeeded` and `Failed`?
+- Is it right for Cartographer to default to `Succeeded`?
+
+# Spec. Changes
+[spec-changes]: #spec-changes
+- All templates will now have an optional `resourceStatusRuleRef` which will take a `name`. In the future, it could also take a type and namespace if we introduce namespace-scoped resourceStatusRules.
+- Owner `status.resources` will now have `stampedStatus` on each resource.
```
scothis: is it worth summarizing the shape of stampedStatus here?

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,322 @@
...
+  - This will also enable integrations that may be looking to create a UI around Cartographer, allowing the UI to display the health of the blueprint without having to make an API call to k8s to retrieve the status of every resource, and without having to know how to define health for every stamped object type.
+- What is the expected outcome?
+  - Users will have more insight into the progression of the blueprint and the health of each underlying resource.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC proposes creating a new CRD, `ClusterResourceStatusRule`. Template authors use this CRD to define the conditions or fields on a stamped resource that determine whether the resource is succeeding or failing. The template author then references this `ClusterResourceStatusRule` from their template.
+
+Cartographer will attempt to read the specified conditions and/or fields and reflect this state on the owner status. If Cartographer cannot determine success or failure, it will show the state as unknown.
```
jwntrs: What if these states were `healthy`, `unhealthy` and `unknown`?

martyspiewak: Updated the RFC with those states

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,321 @@
...
+    - name: service-creator
+      outputs:
+        ...
+      stampedRef:
+        ...
+      templateRef:
+        apiVersion: carto.run/v1alpha1
+        kind: ClusterTemplate
+        name: servicer
+      stampedStatus:
```
squeedee: I want to suggest converting this to one of many possible conditions

```yaml
conditions:
  - type: Healthy
    status: False
    reason: MatchedUnhealthy 
    message: field 'status.conditions[?(@.type=="ReconcileFailed")].status' was 'some-value' with message '...', field..
```  

This leaves room for a top level type (such as "Ready") and for other resource related conditions, some of which do not come from the stamped object, eg Submitted.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: I'd think so. For example, you may have multiple resource types for which health could be defined similarly, like `Ready: True`. But even if you assume each resource type would need its own definition of health, you may have multiple templates that stamp out the same resource type, so it'd be nice to not have to redefine those rules.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: That's fair. I'll update the RFC to mention that.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Good point. I imagine we could find a way around that, like rules selecting templates only if the template doesn't explicitly set the rule, but I think it's probably best for now to not have the selection behavior.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: The RFC has already been updated to use the terms `Healthy` and `Unhealthy` instead.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: I'd be open to adding 'unknown' as a field, but then how would we distinguish between `unknown` as defined by the ClusterResourceStatusRule, and unknown where none of the matchers have resolved and cartographer truly does not know the state?

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Makes sense. Will add that to the RFC.

rfc/rfc-0000-allow-resources-to-report-status.md
emmjohnson: We do allow fields and conditions to be defined. This example is only showing fields that were defined in this rule - https://github.com/vmware-tanzu/cartographer/blob/f3e3254f8cdfa0c960a9333864478a0638e33eb4/rfc/rfc-0000-allow-resources-to-report-status.md?plain=1#L182

rfc/rfc-0000-allow-resources-to-report-status.md
jwntrs: I think this is about surfacing additional information alongside the unknown condition so that the user can make the distinction on their own.

martyspiewak: So something like an extra field for `unknownMessagePath`? Or `uknown.messagePath` in case we want to add more later?

jwntrs: I think it makes sense to have the same syntax (matchFields, matchConditions) for every possible state, so `Unknown` wouldn't be any different. It would mean that int the workload status, sometimes `Unknown` has a corresponding `fields` or `status` entry, and sometimes not.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+    - name: source-provider-2
+      outputs:
+        ...
+      stampedRef:
+        ...
+      templateRef:
+        apiVersion: carto.run/v1alpha1
+        kind: ClusterSourceTemplate
+        name: source
+      stampedStatus:
```
squeedee:
```suggestion
      stamped:
```

Redundant terminology. this maps:

```
  stampedStatus.status -> stamped.status
  stampedStatus.conditions -> stamped.conditions
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+# Summary
+[summary]: #summary
+
+The RFC proposes modifying the existing templates to allow template authors to indicate how Cartographer can read the state of the resource so that it can be reflected on the owner status.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Currently, a blueprint has no knowledge of the current status of its underlying resources. This RFC proposes how resources can report (and categorize) this information strictly for informational purposes.
```
scothis: "underlying resources" for a blueprint is ambiguous. It could reference either the template or stamped resources.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+    - name: resource1
+      ...
+      stampedStatus:
+        status: <[Healthy|Unhealthy|Unknown]>
+        conditions:
+          - <metav1.Condition>
+        fields:
+          - key: <string>
+            value: <string>
+            message: <string>, omitempty
```
scothis: Can we unify these structures so that users don't need to do that work themselves?

A rough idea would be to surface conditions for each resource

```yaml
status:
  resources:
  - name: resource1
    conditions: <metav1.Condition>
    ...
```

These conditions are not a copy of the stamped resource's status conditions, but a set of conditions specific to the resource as viewed by Cartographer. For the purposes of this RFC, we're concerned about a `Heathy` condition summarizing the health of the stamped resource, other condition can be introduced to surface other failure points for a particular resource, like an error applying the template, or an api server error on create/update.

In practice, this would look like:

```yaml
status:
  resources:
  - name: resource-condition
    conditions:
    - type: Healthy
      status: 'True'
      reason: ReadyCondition
      message: ...copied from the stamped resource's Ready condition's message...
      lastTransitionTime: 2022-03-17T00:29:04Z # managed by cartographer
  - name: resource-field
    conditions:
    - type: Healthy
      status: 'False'
      reason: MatchedField
      message: ...defined by the rule's message path...
      lastTransitionTime: 2022-03-17T00:29:04Z # managed by cartographer
```

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: I don't really like `stampedStatus`  as the top level field either, but we already have `stampedRef` as a top level field so I wouldn't want to make another field `stamped`.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+  name: ready-rule
+spec:
+  healthy:
+    matchConditions:
+      - type: Ready
+        status: 'True'
+  unhealthy:
+    matchConditions:
+      - type: Ready
+        status: 'False'
```
scothis: For resources that expose an aggregate condition, it would be nice to reference that condition type once and not need to distinguish between True/False/Unknown. Resource that split the health across multiple conditions  are less common than those that aggregate to a single condition.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
+
+Each template will now have a new field in the spec `resourceStatusRuleRef` where authors can specify the `ClusterResourceStatusRule` to use for that template. If no `ClusterResourceStatusRule` is defined, Cartographer will default to listing the resource as `Healthy` once it has been successfully applied to the cluster and any relevant outputs have been read off the resource.
+
+Templates will now also have a status with a `Ready` condition which will be `True` if the referenced `ClusterResourceStatusRule` exists (or if no rule is referenced), and `False` otherwise. Blueprints will now wait to report that they are `Ready` until their referenced templates are also `Ready`.
```
scothis: We can avoid needing to reconcile all of the template types by inlining the status rules

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+
+A `ClusterResourceStatusRule` requires defining matchers for both the `Healthy` state and the `Unhealthy` state.
+The matchers in the `Healthy` state are ANDed together, whereas the matchers in the `Unhealthy` state are ORed. That is, a resource is considered to be `Healthy` only if _all of_ the matchers are fulfilled, but is considered to be `Unhealthy` if _any one of_ the matchers are fulfilled.
+
+The supported matchers are `matchConditions` and `matchFields`. Each is an array, and an author can define either or both of these matchers for a given state.
+
+For `matchConditions`, authors need only define the `type` and `status` of the condition for Cartographer to read. In the owner status, the full condition will be reflected (with the `reason`, `message`, etc.)
+For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
```
scothis: How is the messagePath aggregated when there are multiple matches?

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,335 @@
...
+            value: <string>
+            message: <string>, omitempty
+```
+- A new CRD, `ClusterResourceStatusRule` will be created with the following spec:
+```yaml
+---
+apiVersion: carto.run/v1alpha1
+kind: ClusterResourceStatusRule
+metadata: {}
+spec:
```
scothis: there should be an unambiguous way to state a resource has no health, and is implicitly healthy.

rfc/rfc-0000-allow-resources-to-report-status.md
squeedee: It still feels "right" so I wonder if we use a two step approach to ratifying it:

1. denormalize:
```yaml
status:
  resources:
    - name: fred
      stampedRef:
        apiVersion: v1
        kind: ConfigMap
      stamped:
        ref:
          apiVersion: v1
          kind: ConfigMap
        status: "Unknown"
        conditions: [...] 
```
(tbh `ref:` feels unnecessary too, `apiVersion` and `kind` could inline under `stamped:`)

2. Create a migration in v1alpha2 (when appropriate) to remove `status.resources[].stampedRef`

I recognize this is more work, but I'd rather we get the shape right, at least eventually :D

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+# Summary
+[summary]: #summary
+
+The RFC proposes modifying the existing templates to allow template authors to indicate how Cartographer can read the state of the resource so that it can be reflected on the owner status.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Currently, a blueprint has no knowledge of the current status of its underlying resources. This RFC proposes how resources can report (and categorize) this information strictly for informational purposes.
```
squeedee:
```suggestion
  - Currently, a blueprint has no knowledge of the current status of its stamped objects. This RFC proposes how resources can report (and categorize) this information strictly for informational purposes.
```

because it's the language I'm using elsewhere. Does not make it right *shrug*.. Could also be `stamped resource` but I tend to think of that as the spec document we apply.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+1. 'Healthy'
+2. 'Unhealthy'
+3. 'Unknown'
+
+It is up to each template author to determine what "healthy" and "unhealthy" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not been able to determine healthy or unhealthy.
+
+Each template will now have a new field in the spec `healthyConditionRule` where authors can specify one of three ways to determine the health of the underlying resource for that template. If no `healthyConditionRule` is defined, Cartographer will default to listing the resource as `Healthy` once it has been successfully applied to the cluster and any relevant outputs have been read off the resource.
+
+The `healthyConditionRule` requires defining one of three fields:
+1. `alwaysHealthy: {}`
```
squeedee: A thought, perhaps
```
forcedCondition: "Healthy" 
```
tells me more, leaves the door open for allowing "Unknown"

rfc/rfc-0000-allow-resources-to-report-status.md
emmjohnson: That is something we could easily add to the spec if we received feedback on it. If we are adding it now, it really becomes 4 states and we are collapsing 2.

rfc/rfc-0000-allow-resources-to-report-status.md
emmjohnson: @squeedee is this still relevant. We are now proposing resources.conditions.

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: I like "stamped resources" since it's the "resources" section of the status, but wouldn't quibble with "stamped objects"

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+apiVersion: carto.run/v1alpha1
+kind: ClusterSourceTemplate
+metadata:
+  name: source
+spec:
+  urlPath: .status.artifact.url
+  revisionPath: .status.artifact.revision
+
+  healthyConditionRule:
+    singleConditionType: Ready
```
scothis: so simple 😄

rfc/rfc-0000-allow-resources-to-report-status.md
squeedee: Not in the slightest

squeedee (approved):

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+
+The RFC proposes modifying the existing templates to allow template authors to indicate how Cartographer can read the state of the resource so that it can be reflected on the owner status.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Currently, a blueprint has no knowledge of the current status of its underlying resources. This RFC proposes how resources can report (and categorize) this information strictly for informational purposes.
+- What use cases does it support?
+  - This feature will enable developers to learn about the progression through the blueprint by looking at the owner status alone. It will also surface information about the health of the underlying resources directly on the owner.
```
scothis: it might be worth calling out that this is augmenting information expoed via resource tracing, and is not a solution for artifact tracing.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+  - This will also enable integrations that may be looking to create a UI around Cartographer, allowing the UI to display the health of the blueprint without having to make an API call to k8s to retrieve the status of every resource, and without having to know how to define health for every stamped object type.
+- What is the expected outcome?
+  - Users will have more insight into the progression of the blueprint and the health of each underlying resource.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC proposes adding a field to the spec of all templates, `healthyConditionRule`. Template authors use field to define the conditions or fields on a stamped resource that determine whether the resource is succeeding or failing.
```
scothis:
```suggestion
This RFC proposes adding a field to the spec of all templates, `healthyConditionRule`. Template authors use field to define the conditions or fields on a stamped resource that determine whether the resource is healthy.
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+
+Cartographer will attempt to read the specified conditions and/or fields and reflect this state on the owner status. If Cartographer cannot determine success or failure, it will show the state as unknown.
```
scothis:
```suggestion
Cartographer will attempt to read the specified conditions and/or fields and reflect this state on the owner status. If Cartographer cannot determine health, it will show the status as unknown.
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+
+```yaml
```
scothis: it would be nice to include a brief preamble for each example that explains what makes it unique. Or add inline comments for the same

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+
+# How it Works
+[how-it-works]: #how-it-works
+
+A stamped resource can be in one of three states:
+1. 'Healthy'
+2. 'Unhealthy'
+3. 'Unknown'
+
+It is up to each template author to determine what "healthy" and "unhealthy" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not been able to determine healthy or unhealthy.
```
scothis:
```suggestion
It is up to each template author to determine what "healthy" and "unhealthy" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not been able to determine health.
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+          lastTransitionTime: "2022-03-23T13:08:22Z"
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+A stamped resource can be in one of three states:
+1. 'Healthy'
+2. 'Unhealthy'
+3. 'Unknown'
```
scothis: This block predates (I think) the switch to use a 'Healthy' condition, it's worth either framing the description as conditions statues or mapping them into condition statuses

```suggestion
1. 'Healthy' (status `True`)
2. 'Unhealthy' (status `False`)
3. 'Unknown' (status `Unknown`)
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+
+It is up to each template author to determine what "healthy" and "unhealthy" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not been able to determine healthy or unhealthy.
+
+Each template will now have a new field in the spec `healthyConditionRule` where authors can specify one of three ways to determine the health of the underlying resource for that template. If no `healthyConditionRule` is defined, Cartographer will default to listing the resource as `Healthy` once it has been successfully applied to the cluster and any relevant outputs have been read off the resource.
+
+The `healthyConditionRule` requires defining one of three fields:
```
scothis: minor: I know the resulting condition is `Healthy (which is good), but I wonder if the field name should be something more like `healthRule`.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+    * The supported matchers are `matchConditions` and `matchFields`. Each is an array, and an author can define either or both of these matchers for a given state.
+    * For `matchConditions`, authors need only define the `type` and `status` of the condition for Cartographer to read. In the owner status, the full condition will be reflected (with the `reason`, `message`, etc.)
+    * For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
+    
+Each owner's status.resources will now include a `conditions` field. To make room for future extensibility, this will be an array of conditions, though this RFC only defines one condition to be included here. The condition will be of `type` 'Healthy' and will report a `status` of 'True', 'False', or 'Unknown'. The condition will also have a `reason` which will be based off the `healthyConditionRule` as follows:
```
scothis: call out that the type of the `conditions` field (e.g. `[]metav1.Condition`)

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+[how-it-works]: #how-it-works
+
+A stamped resource can be in one of three states:
+1. 'Healthy'
+2. 'Unhealthy'
+3. 'Unknown'
+
+It is up to each template author to determine what "healthy" and "unhealthy" mean for the given resource the template is stamping out. "Unknown" strictly represents that Cartographer has not been able to determine healthy or unhealthy.
+
+Each template will now have a new field in the spec `healthyConditionRule` where authors can specify one of three ways to determine the health of the underlying resource for that template. If no `healthyConditionRule` is defined, Cartographer will default to listing the resource as `Healthy` once it has been successfully applied to the cluster and any relevant outputs have been read off the resource.
```
scothis: The alternative is to treat templates without a healthyConditionRule as perpetually unknown. While this would technically be more accurate, if we intend to aggregate the resource Healthy conditions into the owner's Ready condition, that could cause a regression.

We could stage the rollout of this behavior, where the initial pass is to treat the default as unknown, and not aggregate it to the owner Ready. In a later release after templates have has a reasonable opportunity to update, then aggregate the value.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+# Migration
+[migration]: #migration
+
+Given that Cartographer will default to `Healthy` once the resource is applied and outputs are read, this will not be a breaking change.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+- We only allow for reflecting three states, though this design allows for creating more states in the future.
```
scothis: clarify the use of `states` here, it could either be the conditions (of which we define one) or the status of each condition which will always be `True`, `False`, or `Unknown`

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+- This design does not allow authors to define the states themselves.
```
scothis: I view the consistency as a strong positive

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,294 @@
...
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should the two states be `Healthy` and `Unhealthy`?
+- Is it right for Cartographer to default to `Healthy`?
+- Should the owner condition `ResourcesHealthy` have any possible `reason`s other than `HealthyConditionRule`?
```
scothis: remove questions that have been resolved

scothis (approved):
There are a handful of spots that can still be polished

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: I believe this is no longer relevant after the changes made to the RFC.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Adopted this suggestion in most recent updates to the RFC.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: We've decided to move away from the separate CRD in the end.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: I think this is no longer relevant as we have updated the RFC to suggest inlining instead.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: We have moved to using `healthy` true/false/unknown, but this comment is still relevant. Interested to hear what others think about this.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: fixed in latest commit

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Interesting idea. This goes to @cirocosta's comment above. In the multi-stage rollout suggestion, are you suggesting we continue to default to unknown in that later release? That would still be a breaking change so just want to make sure to call that out specifically if that's what you're suggesting.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: added in most recent commit

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Lets move this conversation to the one started below by @scothis

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: To clarify this discussion in light of recent changes the the RFC, if you specified `singleCondition` then if that condition was `Unknown`, you'd still get the message and reason from that condition propagated up to the owner.

So this conversation seems only relevant for the case of `multiMatch`. I think @emmjohnson's suggestion that we later add `unknown` if we receive feedback requesting it makes a lot of sense. Interested to hear what others think, though.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+
+The RFC proposes modifying the existing templates to allow template authors to indicate how Cartographer can read the state of the resource so that it can be reflected on the owner status.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Currently, a blueprint has no knowledge of the current status of the stamped resources. This RFC proposes how resources can report (and categorize) this information strictly for informational purposes.
+- What use cases does it support?
+  - This feature will enable developers to learn about the progression through the blueprint by looking at the owner status alone. It will also surface information about the health of the underlying resources directly on the owner.
```
waciumawanjohi: This certainly surfaces information about the health of the underlying resources directly on the owner.

I wonder if it will truly enable developers to learn about the progression through the blueprint by looking at the owner status alone. Given 1 change (code commit) in flight at a time, that will certainly be the case. Given multiple changes in flight, it's not clear that it will be possible to reason about the status of one change vs another.

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+  - This will also enable integrations that may be looking to create a UI around Cartographer, allowing the UI to display the health of the blueprint without having to make an API call to k8s to retrieve the status of every resource, and without having to know how to define health for every stamped object type.
+- What is the expected outcome?
+  - Users will have more insight into the progression of the blueprint and the health of each underlying resource.
+  - This is augmenting information exposed via resource tracing, and is not a solution for artifact tracing.
+
+# What it is
+[what-it-is]: #what-it-is
+
+This RFC proposes adding a field to the spec of all templates, `healthRule`. Template authors use field to define the conditions or fields on a stamped resource that determine whether the resource is healthy.
```
waciumawanjohi:
```suggestion
This RFC proposes adding a field to the spec of all templates, `healthRule`. Template authors use the field to define the conditions or fields on a stamped resource that determine whether the resource is healthy.
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+    * For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
+    
+Each owner's status.resources will now include a `conditions` field (of type []metav1.condition). To make room for future extensibility, this will be an array of conditions, though this RFC only defines one condition to be included here. The condition will be of `type` 'Healthy' and will report a `status` of 'True', 'False', or 'Unknown'. The condition will also have a `reason` which will be based off the `healthRule` as follows:
+    
+* If `alwaysHealthy: {}`, `reason: AlwaysHealthy`.
+* If `singleConditionType: X`, `reason: XCondition`.
+* If `multiMatch: ...`, `reason: (MatchedCondition|MatchedField)`.
```
waciumawanjohi: Consider the case of a multimatch with a healthy matchCondition and unhealthy matchField. When status is unknown, what would be the reason? (Reason is a required field and may not be an empty string)

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+
+If the `healthRule` is not `alwaysHealthy`, the condition will also have a message. For `singleConditionType`, the message will be propagated from the specified condition. For `multiMatch`, the message will specify the value of the matched condition or field, and then display the message of the matched condition or the message at the path specified in `matchFields`.
```
waciumawanjohi: metav1.Conditions require a message, though it can be empty.

```suggestion
If the `healthRule` is `alwaysHealthy`, the condition message will be a blank string. If the `healthRule` is `singleConditionType`, the message will be propagated from the specified condition. For `multiMatch`, the message will specify the value of the matched condition or field, and then display the message of the matched condition or the message at the path specified in `matchFields`.
```

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+      - It would also mean that templates stamping out the same underlying type would not be able to define different healthy rules, if they desired to.
+- What is the impact of not doing this?
+  - Users/UI creators will have to have in depth knowledge of the underlying resources to determine the health of the owner.
+
+# Prior Art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
```
waciumawanjohi: Does Cartographer need to consider observedGeneration?

Some prior art is the working of the `ClusterDeploymentTemplate`. In that resource, authors of the template define rules to specify when the object has completed work. We specify different manners to ensure that the status we are observing is for the spec that has been submitted. Should users be alerted if the spec and status are out of sync (e.g. spec is generation N while status reports generation N-1)? If they are out of sync in that manner, is the resource still `Healthy`?

waciumawanjohi:
@matt-royal After a workload yields its first built image, would it ever yield later images? E.g. do you need to know if a one-shot workload has succeeded, or do you need to know definition 1 of the workload has succeeded and later to know when definition 2 of the same workload object has succeeded?

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: unhealthy signals are more prominent than unknown signals, unknown signals are more prominent than healthy signals

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: Good question. If we only care to surface the most recent health info we don't have to consider the observed generation. If we want to link that health to the spec, then we do.  Even if we don't care to link the health and specs, if our users want to link them then we should surface it.

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: The functionality proposed is still resource tracing rather than artifact tracing. The health condition surfaced is the latest value reported by the stamped resource, this is identical to how outputs are detected on the stamped resource and passed along the blueprint.

Health is rarely used to identify the path through a system, but more focused on identifying active failure points. One way that Cartographer is distinct from a pipeline is that even when a resource is unhealthy and not producing new outputs, other downstream resources are still available and able to continue processing (they just won't receive the upstream output updates).

waciumawanjohi: I think we're on the same page. To be clear, I'm not advocating for a larger scope here. Rather this suggestion:

```suggestion
  - This feature will surface information about the health of the underlying resources directly on the owner.
```

rfc/rfc-0000-allow-resources-to-report-status.md
waciumawanjohi: This is really the main responsibility of [another RFC](https://github.com/vmware-tanzu/cartographer/pull/799).
Resolving convo.

rfc/rfc-0000-allow-resources-to-report-status.md
martyspiewak: Interesting question. As described currently in the RFC, it would be the first one Cartographer tries to read:
`In the case of multiMatch, if there is more than one matching condition or field that caused the resource to enter it's current state, the first one read will be propagated by Cartographer.`

But perhaps we could have a special reason for Unknown in multiMatch? What do you think?

waciumawanjohi: > But perhaps we could have a special reason for Unknown in multiMatch? What do you think?

I like that idea. Something along the lines of `NoMatchesFulfilled` strikes me as most informative in this case.

martyspiewak: Sounds good to me. Updated the RFC with this suggestion.

waciumawanjohi (approved):

jwntrs (approved):

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+kind: ClusterTemplate
+metadata:
+  name: deploy
+spec:
+  healthRule:
+    multiMatch:
+      healthy:
+        matchConditions:
+          - type: ReconcileSucceeded
+            status: 'True'
```
cirocosta: (nit w/ regards to the example)

it's interesting that a resource like `kapp-controller/App` gives us conditions (which would be useful for either `singleCondition` or `multiMatch` w/ `matchConditions`), but as there's no message/reason under a `ReconcileSucceeed: True` scenario

```yaml
status:
  friendlyDescription: Reconcile succeeded

  conditions:
  - status: "True"              # no message/reason fields
    type: ReconcileSucceeded

  deploy:
    exitCode: 0
    finished: true
    startedAt: "2022-04-12T18:47:21Z"
    stdout: |-
      Target cluster 'https://10.96.0.1:443'
      Changes
      Namespace  Name  Kind     Conds.  Age  Op      Op st.  Wait to    Rs  Ri
      default    app   Service  -       -    create  -       reconcile  -   -
      Op:      1 create, 0 delete, 0 update, 0 noop, 0 exists
      Wait to: 1 reconcile, 0 delete, 0 noop
      6:47:21PM: ---- applying 1 changes [0/1 done] ----
      Pod watching error: Creating Pod watcher: unknown (get pods)
      6:47:21PM: create service/app (serving.knative.dev/v1) namespace: default
      6:47:21PM: ---- waiting on 1 changes [0/1 done] ----
      6:47:21PM: ok: reconcile service/app (serving.knative.dev/v1) namespace: default
      6:47:21PM: ---- applying complete [1/1 done] ----
      6:47:21PM: ---- waiting complete [1/1 done] ----
      Succeeded
...
```

we end up actually being somewhat force to use `multiMatch` with `matchFields` for the `healthy` case too for a complete representation that'd include the `status.friendlyDescription` I believe

rfc/rfc-0000-allow-resources-to-report-status.md
```diff
@@ -0,0 +1,298 @@
...
+    * For `matchConditions`, authors need only define the `type` and `status` of the condition for Cartographer to read. In the owner status, the full condition will be reflected (with the `reason`, `message`, etc.)
+    * For `matchFields`, the spec is based off of `matchFields` in blueprint resource options. Authors need to define the `key` and `operator`, and potentially the `values` (if the operator is `In` or `NotIn`). Authors can also optionally define a `messagePath`, which is a path in the resource where Cartographer can pull off a message to display in the owner status.
+    
+Each owner's status.resources will now include a `conditions` field (of type []metav1.condition). To make room for future extensibility, this will be an array of conditions, though this RFC only defines one condition to be included here. The condition will be of `type` 'Healthy' and will report a `status` of 'True', 'False', or 'Unknown'. The condition will also have a `reason` which will be based off the `healthRule` as follows:
+    
+* If `alwaysHealthy: {}`, `reason: AlwaysHealthy`.
+* If `singleConditionType: X`, `reason: XCondition`.
```
cirocosta: (nit) I wonder if instead we should have `reason: MatchedCondition` 🤔  that way we'd guarantee that regardless of what the rule is, you can depend on that reason if you'd like as it'd be within a bound set of possibilities that we control. wdyt?

cirocosta (approved):
lgtm! left a couple last minute comments, but just nitpicking

rfc/rfc-0000-allow-resources-to-report-status.md
emmjohnson: @scothis what do you  think?

idoru (approved):

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: sounds good

rfc/rfc-0000-allow-resources-to-report-status.md
scothis: `App` will already require the `multiMatch` as it uses independent `Success` and `Failure` conditions (which is a nit I'll pick another day). Even if `App` used a single status condition, I think it's fair to drop down from the `singleCondition` support to pick up the custom message.

`singleCondition` is syntax sugar rather than defining a new capability.

matt-royal:
> @matt-royal After a workload yields its first built image, would it ever yield later images? E.g. do you need to know if a one-shot workload has succeeded, or do you need to know definition 1 of the workload has succeeded and later to know when definition 2 of the same workload object has succeeded?

This is something we need to reason through more, honestly. The most important success/failure for us will be the initial one. Our first design had us creating many "single-use" Workloads, but we've discussed trying to use Cartographer more idiomatically by keeping long-lived Workloads that change over time instead. In this idiomatic case, we'd also need to know the status of resources after they were updated by the most recent Workload change

PR #756, opened 2022-03-28
RFC: Introducing K8s Events

### LINKED ISSUE #711: R4RFC: Publishing events
### Description of problem

Eventing in k8s is an excellent way to surface information to end users. We should use this mechanism to help surface information about whats happening in a supply chain (object stamped, output created, errors, etc). We are requesting an RFC that proposes a solution for (what/when/how/why) events to publish.

Notes:

Info/Data architecture whiteboard: https://miro.com/app/board/uXjVOD4uWt4=/

squeedee:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/rfc-k8s-events/rfc/rfc-0000-k8s-events.md)
Introduce an RFC to describe useful events on Workload, Supply Chain and Runnable

closes #711


rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,170 @@
...
+| ResourceInvalid | `<resource name>` could not be applied as `<group>.<kind>/<name>` due to API server error `<error message>` | This object was (probably) malformed. | Owner |
+| ResourceInvalid | `<resource name>` could not be applied as `<group>.<kind>` due to API server error `<error message>` | This object was (probably) malformed. This lets operators know their templates might have issues | Blueprint |
+| ResourceDoesNotExist | `<resource name>` could not be applied because `<group>.<kind>/<name>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Owner |
+| ResourceDoesNotExist | `<resource name>` could not be applied because `<group>.<kind>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Blueprint |
+| ResourceApplied | `<resource name>` was applied as `<group>.<kind>/<name>` | a resource needed to be created/updated | Owner |
+| ResourceKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template kind changed | Owner |
```
martyspiewak: nit:

as scott pointed out in a previous RFC, this could happen in regular templating too, as the `apiVersion` and `kind` could be templated as well.

squeedee: I missed that, good point.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,170 @@
...
+
+The cartographer controller could emit the following events:
+
+| Reason | Message Format | Description | involvedObject |
+| --- | --- | --- | --- |
+| ResourceExternalSpecChange | an external actor changed the spec of `<group>.<kind>/<name>` | Our cache of the spec, and the spec we just generated match, but the API server has a different one. A lot of these is evidence of thrashing with external resources | Owner |
+| ResourceExternalSpecChange | an external actor changed the spec of resource: `<resource name>` for object `<group>.<kind>` | Our cache of the spec, and the spec we just generated match, but the API server has a different one. A lot of these is evidence of thrashing with external resources | Blueprint |
+| ResourceGetError | `<resource name>` could not retrieve `<group>.<kind>/<name>` due to error: `<error message>` | Loading a resource is failing due to a client.get issue (missing is not an error) | Owner |
+| ResourceGetError | `<resource name>` could not retrieve `<group>.<kind>` due to error: `<error message>` | Loading a resources is failing due to a client.get issue (missing is not an error) - this could be spammy | Blueprint |
```
martyspiewak: I'm curious about this one and why we'd want to expose this event on the blueprint? I'd imagine this would be something like a permissions issue (unless I'm thinking about this wrong). If that is the case, is the thinking that the operator would want to know about this event because they're likely the one supplying the service account the owner is using?

martyspiewak: I realize I'm assuming that the ResourceGetError is that cartographer can't get the applied object from the API server, but maybe that's wrong. Is this supposed to represent that cartographer could not get the template the resource is referring to? If so, what kind of error would this be that would be different from ResourceDoesNotExist?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,170 @@
...
+| ResourceDoesNotExist | `<resource name>` could not be applied because `<group>.<kind>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Blueprint |
+| ResourceApplied | `<resource name>` was applied as `<group>.<kind>/<name>` | a resource needed to be created/updated | Owner |
+| ResourceKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template GVK changed, or templated values in GVK changed | Owner |
+| ResourceOutputChanged | `<resource name>` found a new output in `<group>.<kind>/<name>` | a resource produced a new output | Owner |
+| ResourceHealthyStatusChanged | `<resource name>` found a new status in `<group>.<kind>/<name>` | a resource produced a new healthy status | Owner |
+| SupplyChainChanged | supply chain changed from `<old supply chain name>` to `<new supply chain name>` | Workload selected for a new or different supply chain. Note: `none` is a possible name. | Owner |
+| WorkloadSelected | supply chain matched a workload | A workload was selected for. This makes the count a useful metric | Blueprint |
```
martyspiewak: I'd suggest we only emit this event when a supply chain matches a workload for the first time. I wouldn't want to emit this event every time we enter a reconcile for every workload.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,170 @@
...
+
+* Cache misses? (Not sure about this, we'll miss if the controller restarts or with HA Elections (future concern))
```
martyspiewak: I think an event for a cache miss would be at worst misleading (because it could look like an issue to a developer) and at best irritating (as the developer might see a bunch of cache misses clogging up the events). I'm not sure it's really useful information for the developer- if we restarted and had a cache miss, the developer probably won't care, and if we had a miss because the object changed, we'll emit a ResourceApplied anyway.

rfc/rfc-0000-k8s-events.md
squeedee: No, I was thinking of the most highly visible and problematic partitioning errors are API errors. There are probably far better ways of discovering this about your cluster so the merit here is pretty weak.

rfc/rfc-0000-k8s-events.md
squeedee: Yeah multiple people have proven to me that assuming that was obviously implied was a bad idea. I'll change the description.

rfc/rfc-0000-k8s-events.md
squeedee: I wasn't the one to suggest this, so the reasoning entirely eludes me. I agree with you but I'll leave it as a question a while longer to see if anyone has a case for it.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+| StampedObjectInvalid | `<resource name>` could not be applied as `<group>.<kind>/<name>` due to API server error `<error message>` | This object was (probably) malformed. | Owner |
+| StampedObjectInvalid | `<resource name>` could not be applied as `<group>.<kind>` due to API server error `<error message>` | This object was (probably) malformed. This lets operators know their templates might have issues | Blueprint |
+| StampedObjectKindNotFound | `<resource name>` could not be applied because `<group>.<kind>/<name>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Owner |
+| StampedObjectKindNotFound | `<resource name>` could not be applied because `<group>.<kind>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Blueprint |
+| StampedObjectApplied | `<resource name>` was applied as `<group>.<kind>/<name>` | a stamped object needed to be created/updated | Owner |
```
emmjohnson: Do we want to distinguish between created and updated?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+| StampedObjectGarbageCollected | `<group>.<kind>/<name>` is no longer referenced | This owner has selected a different template, either by supply chain selection or templating, and this object is no longer needed | Owner |
+| ImmutableStampedObjectGarbageCollected | `<n> * <group>.<kind>` historical objects deleted due to garbage collection policy | This runnable's GC policy has caused `n` objects to be removed | Runnable |
+| StampedObjectInvalid | `<resource name>` could not be applied as `<group>.<kind>/<name>` due to API server error `<error message>` | This object was (probably) malformed. | Owner |
+| StampedObjectInvalid | `<resource name>` could not be applied as `<group>.<kind>` due to API server error `<error message>` | This object was (probably) malformed. This lets operators know their templates might have issues | Blueprint |
+| StampedObjectKindNotFound | `<resource name>` could not be applied because `<group>.<kind>/<name>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Owner |
```
emmjohnson: Can we distinguish this error?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+
+The situation is worse for users managing the supply chains (Devops/Ops). By looking at a supply chain, there is very
+little information about how supply chains are behaving, or if any issues are occurring for developers:
+
+![Blueprint](./rfc-k8s-events/blueprint.jpg)
+
+Although these users can access the logs, we can provide them a more concise view of how supply chains are behaving
+on-cluster.
+
+We add events to the Blueprints (Supply Chain/Delivery) to enable debug and reasoning about Cartographer's behavior.
```
emmjohnson: Won't this be a lot of events on the blueprint? Potentially thousands (or tens of thousands) Is this helpful?

rfc/rfc-0000-k8s-events.md
squeedee: Personally I think no. Compare separate:

```
Normal    StampedObjectCreated      1h (x1 over 1h)          workload-controller      `source-controller` was created as `sourcerepository.something/petclinic`
Normal    StampedObjectChanged     1h (x102 over 1h)     workload-controller      `source-controller` was updated as `sourcerepository.something/petclinic`
```

And combined:

```
Normal    StampedObjectApplied.      1h (x103 over 1h)     workload-controller      `source-controller` was applied as `sourcerepository.something/petclinic`
```

Essentially when separate, the creation event will be a singleton (except in rare cases when it was deleted, which is covered by two other events, one for deliberately by us, and the other for "spec change"). So we end up with a singleton event for all the resources, and then the combined applied events. All the singleton tells us is when it was created, which the `age` can do equally well. And if you dive into the event, it will tell you when the first event was pubaished _precisely_, so you can inspect the combined event and discover those details without needing a seperate event.

rfc/rfc-0000-k8s-events.md
squeedee: I dont know but I expect we can, before doing feasibility however, I want to know if we "want it" from a UX perspective. Pretty sure the api errors tell us why we got a non-200 response tho?

rfc/rfc-0000-k8s-events.md
squeedee: What's the concern? Calls to the API, or do you perceive there being thousands of distinct events? (recall that the server aggregates like events into a single event object with a count). 

As far as API calls go, I defer to someone like @scothis who might know what the compromise is for events.

squeedee:
Do i need to limit the list of events to the few we definitely want, drop the rest into an addendum, so this can be approved?

emmjohnson (approved):

rfc/rfc-0000-k8s-events.md
scothis: A cache miss is expected and not user facing. It's worth of being logged, but not an event.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+| StampedObjectKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template GVK changed, or templated values in GVK changed | Owner |
+| ResourceOutputChanged | `<resource name>` found a new output in `<group>.<kind>/<name>` | a resource produced a new output | Owner |
+| ResourceHealthyStatusChanged | `<resource name>` found a new status in `<group>.<kind>/<name>` | a resource produced a new healthy status | Owner |
+| SupplyChainChanged | supply chain changed from `<old supply chain name>` to `<new supply chain name>` | Workload selected for a new or different supply chain. Note: `none` is a possible name. | Owner |
+| WorkloadSelected | supply chain matched a workload | A workload was selected for (really, this is when the supplyChainRef changes). This makes the count a useful metric | Blueprint |
+
+* Cache misses? (Not sure about this, we'll miss if the controller restarts or with HA Elections (future concern))
+
+**Note:** Blueprint event `messages` are usually designed to aggregate on the resource kind, not the individual
+resource (we omit `<resource name>`)
```
scothis: I'm not sold that most of the blueprint events as defined are useful. In order to be understood, the context provided by the owner resource is needed. The owner information (a namespaces resource) likely shouldn't be recorded against the blueprint resource (a cluster scoped resource).

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+key guiding principles:
+
+1. > Let it be something that isn't satisfactorily presented in other user views (eg: `k get workload`).
+2. > Let the event list be as meaningful as is necessary, and no more.
+
+The cartographer controller could emit the following events:
+
+| Reason | Message Format | Description | involvedObject |
+| --- | --- | --- | --- |
+| StampedObjectExternalSpecChange | an external actor changed the spec of `<group>.<kind>/<name>` | Our cache of the spec, and the spec we just generated match, but the API server has a different one. A lot of these is evidence of thrashing with external resources | Owner |
```
scothis: nit: `<group>.<kind>/<name>` isn't a common form. `<resource>.<group>/<name>` is more meaningful. "resource" here is the k8s api definition of resource, which is the plural, lowercase form of "kind".

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+This first diagram shows the objects tracked or managed by the Cartographer controller, except for **Artifact** and
+**Controller Stamp**.
+
+![Cartographer Objects](./rfc-k8s-events/objects.jpg)
+
+## Developer View
+
+To date, Cartographer presents at least some information from all these sources to developers:
+
+![Owner](./rfc-k8s-events/owner.jpg)
```
scothis: To my eyes this diagram is very noisy and distracts from the intent of this RFC more than it illuminates the domain.

rfc/rfc-0000-k8s-events.md
squeedee: > _The owner information (a namespaces resource) likely shouldn't be recorded against the blueprint resource (a cluster scoped resource)._

Agreed, so we do not.

The User Journey I have in mind when adding these:

* I am an operator.
* A developer approaches me with an issue in their workload
* I see the issue, and it can be fixed by changing some configuration
* I wonder who else is affected by this kind of issue, and look to the supply chain for 'all workload' aggregates of those kinds of events.

I am in no rush to introduce the blueprint events, so could put them in an addenda for now, as a possibility for Operator facing UX in the future.

rfc/rfc-0000-k8s-events.md
squeedee: I understand that, conversely I struggle understanding communications from people who can easily relate objects in a system without graphical/spatial representations. We do have plans to have an IA diagram that's publicly accessible, once done I can just drop a link to those diagrams for those of us who need spatial and keyed references.

rfc/rfc-0000-k8s-events.md
squeedee: Oh I lie, we do name the resource. An oversight on my behalf, I don't want events stacking up for each resource instance. Will fix and commit some changes.

rfc/rfc-0000-k8s-events.md
squeedee: Yep, thats what I meant, just a collision of dyslexia and not knowing what to call the lowercase/plural kind.  Will fix

rfc/rfc-0000-k8s-events.md
squeedee: FYI this is an overload issue. `<resource name>` is not an instance name (`metadata.name`) but the name in the supply chain (`spec.resources[].name`)

rfc/rfc-0000-k8s-events.md
squeedee: Should be better now.

rfc/rfc-0000-k8s-events.md
scothis: The diagram is good for general purpose documentation. For an RFC I'd like to be able to focus on the parts of the resource that are new/updated and other parts of the model that are directly relate to the change. Context is key.

Another way I think about it: RFCs hopefully have a long lifespan. Other aspects of the resources that any particular RFC modify will likely themselves be modified down the road. Any information not directly related to the current RFC may grow stale and be more confusing than helpful in the future. Examples in a RFC are obviously of the moment.

None of this is right vs wrong, it's a subjective balance.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,176 @@
...
+
+[how-it-works]: #how-it-works
+
+To avoid over-communicating, and making the view of event's overwhelming and dilute for user's we will start with two
+key guiding principles:
+
+1. > Let it be something that isn't satisfactorily presented in other user views (eg: `k get workload`).
+2. > Let the event list be as meaningful as is necessary, and no more.
+
+The cartographer controller could emit the following events:
```
scothis: Is this RFC proposing that these exact events exist, or that there will be events and here are some types of events that we think are interesting? It would be good to clarify that choice in the summary.

I'd tend to lean towards making these exemplar events and leaving the details a bit more up to the implementation. Minimally, allowing room for the message to evolve can be helpful.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,177 @@
...
+| StampedObjectApplied | `<resource name>` was applied as `<resource>.<group>/<name>` | a stamped object needed to be created/updated | Owner |
+| StampedObjectKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template GVK changed, or templated values in GVK changed | Owner |
+| ResourceOutputChanged | `<resource name>` found a new output in `<resource>.<group>/<name>` | a resource produced a new output | Owner |
+| ResourceHealthyStatusChanged | `<resource name>` found a new status in `<resource>.<group>/<name>` | a resource produced a new healthy status | Owner |
+| SupplyChainChanged | supply chain changed from `<old supply chain name>` to `<new supply chain name>` | Workload selected for a new or different supply chain. Note: `none` is a possible name. | Owner |
+| WorkloadSelected | supply chain matched a workload | A workload was selected for (really, this is when the supplyChainRef changes). This makes the count a useful metric | Blueprint |
+
+**Note:** `<resource name>` represents the `blueprint.spec.resource[].name` and *not* `stampedObject.metadata.name`
```
scothis: nit: the path is on a resource, the resource is not part of the path.

```suggestion
**Note:** `<resource name>` represents the blueprint's `.spec.resource[].name` and *not* the stamped object's `.metadata.name`
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,177 @@
...
+  ----    ------                      ----                   ----                 -------
+Normal    ResourceExternalSpecChange  1m59s (x2 over 1m59s)  workload-controller  an external actor changed the spec of kpack.io.image/java-sample-app
+Normal    ResourceExternalSpecChange  35s (x4 over 22s)      workload-controller  an external actor changed the spec of kappctrl.k14s.io/java-sample-app
+```
+
+# Migration
+
+[migration]: #migration
+
+THis RFC only adds to the API, there are no breaking changes that require migration.
```
scothis:
```suggestion
This RFC only adds to the API, there are no breaking changes that require migration.
```

rfc/rfc-0000-k8s-events.md
squeedee: They are just examples, however I think we do need to use these examples to drive home and document some design guidelines (within this RFC) in the hope of not creating a proliferation of meaningless or confusing events.

rfc/rfc-0000-k8s-events.md
squeedee: I tried to address your concerns in this commit. https://github.com/vmware-tanzu/cartographer/pull/756/commits/ea5bf737a6385fede7b17053ed540bd9f1afbde5

rfc/rfc-0000-k8s-events.md
squeedee: @scothis Did you miss this: https://github.com/vmware-tanzu/cartographer/blob/907cd2d949486463555c6908799a4048b9310e8a/rfc/rfc-0000-k8s-events.md?plain=1#L91-L92

Or is that not what you mean? Or is it perhaps not obvious enough?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+| StampedObjectApplied | `<resource name>` was applied as `<resource>.<group>/<name>` | a stamped object needed to be created/updated | Owner |
+| StampedObjectKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template GVK changed, or templated values in GVK changed | Owner |
+| ResourceOutputChanged | `<resource name>` found a new output in `<resource>.<group>/<name>` | a resource produced a new output | Owner |
+| ResourceHealthyStatusChanged | `<resource name>` found a new status in `<resource>.<group>/<name>` | a resource produced a new healthy status | Owner |
+| SupplyChainChanged | supply chain changed from `<old supply chain name>` to `<new supply chain name>` | Workload selected for a new or different supply chain. Note: `none` is a possible name. | Owner |
+| WorkloadSelected | supply chain matched a workload | A workload was selected for (really, this is when the supplyChainRef changes). This makes the count a useful metric | Blueprint |
```
sclevine: Could we move this to the Owner so that we don't have so many events associated with the Blueprint?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+key guiding principles:
+
+1. > Let it be something that isn't satisfactorily presented in other user views (eg: `k get workload`).
+2. > Let the event list be as meaningful as is necessary, and no more.
+
+These are examples of types of events that could be emitted. This RFC makes no recommendation as to which events are
+required. The cartographer controller could emit the following events:
+
+| Reason | Message Format | Description | involvedObject |
+| --- | --- | --- | --- |
```
scothis: Double quotes are more common in events than backticks for templated values.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+To avoid over-communicating, and making the view of event's overwhelming and dilute for user's we will start with two
+key guiding principles:
+
+1. > Let it be something that isn't satisfactorily presented in other user views (eg: `k get workload`).
+2. > Let the event list be as meaningful as is necessary, and no more.
+
+These are examples of types of events that could be emitted. This RFC makes no recommendation as to which events are
+required. The cartographer controller could emit the following events:
+
+| Reason | Message Format | Description | involvedObject |
```
scothis: Events can have a type of either `Normal` or `Warning`. It may be worth adding that as a column in the table.

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+| StampedObjectObjectGetError | `<resource name>` could not retrieve `<resource>.<group>` due to error: `<error message>` | Loading a resources is failing due to a client.get issue (missing is not an error) - this could be spammy | Blueprint |
+| StampedObjectGarbageCollected | `<resource>.<group>/<name>` is no longer referenced | This owner has selected a different template, either by supply chain selection or templating, and this object is no longer needed | Owner |
+| ImmutableStampedObjectGarbageCollected | `<n> * <resource>.<group>` historical objects deleted due to garbage collection policy | This runnable's GC policy has caused `n` objects to be removed | Runnable |
+| StampedObjectInvalid | `<resource name>` could not be applied as `<resource>.<group>/<name>` due to API server error `<error message>` | This object was (probably) malformed. | Owner |
+| StampedObjectInvalid | `<resource name>` could not be applied as `<resource>.<group>` due to API server error `<error message>` | This object was (probably) malformed. This lets operators know their templates might have issues | Blueprint |
```
scothis: Is this intended as a catch-all error from the api sever, or to specifically handle [invalid](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsInvalid) requests?

Are runtime errors from the templating phase captured?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+| StampedObjectKindNotFound | `<resource name>` could not be applied because `<resource>.<group>/<name>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Owner |
```
scothis:
```suggestion
| StampedObjectKindNotFound | `<resource name>` could not be applied because `<resource>.<group>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Owner |
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+| StampedObjectKindNotFound | `<resource name>` could not be applied because `<resource>.<group>` does not exist on this cluster | Did someone forget to install the CRDs? Otherwise it's a malformed template | Blueprint |
+| StampedObjectApplied | `<resource name>` was applied as `<resource>.<group>/<name>` | a stamped object needed to be created/updated | Owner |
+| StampedObjectKindChanged | `<resource name>` was `<old group>.<old kind>`, now `<new group>.<new kind>` | YTT selection for a template GVK changed, or templated values in GVK changed | Owner |
+| ResourceOutputChanged | `<resource name>` found a new output in `<resource>.<group>/<name>` | a resource produced a new output | Owner |
+| ResourceHealthyStatusChanged | `<resource name>` found a new status in `<resource>.<group>/<name>` | a resource produced a new healthy status | Owner |
```
scothis: it's worth capturing the status (`True`, `False`, `Unknown`) of the healthy condition

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+| SupplyChainChanged | supply chain changed from `<old supply chain name>` to `<new supply chain name>` | Workload selected for a new or different supply chain. Note: `none` is a possible name. | Owner |
```
scothis: Is there a deliverable equivalent, or can we make this more generic between the two?

scothis (approved):

jwntrs (approved):

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+
+| Reason | Message Format | Description | involvedObject |
+| --- | --- | --- | --- |
+| StampedObjectExternalSpecChange | an external actor changed the spec of `<resource>.<group>/<name>` | Our cache of the spec, and the spec we just generated match, but the API server has a different one. A lot of these is evidence of thrashing with external resources | Owner |
+| StampedObjectExternalSpecChange | an external actor changed the spec of resource: `<resource name>` for object `<resource>.<group>` | Our cache of the spec, and the spec we just generated match, but the API server has a different one. A lot of these is evidence of thrashing with external resources | Blueprint |
+| StampedObjectGetError | `<resource name>` could not retrieve `<resource>.<group>/<name>` due to error: `<error message>` | Loading a resource is failing due to a client.get issue (missing is not an error) | Owner |
+| StampedObjectObjectGetError | `<resource name>` could not retrieve `<resource>.<group>` due to error: `<error message>` | Loading a resources is failing due to a client.get issue (missing is not an error) - this could be spammy | Blueprint |
```
waciumawanjohi: `StampedObjectGetError` vs `StampedObjectObjectGetError` Is this a typo? Are these meant to be the same event, just associated with different involvedObjects?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+to provide user's awareness of state-changing event's that are meaningful for debugging and situational awareness.
+
+# Motivation
+
+[motivation]: #motivation
+
+Developers and Operators need to debug issues where workloads stall, do not make it to deployment, or where other level
+changes do not propagate as they should.
+
+Cartographer concern's itself with creating **Stamped Objects** from **Templates** and passing **Artifact**
```
waciumawanjohi:
```suggestion
Cartographer concerns itself with creating **Stamped Objects** from **Templates** and passing **Artifact**
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+
+[what-it-is]: #what-it-is
+
+## Developer View
+
+Cartographer presents useful information to developers in the **owner** (workload/deliverable). This information is
+derived from stamped objects and contributing resources such as **templates** and **blueprints**:
+
+This RFC addresses the absence of awareness that "something is happening", especially after a developer has committed
+new code. With event's we can provide information that show's activity and causality not present in the current snapshot
```
waciumawanjohi:
```suggestion
new code. With events we can provide information that shows activity and causality not present in the current snapshot
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+of state.
+
+We add events to the **Owners** (Workload/Deliverable) to enable debug and reasoning about Cartographer's behavior.
+
+## Operator View
+
+The situation is worse for users managing the **blueprints** (Devops/Ops). Some metrics can be recorded in **
+blueprints**
+and this might be a valid topic for another RFC, however events also present a useful way for authors to avoid trawling
```
cirocosta: indeed, one could certainly make the argument that an ops team that makes use of any solution that converts those in-cluster events to data points to derive insights in some form could make use of these owner-based events for monitoring in some capacity

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+
+[summary]: #summary
+
+Introduce the use
+of [K8s Events](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events)
+to provide user's awareness of state-changing event's that are meaningful for debugging and situational awareness.
```
waciumawanjohi:
```suggestion
to provide users awareness of state-changing events that are meaningful for debugging and situational awareness.
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+and this might be a valid topic for another RFC, however events also present a useful way for authors to avoid trawling
+through logs to get a quick overview of temporal behaviour of **owners** (workload/delivery) as they rely on the
+author's
+**blueprint**.
+
+# How it Works
+
+[how-it-works]: #how-it-works
+
+To avoid over-communicating, and making the view of event's overwhelming and dilute for user's we will start with two
```
waciumawanjohi:
```suggestion
To avoid over-communicating, and making the view of events overwhelming and dilute for users we will start with two
```

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Note:** Blueprint event `messages` are usually designed to aggregate on the resource kind, not the individual
+resource (we omit `<resource name>`)
+
+**Note:** On naming, we're tring to start with the location in the object diagram (first diagram) that represents the
+change. `stamped object` is not well known to users, but we need to rectify that, or use a clearer term to describe "the
+resource on etcd", which would normally be "Resource". Unfortunately we've overloaded resource to mean the collective
+concept of a template, a configuration, it's submission to the API and it's resulting changes from external events.
+
+There are a lot of event's here, and more could exist, but we should review these during the RFC review process. I hope
```
waciumawanjohi:
```suggestion
concept of a template, a configuration, its submission to the API and its resulting changes from external events.

There are a lot of events here, and more could exist, but we should review these during the RFC review process. I hope
```

rfc/rfc-0000-k8s-events.md
cirocosta: > _I am in no rush to introduce the blueprint events,_

I'd +1 reducing the scope of this RFC to just owner-related events. wdyt, @sclevine @rawlingsj ?

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+[prior-art]: #prior-art
+
+See k8s native Pods etc.
+
+# Unresolved Questions
+
+[unresolved-questions]: #unresolved-questions
+
+- How many of these events do we really need
+- How do we avoid an exhaustive and hard to parse list of events.
```
cirocosta: @squeedee was the concern here that we might end up with too many events? wdym by hard to parse?

rfc/rfc-0000-k8s-events.md
cirocosta: Perhaps something along the lines of "BlueprintChanged"? indeed, we'd need to make sure that these are generic enough such that it'd work well with both blueprints and owner types (+ keeping in mind of [blueprint rfc](https://github.com/vmware-tanzu/cartographer/pull/766))

rfc/rfc-0000-k8s-events.md
```diff
@@ -0,0 +1,162 @@
...
+**Note:** Blueprint event `messages` are usually designed to aggregate on the resource kind, not the individual
+resource (we omit `<resource name>`)
+
+**Note:** On naming, we're tring to start with the location in the object diagram (first diagram) that represents the
+change. `stamped object` is not well known to users, but we need to rectify that, or use a clearer term to describe "the
+resource on etcd", which would normally be "Resource". Unfortunately we've overloaded resource to mean the collective
+concept of a template, a configuration, it's submission to the API and it's resulting changes from external events.
+
+There are a lot of event's here, and more could exist, but we should review these during the RFC review process. I hope
+we can prune the list and make it `as meaningful as is necessary, and no more`
```
waciumawanjohi: In Office Hours, there was a discussion that this RFC _not_ define an explicit list of events. Rather that the RFC express a general guideline for events and that the team agree that future RFCs will not be necessary to change the list of event types. Is that a suggestion that is being adopted by this RFC?

waciumawanjohi (approved):

cirocosta (approved):
left a couple nits, lgtm though!

rfc/rfc-0000-k8s-events.md
squeedee: That's SupplyChainChainged. This event(WorkloadSelected) was so that an operator could get temporal stats about their supply chain usage. I personally think this would be valuable and don't understand the general concern I see in comments about blueprint events?

idoru (approved):

rfc/rfc-0000-k8s-events.md
squeedee: definitely a typo

rfc/rfc-0000-k8s-events.md
squeedee: It's pretty commonplace for admins to use monitoring tools with nice temporal graphs showing histograms that highlight usage, network partitioning events, heavy load events, etc.

rfc/rfc-0000-k8s-events.md
squeedee: Too many event types. mentally if i see a long list of low-interest event's in a `describe` I'm much more likely to miss the one or two truly important ones, such as when we're constantly submitting a resource to the API because we're fighting with the resource's own controller mutating the `.spec`

rfc/rfc-0000-k8s-events.md
squeedee: We've reworded this as a list of examples.

rfc/rfc-0000-k8s-events.md
squeedee: The scope of this RFC is to introduce events and to establish some guidelines. Not to force a list of events.

OTOH, I really believe in the value of bluepring scoped events, _especially_ the BlueprintChanged event.

rfc/rfc-0000-k8s-events.md
squeedee: Yeah I did that for consistancy in markdown, not the final events. I'll patch it.

squeedee: See what you made me do: `"\<resource>.\<group>/\<name>"` :P

sclevine (approved):

rfc/rfc-0000-k8s-events.md
Meticu: hai

Meticu (commented):
rfc/rfc-k8s-events/objects.jpg

Issue #148, opened 2021-09-23
Provide guidance on system requirements


cirocosta:
_"I want to be have enough guidance on how to allocate resources for cartographer"_

```gherkin
Given I'm about to install Cartographer
When I finish read the installation guide
Then I feel confident about allocating resources for my system
```

## context

currently, when installing the controller via the configuration that we provide (see ./config/manager), a certain set of memory and cpu get limitted & reserved:

https://github.com/vmware-tanzu/cartographer/blob/d4c3a11548cb4084aedda0453abeebfbcf1c3d67/config/manager/deployment.yaml#L67-L73

while those have been "just fine" based on "that's probably enough", we should aim at least at providing values based on a certain benchmark that we set, as well as provide some clues on how one should go about allocating resources for their system, e.g., making very clear that cartographer is just choreographing other components and that the result of those   other ones acting on changes should be considered.

##

acceptance criteria:

- [ ] documentation updated with sizing guidance
- [ ] a benchmark setup
- [ ] results from running the benchmark posted as a chart of memory and cpu usage as # of workloads scales

squeedee:
Propose:

Submit many workloads, that also trigger many runnables.
Profile with a supposed linear scale (50,100,150... jobs) and see how linearly mem and cpu behave.
Capture for future reference.

## Notes:
* Can use configs or something that does not use3rd party controllers.

cirocosta:
done @ #762 #763

PR #762, opened 2022-03-29
Introduce benchmark test

idoru:

Add a benchmark test that can be run repeatedly against a cluster with Cartographer installed and measure CPU and Memory consumption.

For a quick test, install a Cartographer release and metrics-server [1] on a cluster, make it your current kubectl context and then run:
```
CARTO_BENCH_N_SUPPLY_CHAINS=1 CARTO_BENCH_N_WORKLOADS=1 CARTO_BENCH_N_RUNNABLES=2 CARTO_BENCH_N_RUNNABLE_RUNS=3 go run -modfile hack/tools/go.mod github.com/onsi/ginkgo/ginkgo -r benchmark
```

For something bigger, see `./hack/benchmark.sh`

provides tooling around to support #148

 - [1] https://github.com/kubernetes-sigs/metrics-server#installation

## PR Checklist


tests/performance/performance_suite_test.go
```diff
@@ -0,0 +1,63 @@
...
+package performance_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
+	"github.com/vmware-tanzu/cartographer/tests/resources"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" //FIXME:seriously, for GCP to work? nb: plugins for azure/aws probably needed on those iaas
```
emmjohnson: Might as well add it for all the IAASes

tests/performance/performance_suite_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2021 VMware
```
emmjohnson: Do we want this to live under tests? This isn't something we are planning on running regularly. Maybe it can be its on go module somewhere?

tests/performance/performance_suite_test.go
scothis:
```suggestion
	_ "k8s.io/client-go/plugin/pkg/client/auth"
```

tests/performance/performance_suite_test.go
idoru: Sure. Any objection to moving it to top level package `benchmark`?

martyspiewak: When you say a top level package, do you mean under `/pkg` or at the same level as `/pkg`? If the latter, I'd be ok with that if it's its own go module.

martyspiewak (approved):

PR #766, opened 2022-03-30
RFC: Common Blueprint Architecture

### LINKED ISSUE #731: R4RFC: Common blueprint architecture
### Description of problem

Requesting an RFC that describes how we could merge `SupplyChains`, `Deliveries` and all the template types into a set of typed `Blueprints` (`ClusterSourceBlueprint`, `ClusterImageBlueprint`, `ClusterConfigBlueprint`, `ClusterBlueprint`) in order to simplify our data model for users.

Note: moving towards this architecture implies we will support composability in blueprints.

emmjohnson:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/731-rrfc-common-blueprint-architecture/rfc/rfc-0000-common-blueprint.md)

Closes: #731


sclevine:
Great to see this coming together!

Feedback so far:

The original motivation behind the typed templates was allowing the templates to be safely swapped. E.g., you can always swap a kpack Image template for a tekton/kaniko template because the output is always an image ref in a well-defined format.

Could we bring back types, but not the typed templates? Maybe something closer to strong dynamic types, instead of strong static types?

martyspiewak:
> Maybe something closer to strong dynamic types, instead of strong static types

@sclevine Are you suggesting something along the lines of Cartographer validating that a blueprint supplies the output specified as an input by a downstream resource? If so, I think that makes a lot of sense and we should definitely add that to this RFC.

sclevine:
Apologies for the delay! Don't hesitate to bug me on Slack if you're waiting on me 🙂 

> Are you suggesting something along the lines of Cartographer validating that a blueprint supplies the output specified as an input by a downstream resource?

That also seems important, but it's not quite what I'm suggesting.

We currently require outputs to be typed (source, image, config, or void). This allows any, e.g., ClusterImageTemplate to be swapped with another ClusterImageTemplate without breaking the SupplyChain in unexpected ways. Without typed outputs, templates could pick slightly different names (e.g., `image`, `img`, `imageDigest`) for the same output, or output the same data in different formats (e.g., `my-repo/my-image:latest` vs. `my-repo/my-image@sha256:...`). 

I like the idea of not requiring a separate CRD for each type. However, the type information still needs to be preserved. For example, the template could have a `type` field for each output. The type could be registered via a `ClusterBlueprintType` CR, which defines the structure (and in the future, possibly the runtime behavior) of the type.

martyspiewak:
@sclevine That makes a lot of sense. Updated the RFC to include your suggestions.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+    - name: first-resource
+      inputs:
+        - resource: zero-resource
+          output: zoo
+          name: zoo-resource-name
+      inline:
+        # mutually exclusive - template, ytt
+        template: {}
+        ytt: {}
+
```
cirocosta: just to be sure: the proposal is that `resources` should or should _not_ be one of the possibilities here? I imagine the answer is _not_?

i.e., the following would be invalid:

```yaml
  resources:
    - name: first-resource
      inline:
        resources:
          - name: first-first-resource
             template: {}
          - name: first-second-resource
             ytt: {}
```

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+spec:
+  blueprintTypeRef:
+    name: my-baz-fizz-type
+  
+  # mutually exclusive - resources, template, ytt 
+  resources:
+    - name: first-resource
+      inputs:
+        - resource: zero-resource
+          output: zoo
```
cirocosta: what is this field used for? e.g., when templating something with `ytt`, what's the influence of this? 

e.g., previously when consuming some sources, e.g., 

```
sources: [{name: git-repository, resource: source-provider}]
```

we'd be able to do in ytt (for grabbing the url): 

- `#@ data.values.source.url`, or 
- `#@ data.values.sources['git-repository'].url`

how does this new format compare?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+- What is the expected outcome?
+
+  - Simplified on-boarding and user experience.
+
+# What it is
+[what-it-is]: #what-it-is
+* `ClusterBlueprints` will encapsulate everything the current templates allow, while adding `spec.resources` to enable defining multiple resources (i.e. snippets).
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
+* A `ClusterBlueprintType` defines only `spec.outputs`, which is an array of output names. By leaving the `name` key in the array, we leave open the possibility to later add `type` to each output, though this RFC currently proposes only supporting string outputs.
```
cirocosta: > _though this RFC currently proposes only supporting **string** outputs._

in the sense that any `valuePath` that evaluates to anything other than a string wouldn't be considered valid? I think that'd be a big regression as, e.g., those leveraging today `ClusterConfigTemplate` with `configPath` pointing at a map structure like a podtemplatespec would not be able to do so anymore

cirocosta: is the intention behind a separate CRD here **solely** to allow extending the current types we have (`source`, `image`, and `config`) so that a blueprint could be tailored towards other use-cases than the ones Cartographer currently is built towards?

 I'm wondering if this is a field that should actually _not_ be a ref and instead an enum (ala `type: Source`, `type: Image`, `type: Config`, `type: None` (?))

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+- Start Date: 2022-03-29
+- Author(s): @emmjohnson @martyspiewak
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes collapsing all current "Blueprints" and templates (not including `ClusterRunTemplate`) into a common type, `ClusterBlueprint`.
```
cirocosta: maybe worth calling out **why** not clusterruntemplate? I imagine for most readers not super familiar with it, this might look confusing

perhaps something in the lines of

```suggestion
This RFC proposes collapsing all current "Blueprints" and templates (not including `ClusterRunTemplate`, as that resource is in the domain of [Runnable](https://cartographer.sh/docs/v0.3.0/reference/runnable/)) into a common type, `ClusterBlueprint`. 
```

?

rfc/rfc-0000-common-blueprint.md
emmjohnson: configPath is of type `string`  today - https://github.com/vmware-tanzu/cartographer/blob/120a53cd8577364610563969b27183f83242a4ae/config/crd/bases/carto.run_clusterconfigtemplates.yaml#L58

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+
+- What is the expected outcome?
+
+  - Simplified on-boarding and user experience.
+
+# What it is
+[what-it-is]: #what-it-is
+* `ClusterBlueprints` will encapsulate everything the current templates allow, while adding `spec.resources` to enable defining multiple resources (i.e. snippets).
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
```
cirocosta: today we have those four CRDs

- Cluster**Source**Template, for exposing **source** info (`{url: <string>, revision: <string>}`[^1])
- Cluster**Image**Template, for exposing **image** info (`{image: <string>}`)
- Cluster**Config**Template, for exposing **config** info (`{config: <map[string]interface{}}}`)
- *ClusterTemplate, not exposing anything* 

So, I'd imagine that those first three would look like:

```yaml
---
kind: ClusterBlueprintType
metadata:
  name: source
spec:
  outputs:
    - name: url
    - name: revision

---
kind: ClusterBlueprintType
metadata:
  name: image
spec:
  outputs:
    - name: image

---
kind: ClusterBlueprintType
metadata:
  name: config
spec:
  outputs:
    - name: config
```

but then the fourth (`ClusterTemplate`), how would that look like? 

```yaml
kind: ClusterBlueprintType
metadata:
  name: no-outputs
spec:
  outputs: []
```
?

or is that idea that such kind of blueprint shouldn't even exist in the first place as all blueprints should expose _something_?

---
[^1]: _(in the comment above ^ I added some types for the fields, but, those are not really enforced as of today)_

cirocosta:
I was playing with the idea of creating a blueprint for `gitrepositories.source.. -> images.kpack --> apps.kapp`: https://gist.github.com/cirocosta/affe3ec8c21560a302db1dc2e69c5ae7

While going through it, a couple raw thoughts

- it feels strange to have to create a `BlueprintType` and then point at them - if we're opening up the system to allow other artifacts to be passed from a resource to another, could we rely on the `inputs` to `outputs` mapping that already naturally occur when you're tying a resource to another in a blueprint with `resources: [..]`? 
- as the owner might be either workload/deliverable/<something_else?>, how should ` #@ data.values.(workload|deliverable)` look like in the interpolation? perhaps `#@ data.values.owner. ... `?

rfc/rfc-0000-common-blueprint.md
martyspiewak: Ya that would no be valid- it'd have to be template or ytt

rfc/rfc-0000-common-blueprint.md
cirocosta: ooooooh I see, I was thinking about the type of the value from the jsonpath query, makes sense 👍

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+apiVersion: carto.run/v1alpha1
+kind: ClusterBlueprintSelector
+metadata:
+  name: my-selector
+spec:
+  blueprintRef:
+    name: my-blueprint
+  ownerType:
+    apiVersion: carto.run/v1alpha1
+    kind: Workload
```
sclevine: Do we need another field that defines the param name for template interpolation? Or maybe `$(workload)$` and `$(deliverable)$` get replaced with something generic (like `$(descriptor)$`)?

I could see both approaches working, but the former would allow template re-use across descriptors with common fields.

rfc/rfc-0000-common-blueprint.md
scothis: We should avoid "blueprints" in this RFC as the concept and CRD overloading can get confusing. Where we're talking about ClusterSupplyChain and ClusterDelivery, let call out those CRDs by name.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+        - resource: first-resource
+          output: foo
+          name: a-new-resource-name
+        - resource: zero-resource
+          output: zoo
+          name: a-new-resource-name
+
+    - name: third-resource
+      # mutually exclusive - blueprintName, options
+      blueprintName: another-another-blueprint
```
scothis: Consider making this field a local reference. If we ever want to distinguish types of blueprints or bridge from namespaced to cluster scoped blueprints it could be helpful to have room to grow.
```suggestion
      blueprintRef:
        name: another-another-blueprint
```

rfc/rfc-0000-common-blueprint.md
scothis: If we're treating this resource as a type system definition, it would be nice to capture an (optional?) open api schema for the structure of each type. While I'm not sure runtime type checking is particularly important, this info could be extremely useful for template authoring tools that can understand the structure of available input values.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+  resources:
+    - name: first-resource
+      inputs:
+        - resource: zero-resource
+          output: zoo
+          name: zoo-resource-name
+      inline:
+        # mutually exclusive - template, ytt
+        template: {}
+        ytt: {}
```
scothis: Is it worth pulling the templating options into a single struct? Something like:

```suggestion
        template:
          # mutually exclusive - unstructured, ytt
          unstructured: {}
          ytt: ""
```

nit: ytt templates are a string

rfc/rfc-0000-common-blueprint.md
scothis: Dynamic input types also need to be accounted for in the stamping context

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+
+        outputs:
+          - name: foo
+            valuePath: .status.url
+
+    - name: second-resource
+      # mutually exclusive - blueprintName, options
+      options:
+        - name: another-blueprint
+        - name: diff-blueprint
```
scothis: Does each option also need a selector to decide which is picked?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+          name: a-new-resource-name
+        - resource: second-resource
+          output: bar
+          name: another-new-name
+  
+  outputs:
+    - name: baz
+      valuePath: resources[first-resource].outputs[foo]
+    - name: fizz
+      valuePath: resources[second-resource].outputs[bar]
```
scothis: Where is the data structure these paths are resolved against defined?

rfc/rfc-0000-common-blueprint.md
scothis: I wonder if controller-gen would be smart enough to handle generating the schema for recursive structures

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+[migration]: #migration
+
+This RFC proposes all new CRDs, so there is no need for migration.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+- We would need to support both this new architecture, as well as the currently architecture and CRDs for some period of time, which might be challenging.
```
scothis: It might be possible to support the existing API on top of this new API, but if we make changes to the stamping context, translating the templates will be difficult.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+* A `ClusterBlueprintSelector` defines a `spec.blueprintRef`, `spec.selector` (per the v1alpha2 syntax), and `spec.ownerType` to define the type of owner to match against. This type takes an `apiVersion` and `kind` to leave open the possibility to later support other owner types, though this RFC proposes only supporting the existing workload and deliverable types.
+
+Example with resources, inlining, two outputs, and workload selector:
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: ClusterBlueprint
+metadata:
+  name: my-blueprint
+spec:
+  blueprintTypeRef:
```
scothis: The target of the reference is a ClusterBlueprintType, but from the perspective on the ClusterBlueprint, this is the type being outputted.

If we were to fully expand the ref, this example would look like:
```yaml
blueprintTypeRef:
  kind: ClusterBlueprintType
  name: my-baz-fizz-type
```

```suggestion
  outputTypeRef:
```

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
+* A `ClusterBlueprintType` defines only `spec.outputs`, which is an array of output names. By leaving the `name` key in the array, we leave open the possibility to later add `type` to each output, though this RFC currently proposes only supporting string outputs.
+* A `ClusterBlueprintSelector` defines a `spec.blueprintRef`, `spec.selector` (per the v1alpha2 syntax), and `spec.ownerType` to define the type of owner to match against. This type takes an `apiVersion` and `kind` to leave open the possibility to later support other owner types, though this RFC proposes only supporting the existing workload and deliverable types.
+
+Example with resources, inlining, two outputs, and workload selector:
+```yaml
+apiVersion: carto.run/v1alpha1
```
scothis: As a step towards decomposing Cartographer into components. Now would be a good time to move these new CRDs under a different api group?

Suggestions:
- blueprint.carto.run
- flow.carto.run
- voyage.carto.run
- current.carto.run
- fluid.carto.run

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+
+# Migration
+[migration]: #migration
+
+This RFC proposes all new CRDs, so there is no need for migration.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
```
scothis: This is a major API break. Some percentage of current and prospective users will see this change as a signal that Cartographer is not a stable platform to build on top of.

It's often tolerable for a project to make a change like this once in its life, but that's it. Is this the proposal we want to fire that shot for? Does it encompass every breaking change we want to make going forward? Are we confident that this model is correct enough that we won't need another major breaking change to the model in the near future?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,220 @@
...
+
+# What it is
+[what-it-is]: #what-it-is
+* `ClusterBlueprints` will encapsulate everything the current templates allow, while adding `spec.resources` to enable defining multiple resources (i.e. snippets).
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
+* A `ClusterBlueprintType` defines only `spec.outputs`, which is an array of output names. By leaving the `name` key in the array, we leave open the possibility to later add `type` to each output, though this RFC currently proposes only supporting string outputs.
+* A `ClusterBlueprintSelector` defines a `spec.blueprintRef`, `spec.selector` (per the v1alpha2 syntax), and `spec.ownerType` to define the type of owner to match against. This type takes an `apiVersion` and `kind` to leave open the possibility to later support other owner types, though this RFC proposes only supporting the existing workload and deliverable types.
```
scothis: Should Cartographer continue to have direct knowledge of Workload and Deliverable? If we want to make this a generic extension point in the future, having direct knowledge of some resource now may make supporting custom types harder.

Even if we don't expose custom type to users, the implementation should be as generic as possible to minimize the effort to later go fully generic.

matt-royal:
I suspect these new resources would be simpler to understand if you included an example of how you'd replace an existing resource (e.g. ClusterImageSource) using them early in the RFC, then build to a more complex but realistic example. The current examples are filled with fake-sounding names and aren't part of a real workflow, which I didn't find very helpful.

rfc/rfc-0000-common-blueprint.md
emmjohnson: So before we introduced `ClusterBlueprintTypes`, an input had three fields, `resource` - name of the resource to get the output, `output` - which output to use from the resource, `name` - a name to  refer to the resource in this resource (swapability). Now, we should be able to get rid out of output and should be the same as before.

rfc/rfc-0000-common-blueprint.md
emmjohnson: Our outputs today our too prescriptive, this allows for the operators to define the boundaries for their platform.

rfc/rfc-0000-common-blueprint.md
emmjohnson: I think `ClusterTemplate` should be defined how you described so that it is explicit in the blueprint. I don't think we should change the behavior of a blueprint to  have to expose something.

rfc/rfc-0000-common-blueprint.md
emmjohnson: I think another field could make sense, otherwise `owner` seems sensible.

emmjohnson: @scothis what do you mean? ownerTypes and outputs would need to be available in the context for stamping.

rfc/rfc-0000-common-blueprint.md
emmjohnson: Yes, we were being brief. Can add for clarity

rfc/rfc-0000-common-blueprint.md
emmjohnson: Today, we do some internally translating of arrays to maps for ease of you (params). I think we were thinking along those same lines. Do you prefer other notation @scothis?

rfc/rfc-0000-common-blueprint.md
emmjohnson: Keen to hear others thoughts

rfc/rfc-0000-common-blueprint.md
emmjohnson: Do you have suggestions towards fully generic? Right now we have different reconcilers around workload and deliverable. I imagine if we were to allow an extension we would need to register a new kind to reconcile against or we should move all of this logic into the blueprint reconciler.

rfc/rfc-0000-common-blueprint.md
emmjohnson: > of how you'd replace an existing resource (e.g. ClusterImageSource) using them early in the RFC, then build to a more complex but realistic example. The current examples are filled with fake-sounding names and aren't part of a real workflow, which I didn't find very helpful.

Hey @matt-royal did  you see this - https://gist.github.com/cirocosta/affe3ec8c21560a302db1dc2e69c5ae7

rfc/rfc-0000-common-blueprint.md
scothis: The stamping context currently exposes the path to input values based on their type. When the input type becomes a degree of freedom, we no longer control the top-level key space, collisions between types and other existing top-level keys can happen.

For example, these values are currently in our samples

https://github.com/vmware-tanzu/cartographer/blob/6d592c8f8d7ad2a961f556976f9bd33460f9b726/examples/shared/app-operator/app-deploy-template.yaml#L80

https://github.com/vmware-tanzu/cartographer/blob/6d592c8f8d7ad2a961f556976f9bd33460f9b726/examples/shared/app-operator/kpack-image-template.yaml#L47

rfc/rfc-0000-common-blueprint.md
scothis:
```suggestion
      options:
        - name: another-blueprint
          selector: {}
        - name: diff-blueprint
          selector: {}
```

rfc/rfc-0000-common-blueprint.md
scothis: I'm less concerned about the mechanics and more so how a user will know what they can specify. Do we define the structure the path is executing against?

rfc/rfc-0000-common-blueprint.md
scothis: - The spec for each resource can be freeform, I don't think cartographer does anything with the spec of these resource other than to expose it to the templates. 
- The status is where things become more important as it's a structure that cartographer will need to write to. If they are not already, the shape and semantics of the status should be identical. 
- Anywhere we have references in templates to "workload" or "deliverable" those should be normalized, or that key made configurable by the selector. See other comments about concerns for stamping context collisions.
- Any differences in functionality should be normalized

Bonus points if the controller implementation for each resource is declarative, even if not configurable.

scothis:
> I suspect these new resources would be simpler to understand if you included an example of how you'd replace an existing resource (e.g. ClusterImageSource) using them early in the RFC, then build to a more complex but realistic example. The current examples are filled with fake-sounding names and aren't part of a real workflow, which I didn't find very helpful.

There are fewer resource kinds/CRDs, but I wouldn't call any of them simpler. The problem domain Cartographer is operating in is the same, as are the capabilities. If anything, the individual resources are slightly more complicated to account for some of the dynamic typing being proposed.

matt-royal:
> There are fewer resource kinds/CRDs, but I wouldn't call any of them simpler. The problem domain Cartographer is operating in is the same, as are the capabilities. If anything, the individual resources are slightly more complicated to account for some of the dynamic typing being proposed.

My comment was more of a meta-commentary on the structure of the RFC itself. I agree that the domain is the same and the new resources are more complex.

rfc/rfc-0000-common-blueprint.md
matt-royal: You could introduce these new CRDs as an experimental extension to handle cases that the existing resources can't handle. This would give you hands-on experience and the ability to change their API until they're a suitable replacement for the existing resources. At that point you could do a major version bump and remove the original resources

rfc/rfc-0000-common-blueprint.md
cirocosta: > _**The status is where things become more important as it's a structure that cartographer will need to write to**. If they are not already, the shape and semantics of the status should be identical._

indeed, that's a big one

rfc/rfc-0000-common-blueprint.md
cirocosta: ^ we also have some aliasing in case there's only one of those (e.g., if you have only a single source, then rather than `$(sources.source.url)$`, one can do `$(source.url)$`

cirocosta:
some thoughts:

- it'd be interesting to try out (similar to how we did for multipath/snippets) an example that's more close to some supplychains we have

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+- Author(s): @emmjohnson @martyspiewak
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes collapsing `ClusterSupplyChain`, `ClusterDelivery` and templates (not including `ClusterRunTemplate`, as that resource is in the domain of [Runnable](https://cartographer.sh/docs/v0.3.0/reference/runnable/)) into a common type, `ClusterBlueprint`. 
+As part of this, the RFC proposes that outputs be dynamically typed via a new CR `ClusterBlueprintType` and remove the statically typed outputs that exist today (url, revision, image, config). This allows us to maintain the contract of swappable templates, without restricting outputs to a predefined set.
```
evankanderson: Do we need a formal CR for this?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+This RFC proposes collapsing `ClusterSupplyChain`, `ClusterDelivery` and templates (not including `ClusterRunTemplate`, as that resource is in the domain of [Runnable](https://cartographer.sh/docs/v0.3.0/reference/runnable/)) into a common type, `ClusterBlueprint`. 
+As part of this, the RFC proposes that outputs be dynamically typed via a new CR `ClusterBlueprintType` and remove the statically typed outputs that exist today (url, revision, image, config). This allows us to maintain the contract of swappable templates, without restricting outputs to a predefined set. 
+Additionally, the selectors will now be specified in a new CR `ClusterBlueprintSelector`.
```
evankanderson: 👍  It looks like this is collapsing 7 resources (concepts) into 3, which seems like it may make the system slightly more approachable. I'm slightly concerned about decomposing ClusterSupplyChain into orthogonal ClusterBlueprintSelector and ClusterBlueprints. I do still wonder whether an even smaller of concepts is possible.

Today (11 concepts):
* Workload
* Delivery
* ClusterSupplyChain
* ClusterDelivery
* 5 x ClusterFooTemplate
* ClusterRunTemplate
* Runnable

After (7 concepts):
* Workload
* Delivery
* ClusterBlueprint
* ClusterBlueprintType
* ClusterBlueprintSelector
* ClusterRunTemplate
* Runnable

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+
+  - Today we have seven CRDs that are needed to create a supply chain and delivery. To simplify the data model for users, we can
+collapse all seven CRDs into a few common CRDs. Additionally, whereas today the overhead to start reconciling templates would be adding five identical
+reconcilers, this would enable us to have one reconciler for all blueprints and more easily add meaningful data to their statuses.
```
evankanderson: This may also enable users in the future to define their own blueprints for additional types, or have a blueprint which vends both an image and source.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+  - The `ClusterBlueprintSelector` will enable users to separate RBAC between template authors and operators who want to restrict what blueprints can be selected.
+
+- What is the expected outcome?
+
+  - Simplified on-boarding and user experience.
+
+# What it is
+[what-it-is]: #what-it-is
+* `ClusterBlueprints` will encapsulate everything the current templates allow, while adding `spec.resources` to enable defining multiple resources (i.e. snippets).
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
```
evankanderson: This feels weird to me. It seems like the big change is that `spec.resources` is an object map (key `name`), and that resources list may contain any of `template`, `ytt`, or `ref` (`blueprintRef`).

Do we need to support both length-1 `resources` list and `template`?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
```
evankanderson: Can we use a [LocalObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core) or just a name for `blueprintTypeRef` here (and just call it `type`)?

I don't think we want to vary on the ClusterBlueprintType, so we don't need apiVersion / kind here.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+* A `ClusterBlueprintType` defines only `spec.outputs`, which is an array of output names and optional schema (openAPI schema). Adding a schema would allow for template authoring tools to understand the structure of the outputs.
```
evankanderson: You'll think I'm strange for writing this, but does `ClusterBlueprintType` need `spec` and `status`?

Counterpoint: Yes, so that you can indicate that the type is invalid because the OpenAPI is nonsensical / forbidden.

Counter-counterpoint: invalid OpenAPI should be rejected at admission time.

I'm asking because this feels like a "policy" type resource that doesn't have an explicit status (like "Role" or "ConfigMap").

It also feels like the API would work even without this resource, so I may be blowing it out of proportion.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+# How it Works
+[how-it-works]: #how-it-works
+
+This RFC does not propose any new functionality (other than snippets), but rather proposes re-defining the API and CR's to encapsulate the same functionality that exists today.
+
+
+# Migration
+[migration]: #migration
+
+This RFC proposes all new CRDs, so there is no need for migration.
```
evankanderson: I'm not sure this is valid -- there are existing Cartogrpher users; how do they migrate their existing templates to the new system?

Straw-person: the existing ClusterFooTemplate types will become (optional) controllers which simply translate ClusterFooTemplate objects into ClusterBlueprint objects.

rfc/rfc-0000-common-blueprint.md
evankanderson: See my suggestion above that the ClusterFooTemplates could reconcile into ClusterBlueprint resources, giving a "bridge" duration from old API to new.

rfc/rfc-0000-common-blueprint.md
evankanderson: Question: do we need to have both Delivery and Workload? If we had a "Do It (repeatedly)" resource that supplied parameters to be applied to a ClusterBlueprintSelector, would we need to be able to handle multiple "run one of these" resource types?

rfc/rfc-0000-common-blueprint.md
evankanderson: The doc calls out Runnable / ClusterRun as a distinct unit. The rest of the resources seem related; I'd be inclined to have shorter apigroup names, all things being equal.

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+
+  - Simplified on-boarding and user experience.
+
+# What it is
+[what-it-is]: #what-it-is
+* `ClusterBlueprints` will encapsulate everything the current templates allow, while adding `spec.resources` to enable defining multiple resources (i.e. snippets).
+* A `ClusterBlueprint` can take one of `spec.resources`, `spec.template`, and `spec.ytt`. It will allow inlining of templates in `spec.resources`.
+* A `ClusterBlueprint` must specify a `blueprintTypeRef` which defines the outputs required from the blueprint. The `ClusterBlueprint`'s `spec.outputs` must match those outputs.
+* A `ClusterBlueprintType` defines only `spec.outputs`, which is an array of output names and optional schema (openAPI schema). Adding a schema would allow for template authoring tools to understand the structure of the outputs.
+* A `ClusterBlueprintSelector` defines a `spec.blueprintRef`, `spec.selector` (per the v1alpha2 syntax), and `spec.ownerType` to define the type of owner to match against. This type takes an `apiVersion` and `kind` to leave open the possibility to later support other owner types, though this RFC proposes only supporting the existing workload and deliverable types.
```
evankanderson: I feel like you might want to add a `spec.namespaceSelector` to allow limiting certain "golden paths" to certain namespaces, but that might be a separate feature request. (You're rewriting everything, so it's hard for me to tell what's up for negotiation.)

rfc/rfc-0000-common-blueprint.md
evankanderson: `outputType` or `type` for me, if we need typing at all. (Not sure how it's enforced / implemented)

rfc/rfc-0000-common-blueprint.md
evankanderson: I'm not sure I understand what the selector here is supposed to do; this is the first reference to it.

I'm slightly concerned about casting a wide net with selectors in ClusterBlueprints -- is the extra power & complexity worth it?

rfc/rfc-0000-common-blueprint.md
```diff
@@ -0,0 +1,224 @@
...
+  selector:
+    matchLabels: []
+    matchFields: []
+    matchExpressions: []
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This RFC does not propose any new functionality (other than snippets), but rather proposes re-defining the API and CR's to encapsulate the same functionality that exists today.
```
evankanderson: It feels like Blueprint-as-label-selection-of-other-Blueprints is new functionality. Additionally, it's not clear what `type` does; today the ClusterFooType resources use the type to vary the `outputs` schema.

evankanderson (commented):
I'm generally positive on this change, despite the number of comments.

My biggest comment is that I'd consider some of the naming more closely -- in particular, ClusterBlueprintSelector feels like the powerful lever here connecting developer CI requests with ops teams' paved paths catalogues. As such, it seems like it deserves a first-class name, not simply a descriptor of the fields.

Personally, I'd also think about how to bridge the current APIs to the new API, rather than a "throw it out and start over" approach. If nothing else, it will give you illustrative examples of behavior you need to continue to support.

sclevine:
Here's a gist I put together with some recommended adjustments: https://gist.github.com/sclevine/2ee1650b566f7562fbee5b6c6d346135

rfc/rfc-0000-common-blueprint.md
bmoussaud: is it worth keeping the `Workload` & `Delivery` resources, if both can match `ClusterBlueprint` resources?

squeedee (approved):
Passed all internal reviews

PR #767, opened 2022-03-30
RFC for Maven and Artifactory support

### LINKED ISSUE #741: R4RFC: Workload spec update for maven artifacts
### Description of problem

Currently the [workload spec](https://cartographer.sh/docs/v0.2.0/reference/workload/) has fields for `source.git` and `source.image`, neither of which is suitable for basing a workload off a maven artifact. We are requesting an RFC that proposes an update to the workload spec to address this use case.

### Open question

What about S3 artifacts ([fluxcd Bucket resource](https://fluxcd.io/docs/components/source/buckets/))? Would we treat these the same as maven artifacts?

emmjohnson:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/741-maven-artifacts/rfc/rfc-0000-maven-artifacts-support.md)

### DIFF
--- a/rfc/rfc-0000-maven-artifacts-support.md
+++ b/rfc/rfc-0000-maven-artifacts-support.md
@@ -10,7 +10,7 @@
 # Summary
 [summary]: #summary
 
-Modify `owner.spec.source` to allow for specifying a JAR file.
+Modify `owner.spec.source` to allow for specifying a Maven and Artifactory repository.
 
 # Motivation
 [motivation]: #motivation
--- a/rfc/rfc-0000-maven-artifacts-support.md
+++ b/rfc/rfc-0000-maven-artifacts-support.md
@@ -1,0 +1,160 @@
+# Meta
+[meta]: #meta
+- Name: Maven Artifacts Support
+- Start Date: 2022-03-30
+- Author(s): @emmjohnson
+- Status: Draft
+- RFC Pull Request: (leave blank)
+- Supersedes: N/A
+
+# Summary
+[summary]: #summary
+
+Modify `owner.spec.source` to allow for specifying a JAR file.
+
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Today we support workloads being built from git and image source. Many developers in the field also need support for JARs.
+- What use cases does it support?
+  - Maven
+- What is the expected outcome?
+  - To enable more types of workloads.
+
+# What it is
+[what-it-is]: #what-it-is
+We need to modify the Owner specs to create an unambiguous way for specifying a JAR file. There are two proposed options, the goal is to get community buy-in on a solution before ratifying and the RFC will then be modified to suggest the agreed upon spec.
+
+This option allows for the developer to know what is necessary for a maven workload by examining the spec.
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: Workload
+metadata:
+  name: my-workload
+spec:
+  source:
+      maven: 
+        url: https://my-artifactory/my-projects/my-app.jar
+
+---
+apiVersion: carto.run/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: my-supply-chain
+spec:
+  resources:
+    - name: source-provider
+      templateRef:
+        kind: ClusterSourceTemplate
+        options:
+          - name: source-template
+            selector:
+              matchFields:
+                - key: spec.maven.url
+                  operator: Exists
+```
+
+This option allows us to support many artifacts, but the platform operator would have to communicate to the developer what parameter to use. 
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: Workload
+metadata:
+  name: my-workload
+spec:
+  params:
+    - name: source-type
+      value: maven
+  source:
+      url: https://my-artifactory/my-projects/my-app.jar
+
+---
+apiVersion: carto.run/v1alpha1
+kind: ClusterSupplyChain
+metadata:
+  name: my-supply-chain
+spec:
+  resources:
+    - name: source-provider
+      templateRef:
+        kind: ClusterSourceTemplate
+        options:
+          - name: source-template
+            selector:
+              matchFields:
+                - key: spec.params[?(@.name=="source-type")].value
+                  operator: In
+                  values:
+                    - "maven"
+```
+
+# How it Works
+[how-it-works]: #how-it-works
+
+This is the technical portion of the RFC, where you explain the design in sufficient detail.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+# Migration
+[migration]: #migration
+
+This section should document breaks to public API and breaks in compatibility due to this RFC's proposed changes. In addition, it should document the proposed steps that one would need to take to work through these changes. Care should be give to include all applicable personas, such as application developers, supply chain/delivery authors, and template authors.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs have been considered?
+  
+  This option would allow all of the logic to be put on the operator. Wildcards are currently not supported today.
+  ```yaml
+    apiVersion: carto.run/v1alpha1
+    kind: Workload
+    metadata:
+      name: my-workload
+    spec:
+      source:
+        url: https://my-artifactory/my-projects/my-app.jar
+    
+    ---
+    apiVersion: carto.run/v1alpha1
+    kind: ClusterSupplyChain
+    metadata:
+      name: my-supply-chain
+    spec:
+      selector:
+        integration-test: "options-with-values"
+      resources:
+        - name: source-provider
+          templateRef:
+            kind: ClusterSourceTemplate
+            options:
+              - name: source-template
+                selector:
+                  matchFields:
+                    - key: spec.source.url
+                      operator: In
+                      values:
+                        - "*.jar"
+  ```
+
+- Why is this proposal the best?
+- What is the impact of not doing this?
+
+# Prior Art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+- Do we need any additional information from the Owner besides the url of the JAR?
+
+# Spec. Changes (OPTIONAL)
+[spec-changes]: #spec-changes
+Does this RFC entail any proposed changes to CRD specs? If so, please document changes here.
+This section is not intended to be binding, but as discussion of an RFC unfolds, if spec changes are necessary, they should be documented here.
--- a/rfc/rfc-0000-maven-artifacts-support.md
+++ b/rfc/rfc-0000-maven-artifacts-support.md
@@ -16,15 +16,18 @@ Modify `owner.spec.source` to allow for specifying a JAR file.
 [motivation]: #motivation
 
 - Why should we do this?
-  - Today we support workloads being built from git and image source. Many developers in the field also need support for JARs.
+  - Today we support workloads being built from git and image source. Many developers in the field also need support
+    Maven and Artifactory repositories hosting their JAR and WAR files.
 - What use cases does it support?
-  - Maven
+  - Maven and Artifactory repositories
 - What is the expected outcome?
   - To enable more types of workloads.
 
 # What it is
 [what-it-is]: #what-it-is
-We need to modify the Owner specs to create an unambiguous way for specifying a JAR file. There are two proposed options, the goal is to get community buy-in on a solution before ratifying and the RFC will then be modified to suggest the agreed upon spec.
+We need to modify the Owner specs to create an unambiguous way for specifying a JAR file. There are two proposed
+options, the goal is to get community buy-in on a solution before ratifying and the RFC will then be modified to suggest
+the agreed upon spec.
 
 This option allows for the developer to know what is necessary for a maven workload by examining the spec.
 ```yaml
@@ -35,7 +38,7 @@ metadata:
 spec:
   source:
       maven: 
-        url: https://my-artifactory/my-projects/my-app.jar
+        url: https://maven.pkg.jetbrains.space/mycompany/p/projectkey/my-maven-repo
 
 ---
 apiVersion: carto.run/v1alpha1
@@ -43,16 +46,14 @@ kind: ClusterSupplyChain
 metadata:
   name: my-supply-chain
 spec:
+  selectorMatchFields:
+    - key: spec.maven.url
+      operator: Exists
   resources:
     - name: source-provider
       templateRef:
+        name: source-template
         kind: ClusterSourceTemplate
-        options:
-          - name: source-template
-            selector:
-              matchFields:
-                - key: spec.maven.url
-                  operator: Exists
 ```
 
 This option allows us to support many artifacts, but the platform operator would have to communicate to the developer what parameter to use. 
@@ -66,7 +67,7 @@ spec:
     - name: source-type
       value: maven
   source:
-      url: https://my-artifactory/my-projects/my-app.jar
+      url: https://maven.pkg.jetbrains.space/mycompany/p/projectkey/my-maven-repo
 
 ---
 apiVersion: carto.run/v1alpha1
@@ -74,18 +75,16 @@ kind: ClusterSupplyChain
 metadata:
   name: my-supply-chain
 spec:
+  selectorMatchFields:
+    - key: spec.params[?(@.name=="source-type")].value
+      operator: In
+      values:
+        - "maven"
   resources:
     - name: source-provider
       templateRef:
+        name: source-template
         kind: ClusterSourceTemplate
-        options:
-          - name: source-template
-            selector:
-              matchFields:
-                - key: spec.params[?(@.name=="source-type")].value
-                  operator: In
-                  values:
-                    - "maven"
 ```
 
 # How it Works
@@ -109,38 +108,9 @@ Why should we *not* do this?
 [alternatives]: #alternatives
 
 - What other designs have been considered?
-  
-  This option would allow all of the logic to be put on the operator. Wildcards are currently not supported today.
-  ```yaml
-    apiVersion: carto.run/v1alpha1
-    kind: Workload
-    metadata:
-      name: my-workload
-    spec:
-      source:
-        url: https://my-artifactory/my-projects/my-app.jar
-    
-    ---
-    apiVersion: carto.run/v1alpha1
-    kind: ClusterSupplyChain
-    metadata:
-      name: my-supply-chain
-    spec:
-      selector:
-        integration-test: "options-with-values"
-      resources:
-        - name: source-provider
-          templateRef:
-            kind: ClusterSourceTemplate
-            options:
-              - name: source-template
-                selector:
-                  matchFields:
-                    - key: spec.source.url
-                      operator: In
-                      values:
-                        - "*.jar"
-  ```
+  - We did consider using options to select a template for a generic spec.source.url, however
+    we don't believe matchFields will be able to distinguish between different repository types.
+    That information tends to be hidden inside the particular protocol of the repository and not the URL.
 
 - Why is this proposal the best?
 - What is the impact of not doing this?
@@ -152,7 +122,7 @@ Discuss prior art, both the good and bad.
 
 # Unresolved Questions
 [unresolved-questions]: #unresolved-questions
-- Do we need any additional information from the Owner besides the url of the JAR?
+- Do we need any additional information from the Owner besides the url of the Artifact?
 
 # Spec. Changes (OPTIONAL)
 [spec-changes]: #spec-changes


rfc/rfc-0000-maven-artifacts-support.md
```diff
@@ -0,0 +1,160 @@
...
+This option allows for the developer to know what is necessary for a maven workload by examining the spec.
+```yaml
+apiVersion: carto.run/v1alpha1
+kind: Workload
+metadata:
+  name: my-workload
+spec:
+  source:
+      maven: 
+        url: https://my-artifactory/my-projects/my-app.jar
```
scothis: I'm not familiar with maven URLs like this, granted it's been several years since I've been deep in this ecosystem. This example is just a jar on an http server. That jar/http server may happen to be part of a maven repository, but that'd be coincidence.

Maven artifacts are identified by a `groupId`, `artifactId`, `version` (which may be a snapshot and are a bit more complicated to resolve) and a `type` (often `jar`). These coordinates are resolved against a set of repositories. Not every repository will hold every artifact, there's no way to know without asking. Each repository has a base URL that the artifacts are then resolved from. The `groupId` and `artifactId` translate into path segments under the base URL. For each artifact there's a `maven-metadata.xml` index that lists the specific versions and the artifacts. The `version` key may be listed directly in this index, or if a snapshot the newest match is used. Only after using the repository URL, the groupId, the artifactId, the index file for the artifact and the version, can you then resolve the fully qualified URL for the artifact.

Client specific configuration about how to talk to each repository (like auth) are managed in a separate config file, usually located at [`~/.m2/settings.xml`](https://maven.apache.org/settings.html).

scothis (commented):
What does the resource we expect the supply chain to stamp out for maven source look like? It may be worth hitting pause on this until we have a better sense of what's resources are being choreographed.

rfc/rfc-0000-maven-artifacts-support.md
emmjohnson: Thanks @scothis. Definitely outside of my element here. We did modify the example to drop the `.jar`. So you are suggesting we need something like:
```
source:
  maven:
    groupid:
    artifactid:
    url:
    index:
    version:
    config:
```

emmjohnson: We searched around on an artifactory and I'm not sure I saw all of the fields that you listed. Would love it if you had an example you could point me to.

scothis: > We searched around on an artifactory and I'm not sure I saw all of the fields that you listed. Would love it if you had an example you could point me to.

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management

scothis: > Thanks @scothis. Definitely outside of my element here. We did modify the example to drop the `.jar`. So you are suggesting we need something like:
> 
> ```
> source:
>   maven:
>     groupid:
>     artifactid:
>     url:
>     index:
>     version:
>     config:
> ```

My suggestion is to not change the Workload resource until we have a sense of what the resource being stamped out looks like. I would expect something like this, but that's possibly insufficient based on the how the stamped resource defines the model. Most prominently this model assumes the set of repositories to resolve against are defined somewhere else.

```yaml
spec:
  source:
    maven:
      groupId:
      artifactId:
      version:
      type:
```

squeedee: At the very least that spec is missing the repository location. I was hoping there was a Resourceful URL that properly represented an artifact.

squeedee: perhaps?
```yaml
spec:
  source:
    maven:
      repository: 
      groupId: 
      artifactId:
      version:
      type:
```

where repository is just the repository url?

emmjohnson: Here is what concourse was using for its parameters for maven - https://github.com/nulldriver/maven-resource. I don't see version or know what we mean by type.

Also, another open question is any of the information needed only from the platform operator? Or should all of this be coming from the developer?

rszumlakowski: It was my intent that the new `MavenRepository` resource which will be stamped out by `ClusterSourceTemplate` will have input fields very similar to what the Maven Resource for Concourse (linked by @emmjohnson in the above comment) uses:

* `url`: Required. The location of the repository.
* `snapshot_url`: Optional. The location of the snapshot repository.
* `artifact`: Required. The artifact coordinates in the form of groupId:artifactId:type[:classifier]
* `username`: Optional. Username for accessing an authenticated repository.
* `password`: Optional. Password for accessing an authenticated repository.
* `skip_cert_check`: Optional. If set to true, will ignore all certificate errors when accessing an SSL repository. NOTE: this will supersede the repository_cert configuration if also specified.
* `repository_cert`: Optional. CA/server certificate to use when accessing an SSL repository.

The `Workload` resource should take the same inputs (or maybe some platform-operator-subset of them could be pushed off to the supply chain configuration, as appropriate).

rfc/rfc-0000-maven-artifacts-support.md
```diff
@@ -0,0 +1,130 @@
...
+    - key: spec.maven.url
+      operator: Exists
+  resources:
+    - name: source-provider
+      templateRef:
+        name: source-template
+        kind: ClusterSourceTemplate
+```
+
+This option allows us to support many artifacts, but the platform operator would have to communicate to the developer what parameter to use.
```
rszumlakowski, line 59: I favour this second approach. We have been called on to support not just general Maven artifacts but specifically Artifactory artifacts, too. I wouldn't be surprised if there are customer use cases for Nexus artifacts or even things like NPM, too. This second approach seems more scaleable.

rfc/rfc-0000-maven-artifacts-support.md
squeedee: Strong opinion: Those secrets (PW, user, cert) need to be k8s secrets and not params.
Weak opinion: the artifact is probably easier to build as 3 fields as per POM layout.

rfc/rfc-0000-maven-artifacts-support.md
scothis: Artifactory is a server side implementation of many different types of artifact repositories. Clients should have no knowledge that an artifact is hosted by an Artifactory repository. A client for one type of artifact repository (maven) has no ability to pull artifacts in a different type of repository (npm, oci, etc) unless it also happens to understand that repository api (unlikely).

rfc/rfc-0000-maven-artifacts-support.md
scothis: > Also, another open question is any of the information needed only from the platform operator? Or should all of this be coming from the developer?

This is a good question to focus on. The Workload intentionally doesn't support every field that exists on the Flux GitRepository resource under `.spec.source.git`, only the most essential fields a developer is expected to specify.

scothis: > perhaps?
> 
> ```yaml
> spec:
>   source:
>     maven:
>       repository: 
>       groupId: 
>       artifactId:
>       version:
>       type:
> ```
> 
> where repository is just the repository url?

While an artifact exists in a repository, it's uncommon to look in exactly one repository for that artifact. Even for the same groupId/artifactId, it's very common to split releases, milestones and snapshots into distinct repositories.

emmjohnson: @scothis What is `spec.source.maven.repository.type` here?

scothis: > What is `spec.source.maven.repository.type` here?

It's the "type" of "packaging" for the artifact. By default the type is `jar`, but could also be `pom`, `maven-plugin`, `ejb`, `war`, `ear`, `rar`, etc. They use different names for the same concept in different contexts...

https://maven.apache.org/pom.html#packaging

https://maven.apache.org/pom.html#dependencies

scothis: reading https://maven.apache.org/pom.html#dependencies, `classifier` could also be important

heyjcollins:
@rashedkvm @atmandhol

rfc/rfc-0000-maven-artifacts-support.md
rszumlakowski: I have updated the design document again.  The new suggested specification for the `MavenResource` fields are now:

```
spec:
  artifact:
    id: groupId:artifactId:type[:classifier] # required
    versionFilter: ">=3.1.0-rc.1 <3.2.0"     # optional
  repository:
    url: https://my-maven-repository.com     # required
    authentication:
      secretRef: my-authn-secret             # optional
  transport:                                 # optional
    skipSslCheck: true or false              # optional
    caCertificate:
      secretRef: my-ca-certificate           # optional
  interval: 5m                               # optional, default 1m
```

This specification accommodates the `repository.type` and `classifier` fields in the `artifact.id` field.  The secrets have been broken out and will be specified with `secretRef` fields.  I figured we would need one of them for the application-level authentication (i.e.: user name and password for the repository) and another secret for transport-level authentication (e.g. CA certificate).

When it comes to dividing parameters between the `Workload` and the rest of the supply chain, maybe only the fields under `artifact` need to go under the `Workload` and the rest could be handled by the supply chain?

rfc/rfc-0000-maven-artifacts-support.md
rszumlakowski: The reason that we are calling out Artifactory repositories specifically is that customers are asking for Artifactory support explicitly.

sclevine: > customers are asking for Artifactory support explicitly

Let's focus on use cases for Cartographer as an open source project in these spaces 🙂

rszumlakowski: I have removed references to "Artifactory" from my design document.  Our team has decided to focus on Maven functionality first without any Artifactory-specific functionality.  That should suit the open source project, right?

rfc/rfc-0000-maven-artifacts-support.md
xyloman: @rszumlakowski Not sure the language that the MavenResource would be written in.  However, if it were a native java application it could make use of [DefaultArtifact](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/DefaultArtifact.java) which comes from the Aether library which is what is used by maven internally to resolve dependneices.  I contributed to a similar [capability in Spinnaker](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/maven/MavenArtifactCredentials.java#L51) many years ago.  Maven [version resultion](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/maven/MavenArtifactCredentials.java#L148) worked similar to the `spec.artifact.id` that you are referencing above.

xyloman: I agree with accounting for authentication and the possibility that the maven repository would have a custom ca.  The downside of using the artifact id notation `groupId:artifactId:type[:classifier]` is it does not carry with it the repository the artifact might resolve like in the case of a container image registry reference.  In the case of an ImageRepository reference, we get the k8s native capability for configuring PullSecrets to use by configuring registry credentials in the developer namespace.  Since this is a CRD I would propose that there be a custom secret of type MavenRepositorySecret.

```
apiVersion: v1
kind: Secret
metadata:
  name: MyMavenRepositorySecret
type: maven.org/MavenRepositorySecret
data:
  uri: https://my-maven-repository.com
  username: repo-username
  password: repo-password
  skipSslCheck: true or false
  caCertificate: |
    <certificate details go here>
```

Then the workload could reference the MavenRepository:

```
spec:
  artifact:
    id: groupId:artifactId:type[:classifier] # required
    versionFilter: ">=3.1.0-rc.1 <3.2.0"     # optional
    repository: 
      secretRef: MyMavenRepositorySecret
```

If a repository.secretRef is not provided then it could default to say Maven Central or a default secret of type maven.org/MavenRepositorySecret in the namespace.  If more than one is declared then something would have to error out.  Taking this approach would eliminate the need for the developer to need to configure all of the repository metadata for each workload.  Given a namespace could many artifacts declared that could come from the same maven repository the amount of boiler plate per workload could be cut down by using this proposal.

rfc/rfc-0000-maven-artifacts-support.md
```diff
@@ -0,0 +1,130 @@
...
+# Motivation
+[motivation]: #motivation
+
+- Why should we do this?
+  - Today we support workloads being built from git and image source. Many developers in the field also need support
+    Maven and Artifactory repositories hosting their JAR and WAR files.
+- What use cases does it support?
+  - Maven and Artifactory repositories
+- What is the expected outcome?
+  - To enable more types of workloads.
```
rashedkvm, line 24: @emmjohnson should the reference to `Artifactory` be more specific and not confused as a source type here? Similar to the comment in https://github.com/vmware-tanzu/cartographer/pull/767#discussion_r839942652

rfc/rfc-0000-maven-artifacts-support.md
rszumlakowski: @xyloman : Unfortunately, the controller that will be managing the `MavenRepository` resource will be written in Golang, like the rest of Cartographer and related TAP software, so we won't be able to use those handy Java APIs.

Also, I've update the `MavenRepository` specification to break up the `artifactId` field into individual fields, like this:

```yaml
spec:
  artifact:
    groupId: com.my.place                    # required
    artifactId: my-application               # required
    packaging: jar                           # required
    classifier: real-classy                  # optional
    semver: ">=3.1.0-rc.1 <3.2.0"            # optional
  repository:
    url: https://my-maven-repository.com     # required
    authentication:
      secretRef: my-authn-secret             # optional
  transport:                                 # optional
    skipSslCheck: true or false              # optional
    caCertificate:
      secretRef: my-ca-certificate           # optional
  interval: 5m                               # optional, default 1m
```

I like the idea of the custom `MavenRepositorySecret` secret type.  Are there any thoughts from the Cartographer team?  @jwntrs @emmjohnson ?  I'm not sure if we will have time to implement it for the first release, though.

xyloman: @rszumlakowski I agree with breaking up the components that make up the ID.  Will semver support maven snapshots?  I proposed the MavenRepositorySecret primarily to reduce the amount of boilerplate and repetition that a developer would have to go through for one artifact.  Now multiply that by a namespace that might have 10s to 100 artifacts and that might be a poor developer experience.  Given I would assume that most artifacts in a given namespace might all come from the same maven repository.

rszumlakowski: I figure that the `classifier` field will be used to indicate that you're looking for a SNAPSHOT or not.

xyloman: `classifier` can be used in conjunction with SNAPSHOT as a part of the version. 

image

In addition, in order to detect an update to a snapshot typically means that updates are made to the maven-metadata.xml associated with the artifact.  The SNAPSHOT portion of the artifact name will be replaced with a timestamp.  Here is an example of snapshots for the `spring-boot` artifact: 

https://repo.spring.io/ui/native/snapshot/org/springframework/boot/spring-boot/3.0.0-SNAPSHOT/

Notice each build of the snapshot has a timestamp replacing the SNAPSHOT placeholder in the version. The version in the k8s resource would be 3.0.0-SNAPSHOT but upon resolution it would return `spring-boot-3.0.0-20220418.130925-428.jar`.  The maven-metadata.xml can be useful hear to know what latest artifact is: 

https://repo.spring.io/artifactory/snapshot/org/springframework/boot/spring-boot/3.0.0-SNAPSHOT/maven-metadata.xml

Also, take note that the `classifier` will be called out in the metadata for `sources` and `javadoc` which are artifacts purposed at the same time but for different `classified` use cases.

rszumlakowski: I assume that if a user is watching for new SNAPSHOT builds of some artifact then they will want to see the supply chain triggered for each individual build, right?  And each build version will have a different timestamp in it?

I will need to check if the Mastermind semver library supports strings like "SNAPSHOT" in the semver strings.  If not, then I'll need to add a boolean "is_semver" field or something similar.

xyloman: > I assume that if a user is watching for new SNAPSHOT builds of some artifact then they will want to see the supply chain triggered for each individual build, right? And each build version will have a different timestamp in it?

Yes, this is exactly what the desired outcome would be.   I imagine that the resource would poll every minute or every few minutes the [maven-metadata.xml](https://repo.spring.io/artifactory/snapshot/org/springframework/boot/spring-boot/maven-metadata.xml) at the root of the repo for a given groupId and artifactId.  This would tell me which versions are available.  This could be helpful in the case of the semver example you provided above.  If its not in the list then a folder will not exist that corresponds with that version. If the version does exist then a [maven-metadata.xml](https://repo.spring.io/artifactory/snapshot/org/springframework/boot/spring-boot/3.0.0-SNAPSHOT/maven-metadata.xml) must exist for that given version, which will contain the information such as the following for the lastUpdated version: 

```
<?xml version="1.0" encoding="UTF-8"?>
<metadata modelVersion="1.1.0">
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot</artifactId>
  <version>3.0.0-SNAPSHOT</version>
  <versioning>
    <snapshot>
      <timestamp>20220419.103253</timestamp>
      <buildNumber>431</buildNumber>
    </snapshot>
    <lastUpdated>20220419103350</lastUpdated>
    <snapshotVersions>
...
    </snapshotVersions>
</metadata>
```

This would apply to snapshots.  When a version does not contain -SNAPSHOT suffix then maven-metadata.xml will not exist in the version folder similar to [this](https://repo.spring.io/ui/native/libs-release/org/springframework/boot/spring-boot/2.6.6/).  However, a [maven-metadata.xml](https://repo.spring.io/artifactory/libs-release/org/springframework/boot/spring-boot/maven-metadata.xml) will exist at the groupId + artifactId level of the maven repository containing what the latest released version is based upon semver and previous versions that have been released in the past that might be candidates for a semver filter.

rszumlakowski: Okay, so it sounds like if an end user wants to deploy SNAPSHOT builds on their cluster then the supply chain would have to go through a different flow:  it would need to look at different metadata files in different folders.  It would have to look for the builds with the newest timestamp and supply that version to the rest of the supply chain.

I'll add an `isSnapshot` boolean field to indicate that the controller will need to follow a snapshot flow.

rszumlakowski: Oh, I understand more now.  We won't need an `isSnapshot` field to distinguish the two flows.  The presence of a -SNAPSHOT versions in in the base `metadata.xml` and individual `metadata.xml` files per version indicate the presence of SNAPSHOT builds.  The controller will be able to figure out which flow to use without an extra field.  Thank you for your explanations.  They were very helpful.

xyloman: I agree, @rszumlakowski, yes I believe the presence of -SNAPSHOT in the semver field should designate to look at the snapshotVersions section of the maven-metadata.xml

xyloman: Also, wanted to state that `packaging` should be optional and the default value would be jar.  I believe the majority of users are going to use jar packaging format. 

If semver is optional what is going to be the default behavior? 
Should the semver field be called version to be more consistent with the maven naming conventions?

rszumlakowski: We are aiming for the name `semver` since that's what the FluxCD `GitRepository` resource uses.  We are using the `GitRepository` resource to model this behaviour.

If the semver field is left out then the controller would always return the highest version of the artifact.

I agree that the `packaging` should be optional and the default should be `jar`.

xyloman: > We are aiming for the name semver since that's what the FluxCD GitRepository resource uses. We are using the GitRepository resource to model this behaviour.

I thought a lot about this and consistency is important.  However, the audience of the MavenResource is likely to get confused by the field semver.  Other fields that exist on the resource map to their counter parts in the pom.xml (e.g. groupId, artifactId, packaging etc.)  However, semver does not match.  I propose for usability sake the semver field be named `version` to be consistent with the maven pom instead of consistent with the GitRepository resource.  Furthermore, semver for the GitRepository resource makes sense since git uses that terminology.

scothis: The `semver` field is an option on a Flux GitRepository and not the only way to specify a commit. You can also specify a commit directly by its `commit` ref, `tag` or the HEAD of a `branch`. Curiously, Cartographer does not expose `semver` as an option for git source and nobody has asked for it yet.

The `version` field in maven has its own [range syntax](https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification).

xyloman: @scothis that is a great point on the range syntax that maven has is slightly different than the syntax that was shared above.  I would assume that keeping the syntax similar to what Maven users are familiar with.

rfc/rfc-0000-maven-artifacts-support.md
sclevine: Was more calling out the unqualified reference to "customers" -- Artifactory is fine :)

rfc/rfc-0000-maven-artifacts-support.md
rszumlakowski: At Cartographer office hours today we decided to go use the Maven field name `version` and the Maven range syntax rather than the `semver` field as previously discussed.

PR #783, opened 2022-04-04
RFC: Submitted Status in Resources

### LINKED ISSUE #764: R4RFC Submitted Condition on owner.status.resources[].conditions
### Description of problem

If ratified, #738 adds a `owner.status.resource[].conditions` condition with `Type: Healthy`

In this RFC we should propose adding `Type: Submitted` that is the same as `owner.status.conditions[Type=ResourcesSubmitted]` but as it relates to the individual resource. This way we no longer hide multiple `Submission` errors as we do today.

If #738 does not introduce a top level `Type:Ready` to `owner.status.resource[].conditions`, we should do that also, and it's value should be the lesser of the two `Type:Healthy` and `Type:Submitted`

squeedee:
[#764]

### DIFF
--- a/rfc/rfc-0000-submitted-status-in-resources.md
+++ b/rfc/rfc-0000-submitted-status-in-resources.md
@@ -3,7 +3,7 @@
 [meta]: #meta
 
 - Name: Include submitted condition in owner.status.resources.conditions
-- Start Date: 2022-4-1
+- Start Date: 2022-04-01
 - Author(s): [Rasheed Abdul-Aziz](https://github.com/squeedee) [Todd Ritchie](https://github.com/pivotal-todd-ritchie)
 - Status: Draft
 - RFC Pull Request: (leave blank)
--- a/rfc/rfc-0000-submitted-status-in-resources.md
+++ b/rfc/rfc-0000-submitted-status-in-resources.md
@@ -1,0 +1,142 @@
+# Meta
+
+[meta]: #meta
+
+- Name: Include submitted condition in owner.status.resources.conditions
+- Start Date: 2022-4-1
+- Author(s): [Rasheed Abdul-Aziz](https://github.com/squeedee) [Todd Ritchie](https://github.com/pivotal-todd-ritchie)
+- Status: Draft
+- RFC Pull Request: (leave blank)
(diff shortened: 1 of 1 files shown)


martyspiewak (approved):

emmjohnson (approved):

rfc/rfc-0000-submitted-status-in-resources.md
```diff
@@ -0,0 +1,142 @@
...
+# How it Works
+
+[how-it-works]: #how-it-works
+
+During the creation of all stamped objects, the realizer provides a conditions collection for each resource. This
+includes any existing conditions (
+see [Resource Status on Workload RFC](https://github.com/vmware-tanzu/cartographer/blob/rfc-resources-report-status/rfc/rfc-0000-allow-resources-to-report-status.md))
+and a calculated `type:Ready` condition.
+
+Each `type:Submitted` condition follows the same implementation as the top level `type:ResourcesSubmitted` condition.
```
scothis, line 96: Is the desire to keep the existing top-level `ResourcesSubmitted` condition? With its status aggregated from all of the individual resource level `Submitted` conditions?

It's interesting to have a couple different dimensions of condition aggregation.

rfc/rfc-0000-submitted-status-in-resources.md
```diff
@@ -0,0 +1,142 @@
+# Meta
+
+[meta]: #meta
+
+- Name: Include submitted condition in owner.status.resources.conditions
+- Start Date: 2022-4-1
```
scothis: nit
```suggestion
- Start Date: 2022-04-01
```

scothis (approved):

rfc/rfc-0000-submitted-status-in-resources.md
squeedee: I hadn't planned to change it's current spec, which is not aggregated, only "first occurrence"

Do you feel strongly that we should add behavior change to this RFC?

rawlingsj (approved):

idoru (approved):

jwntrs (approved):

PR #786, opened 2022-04-04
Create 0.3.0 docs

squeedee:
Creates version 3 documentation

- introduces the tutorials
- fixes redirect issues when changing versions
- ~removes patch releases.~ (We will do this during site migration, looking for automation opportunities as we go)
- fix broke tables in the style guides


cirocosta (approved):
looked at the preview environment and the set of files changed, lgtm!

PR #792, opened 2022-04-05
Mention the pre-reqs so the easy install actually works

squeedee:
Mention the pre-reqs so the easy install actually works


martyspiewak (changes requested):
Do we want to link to the prereqs in v0.2.0? It seems likely that that'll get outdated. 

I'm also not convinced we need to list prereqs here- if people are using cartographer for the first time i'd expect them to go to docs for installation instructions

squeedee:
Agree with Marty.

PR #794, opened 2022-04-06
config: add `app.kubernetes.io/component` label

cirocosta:

- add an overlay used by the release process to add to all objects the `app.kubernetes.io/component` label to identify objects as being all part of cartographer

following the best practices when it comes to labelling objects (see
[1]), by making use of the recommended `app.kubernetes.io/component`,
we're able to clearly identify those objects that are part of
cartographer (useful for any tool that might want to do some form of
matching against anything cartographer-related, e.g., getting rid of all
of them based on matching on such label).

the approach I took here is to make use of ytt overlays in order to add
the label to all components _except_ the Namespace.

[1]: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

## Release Note

- Add `app.kubernetes.io/component: cartographer.carto.run` label to all cartographer objects (except namespace).

## PR Checklist

### DIFF
--- a/hack/overlays/component-labels.yaml
+++ b/hack/overlays/component-labels.yaml
@@ -20,4 +20,4 @@ metadata:
   #@overlay/match missing_ok=True
   labels:
     #@overlay/match missing_ok=True
-    app.kubernetes.io/component: cartographer.carto.run
+    app.kubernetes.io/component: cartographer
--- a/hack/overlays/component-labels.yaml
+++ b/hack/overlays/component-labels.yaml
@@ -14,14 +14,7 @@
 
 #@ load("@ytt:overlay", "overlay")
 
-#@ def not_matcher(marg):
-#@   def matcher(*args, **kwargs):
-#@     return not marg(*args, **kwargs)
-#@   end
-#@   return matcher
-#@ end
-
-#@overlay/match by=not_matcher(overlay.subset({"kind":"Namespace"})),expects="1+"
+#@overlay/match by=overlay.not_op(overlay.subset({"kind":"Namespace"})),expects="1+"
 ---
 metadata:
   #@overlay/match missing_ok=True
--- a/hack/overlays/component-labels.yaml
+++ b/hack/overlays/component-labels.yaml
@@ -1,0 +1,30 @@
+# Copyright 2021 VMware
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


cirocosta:
cc @scothis

hack/overlays/component-labels.yaml
```diff
@@ -0,0 +1,30 @@
...
+#@ load("@ytt:overlay", "overlay")
+
+#@ def not_matcher(marg):
+#@   def matcher(*args, **kwargs):
+#@     return not marg(*args, **kwargs)
+#@   end
+#@   return matcher
+#@ end
+
+#@overlay/match by=not_matcher(overlay.subset({"kind":"Namespace"})),expects="1+"
```
scothis: https://carvel.dev/ytt/docs/v0.40.0/lang-ref-ytt-overlay/#overlaynot_op

```suggestion
#@overlay/match by=overlay.not_op(overlay.subset({"kind":"Namespace"})),expects="1+"
```

hack/overlays/component-labels.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+metadata:
+  #@overlay/match missing_ok=True
+  labels:
+    #@overlay/match missing_ok=True
+    app.kubernetes.io/component: cartographer.carto.run
```
emmjohnson: Reading thru best practices - https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ I'm not sure if component or part-of is the best choice. Either way, they don't appear to be qualified, are there examples of this? Otherwise I would just expect the label to be `cartographer`.

hack/overlays/component-labels.yaml
cirocosta: awesome!

hack/overlays/component-labels.yaml
scothis: I don't have a strong feeling here, but we should be consistent between Cartographer and Cartographer Conventions (and all future components).

cirocosta: > _I'm not sure if component or part-of is the best choice_

I think from the perspective of cartographer as an umbrella project, `component` makes sense per the definition of _"the component within the architecture"_, such that `conventions` is a component, and ... `cartographer` is another? maybe cartographer should be more like ... `core` or something? (imo `cartographer` and `conventions` as the "components" make sense having not heard any of us mentioning "core" very often)

> _they don't appear to be qualified,_

yeah I noticed that too - I personally think it's nice to qualify, but indeed, seems not to be the most common (I think we should then make it not fully qualified)


---

tl;dr: I think `app.kubernetes.io/component` makes sense, fully qualified name could be dropped in favor of just plain `cartographer` here (and `conventions` there in `cartographer-conventions`)

wdyt?

scothis: from [slack](https://kubernetes.slack.com/archives/C02HKPSEKV1/p1649270158823619?thread_ts=1649263370.179639&cid=C02HKPSEKV1)

```suggestion
    app.kubernetes.io/component: cartographer
```

scothis (approved):

emmjohnson (approved):

PR #799, opened 2022-04-08
Introduce RFC to correlate resource inputs with output

### LINKED ISSUE #744: R4RFC: Artifact tracing
### Description of problem

For a particular resource in a supply chain, we would like to be able to assert that a given input produced a specific output.

We are requesting an RFC that describes how to achieve that result using the following mechanics:

- Allow resources (via template) to define fields in their status that indicate which inputs and outputs to compare. If the input matches a specified input field, then you can trust the output (transforms input e.g. source to image)
- Allow resources (via template) to defined fields in their status that indicate which field should match a particular input (pass through input e.g. source to source).
- Allow resources (via template) to indicate that by holding spec updates we can trust a combination of the observed generation and some "healthy" status condition to know that a value in the spec produced a result in the status.

idoru:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/744-input-output-correlation/rfc/rfc-0000-input-output-correlation.md)

closes #744

## PR Checklist


rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+
+[Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738)
+
+# Unresolved Questions
+
+[unresolved-questions]: #unresolved-questions
+
+It is unclear how to solve for resources which require matching against their spec, but do not conform to the
+`observedGeneration` pattern. The same can be said for resources which do not expose sufficient detail to write proper
+health determination rules.
```
waciumawanjohi, line 185: The [ClusterDeploymentTemplate](https://cartographer.sh/docs/v0.3.0/reference/template/#clusterdeploymenttemplate) is relevant prior art in Cartographer. ObservedCondition is similar to specifying a health rule. ObservedCondition requires observedGeneration to equal generation.

ObservedMatch is used particularly in the case where the resource does not use observedGeneration. Users can use observedMatch to specify a field in the status that reflects a field in the spec. When the two match, we can assume that the status is the result of the spec.

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+An interesting side effect of these mechanics is that Cartographer will not propagate artifacts output by a resource to
+the next resource unless they can be correlated to input provided to the resource by Cartographer. This is desirable
+from a security perspective, and the outcomes of this behavior are explored in detail in the
+[Draft RFC Read Resources Only When In Success State](https://github.com/vmware-tanzu/cartographer/pull/556).
+
+# What it is
+
+[what-it-is]: #what-it-is
+
+New fields are added to templates which allow template authors to specify correlation rules using `observedMatches`,
```
waciumawanjohi: Cartographer already has a template with the top level field `observedMatches`: [ClusterDeploymentTemplate](https://cartographer.sh/docs/v0.3.0/reference/template/#clusterdeploymenttemplate).

The proposed `ObservedMatches` works differently than this prior art. The existing implementation matches two fields that exist on the stamped object. The proposal matches fields on the template to fields on the stamped object.

1. Should the name change so that artifact tracing can work with ClusterDeploymentTemplate?
2. Should the name change so that we don't overload a term with two implementations?

waciumawanjohi: We can assume that fields in the template exist on the spec of the object. Do we need ObservedMatches to match from template to object? Or can we use the same implementation that exists on ClusterDeploymentTemplate, where users are expected to compare the object's spec to its status?

This implementation has the advantage of allowing users to specify `metadata.generation` must equal `status.observedGeneration` explicitly.

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+    output: "$(spec.sourceUrl)"                 #evaluated against resource spec
+  - input: "$(sources.source.revision)"         #evaluated against inputs in template context
+    output: "$(spec.sourceRevision)"            #evaluated against resource spec
+```
+
+Since the spec of the resource may change independently of the output, when the `outputMatches` makes any assertion
+against the `spec`, Cartographer must also ensure that the current output is a result of processing the current spec by
+making additional assertions:
+
+* Resource is healthy, as per [Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738)
```
waciumawanjohi, line 139: Won't healthy/unhealthy be a concern for resources where the observedMatch is to a field in the status (as well as for resources where the observedMatch is to a field in the spec).

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+
+It is unclear how to solve for resources which require matching against their spec, but do not conform to the
+`observedGeneration` pattern. The same can be said for resources which do not expose sufficient detail to write proper
+health determination rules.
+
+# Spec. Changes
+
+[spec-changes]: #spec-changes
+
+All templates will now have a mandatory `observedMatch` field, which must be an array containing at least one element
```
waciumawanjohi: Making this optional, rather than required, will mean that Carto will still be able to work with all resources. It will also allow end users to determine if they want to tradeoff performance for artifact tracing.

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+* Resource is healthy, as per [Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738)
+* Resource `observedGeneration == generation`
+
+If these assertions cannot be made, for example, if the resource does not implement the `observedGeneration` pattern,
+then Cartographer stops processing the resource and flags an error on it. Great care must also be taken that `healthy`
+status rules specified by the template author corresponds to the situation where the output is indeed a result of the
+spec with the current `observedGeneration`, otherwise outputs may not truly be the result of an input.
+
+Additionally, relying on the spec to make assertions imposes a further restriction to avoid stalling artifact
+propagation: Cartographer must not stamp out any further changes to the resource until it has finished processing the
```
waciumawanjohi, line 147: Above, the RFC states that
> Cartographer will not propagate artifacts output
which I refer to as _reading_ an object.
Here the RFC states
> Cartographer must not stamp out any further changes to the resource
which I refer to as _writing_ an object.

Is this proposal to hold reads or writes or both?

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,217 @@
...
+
+[migration]: #migration
+
+All templates must have `outputMatches` added to them.
+
+Any template which requires `outputMatches` against the resource spec must also be configured to identify success,
+failure and unknown states - which map to Healthy, Unhealthy and Unknown if the mechanism to be relied upon is
+[Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738).
+
+# Drawbacks
```
waciumawanjohi, line 169: Supply chains often rely on resources that continually report the latest good output (e.g. kpack). When such an object is processing new input, Cartographer is still able to read the output and use it to assert the proper shape of the following object.

WIth this proposal, Cartographer will need to adopt a caching strategy after reading good outputs of objects. Consider the case of a supply chain where resource B relies on resource A. When resource A has a good output, it will at some point get an update to its spec. It will begin process the new spec and `Healthy` will become unknown. As a result, Cartographer will no longer read resource A's output. That means that Cartographer will now be unable to assure that resource B has the proper spec, so resource B will be unreadable as well.

rfc/rfc-0000-input-output-correlation.md
idoru: Both.

rfc/rfc-0000-input-output-correlation.md
idoru: Good catch on naming. Actually I noticed the names are not even consistent in my document. I'm going to rename to `correlationRules` for the next iteration.

With regards to the matching implementation, I think matching metadata.generation to status.observedGeneration implicitly in the cases where it is needed helps guarantee mechanics work as desired. There are also stipulations under #spec-changes which also try to limit jsonpath expressions to intended uses.

rfc/rfc-0000-input-output-correlation.md
idoru: The healthy check only exists when the `spec` is implicated because we're relying on fields in the spec corresponding to the status.

Once the output is in the status, it is true that the resource may reconcile again and be unhealthy, but as long as the exact inputs specified in correlation rules match the output we can reasonably assume it is the same input.

This, as `#drawbacks` states is the reason why the template authors must take great care. The matching rules must uniquely identify the instance of the input (e.g. as with URL+revision for a source)

rfc/rfc-0000-input-output-correlation.md
idoru: I see. A cache, that is, or something like `lastGoodOtput` on the status in your [original RFC](https://github.com/vmware-tanzu/cartographer/pull/556)?

rfc/rfc-0000-input-output-correlation.md
idoru: Actually, I take that back. I think you're right, otherwise we may not always be able to correlate inputs and outputs.

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,223 @@
...
+spec:
+  template:
+    apiVersion: my-scanner/v1
+    kind: SourceScan
+    metadata:
+      name: ...
+    spec:
+      scanUrl: "$(sources.source.url)$"
+      sourceRevision: "$(sources.source.url)$"
+correlationRules:
```
scothis:
```suggestion
  correlationRules:
```

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,223 @@
...
+  - input: "$(sources.source.url)"              #evaluated against inputs in template context
+    output: "$(status.outputs.url)"             #evaluated against resource status
```
scothis: a few minor changes in this one suggestion:
- `input` and `output` are overloaded for resource artifacts. Using `expected` and `actual`, other terms can fit as well
- the "input" is setting a expectation for the value to find, updated to use the regular cartographer template syntax
- the output is a path on the stamped object, dropping the template syntax since we don't use it on other values read from the stamped object.

```suggestion
  - expectedValue: "$(sources.source.url)$"              #evaluated against inputs in template context
    actualPath: ".status.outputs.url"             #evaluated against the in-cluster resource
```

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,223 @@
...
+  - input: "$(sources.source.revision)"         #evaluated against inputs in template context
+    output: "$(spec.sourceRevision)"            #evaluated against resource spec
+```
+
+Since the spec of the resource may change independently of the output, when the `correlationRules` makes any assertion
+against the `spec`, Cartographer must also ensure that the current output is a result of processing the current spec by
+making additional assertions:
+
+* Resource is healthy, as per [Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738)
+* Resource `observedGeneration == generation`
```
scothis, line 140: the template author should be required to indicate that the resource supports observedGeneration, many resources don't

rfc/rfc-0000-input-output-correlation.md
```diff
@@ -0,0 +1,223 @@
...
+
+If these assertions cannot be made, for example, if the resource does not implement the `observedGeneration` pattern,
+then Cartographer stops processing the resource and flags an error on it. Great care must also be taken that `healthy`
+status rules specified by the template author corresponds to the situation where the output is indeed a result of the
+spec with the current `observedGeneration`, otherwise outputs may not truly be the result of an input.
+
+Additionally, relying on the spec to make assertions imposes a further restriction to avoid stalling artifact
+propagation: Cartographer must not stamp out any further changes to the resource until it has finished processing the
+current spec completely. That is, it must have reached either a Healthy or Unhealthy status (as per
+[Allow Resources to Report Status](https://github.com/vmware-tanzu/cartographer/pull/738)).
```
scothis, line 149: We need to avoid situations that could deadlock the supply chain. If an input is passed in that never resolves to Healthy True or False, the stamped resource is stuck, and can never be unstuck because we're no longer writing to its spec.

rfc/rfc-0000-input-output-correlation.md
scothis: Where is the cache stored? Assume that the controller pod will crash or be rescheduled at an inconvenient time; a new controller instance needs to be able to pickup and continue processing as if nothing happened.

idoru: It would be very convenient to put it on the owner status, but this may cause us to exceed maximum size. One thing we've discussed privately within the team is storing outputs as another custom resource. The latter feels like it may be the only choice, barring the addition of another component for durable cache storage.

rfc/rfc-0000-input-output-correlation.md
idoru: I've added a `correlationTimeout` setting to the template for now; I'm not sure there's a better solution whilst remaining agnostic.

PR #826, opened 2022-04-27
Check if owner.status.resource[x] has changed before updating owner.status

emmjohnson:

closes # 

## Release Note

## PR Checklist


cirocosta:
Screen Shot 2022-04-27 at 12 50 07 PM

pkg/controllers/deliverable_reconciler.go
```diff
@@ -169,17 +170,20 @@ func (r *DeliverableReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	return r.completeReconciliation(ctx, deliverable, resourceStatuses, err)
 }
 
-func (r *DeliverableReconciler) completeReconciliation(ctx context.Context, deliverable *v1alpha1.Deliverable, realizedResources []v1alpha1.RealizedResource, err error) (ctrl.Result, error) {
+func (r *DeliverableReconciler) completeReconciliation(ctx context.Context, deliverable *v1alpha1.Deliverable, resourceStatuses statuses.ResourceStatuses, err error) (ctrl.Result, error) {
 	log := logr.FromContextOrDiscard(ctx)
 	var changed bool
 	deliverable.Status.Conditions, changed = r.conditionManager.Finalize()
 
 	var updateErr error
```
cirocosta, line 181: nit: seems like `updateErr` is only used in place inside the branch below - I think we could go with a straight `:=` right there instead

pkg/realizer/statuses/resource_statuses.go
```diff
@@ -0,0 +1,152 @@
...
+		if status.current == nil {
+			return true
+		}
+		if status.previous == nil {
+			return true
+		}
+		if status.conditionsChanged {
+			return true
+		}
+		if !reflect.DeepEqual(status.current.RealizedResource, status.previous.RealizedResource) {
```
cirocosta, line 75: I remember emj brought up some struggle w/ using deepequal and dealing with timestamps - is that still a problem? I ask that here because outputs have `latestTransitionTime`, and I'm not sure we want to check that in this equality (i.e., I believe we'd like to ignore that value when performing the deep equal).

cirocosta: (a suggestion for achieving that is to use semantic.deepequal from k8s packages and then apply a noop for metav1.Time, or perhaps a custom type that wraps metav1.Time so we can clearly specify that for that particular type we don't care about equality checks - ala [kpack's use of VolatileTime](https://github.com/pivotal/kpack/blob/ed7fe5056666f0ec1aa4378c92db1d52a0a872c5/pkg/apis/core/v1alpha1/volatile_time.go))

cirocosta (approved):
overall lgtm!

PR #832, opened 2022-05-02
Build time service bindings

### LINKED ISSUE #825: R4RFC: Service claims for non-runtime resources
### Description of problem

Currently the workload contains top level service claims that are meant to be exposed to runtime resources. There is currently no mechanism to define service claims that are targeted to other resources in the cluster (e.g. build).

### Proposed solution

Requesting an RFC that outlines a solution for how we could introduce service claims to be consumed by other resources.

squeedee:

RFC to propose mechanisms for non-runtime service bindings.

[closes #825]


squeedee:
Closing without implementing, see https://github.com/vmware-tanzu/cartographer/issues/825

PR #865, opened 2022-05-18
Don't use a third party repo

squeedee:
- allows forks to use the hack.sh local registry
- forks should no longer fail
- means that release testing is less complete, unfortunately

Proof that it works: #867


cirocosta (approved):
lgtm!

PR #874, opened 2022-05-24
Fix a small logging mistake and clean up

squeedee:
Fix a small logging mistake and clean up

# clean up
- Use a specific error name for each section of the realizer (see note)
- merge two realizeError conditionals

# fix
logging tracking error instead of realize error when the tracking fails

# Note

The number of different actions with different errors in #Reconcile
feels like a code smell. We're already concerned by the complexity
of reconcile so this is not a surprise, just further evidence.


idoru (approved):
Really good find! +1 on naming the errors there are so many.

pkg/controllers/deliverable_reconciler.go
```diff
@@ -110,34 +110,34 @@ func (r *DeliverableReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
 	resourceStatuses := statuses.NewResourceStatuses(deliverable.Status.Resources, conditions.AddConditionForResourceSubmittedDeliverable)
-	err = r.Realizer.Realize(ctx, resourceRealizer, delivery.Name, realizer.MakeDeliveryOwnerResources(delivery), resourceStatuses)
 
-	if err != nil {
-		conditions.AddConditionForResourceSubmittedDeliverable(&r.conditionManager, true, err)
-	} else {
-		r.conditionManager.AddPositive(conditions.ResourcesSubmittedCondition(true))
-	}
+	var reconcileErr error
```
cirocosta, line 129: when first looking at the code I noticed that errors above this line seems to not be "aggregated" into `reconcileErr`, so we could have them as plain `err` - wdyt of that?

I think it's indeed worth having specialized naming for the errors after here, but not sure about the ones before (although I can see the argument for consistency 🤔 )

squeedee: We were, by mistake, logging err instead of trackingErr. I believe this occurred because there are so many error return values in this code block that the issue did not stand out during a refactor/shift of code (or the subsequent review).

I'd like to avoid shadowing `err` here, for now,  until we find an abstraction where the ONLY kind of err return value in this block, is one we use or ignore.

cirocosta (approved):

squeedee:
This is not mergable due to realizers getting unified. Will re-implement

squeedee:
implemented here: https://github.com/vmware-tanzu/cartographer/pull/1040

Issue #812, opened 2022-04-18
Allow resources to report status.


jwntrs:
### Description of problem

Currently there is no indication that a particular resource in the supply chain is healthy, unhealthy or in an unknown state.

### Proposed solution

Allow individual resources to report health status (via each template type) as per https://github.com/vmware-tanzu/cartographer/pull/738.

PR #875, opened 2022-05-26
Introduce support for health rules

idoru:
Supports AlwaysHealthy and SingleConditionTypes

When no health rule is specified, Healthy is Unknown until valid outputs are produced from a resources, at which time they will Healthy will become True.

Updates #812

Follows specification from https://github.com/vmware-tanzu/cartographer/pull/738 with the following exceptions:

When no health rule is specified for a template, the reason for a resource's Healthy condition may also be attributed these other values, not originally in the RFC:

- NoResource
- OutputsAvailable
- OutputNotAvailable

Parts not implemented in this PR:

- MultiMatch rules
- Healthy condition in owner conditions

## Release Note

Support reporting Healthy status on resources statuses.

## PR Checklist


pkg/realizer/healthcheck/healthcheck_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
+	v1 "k8s.io/api/core/v1"
```
cirocosta: we usually name the import as `corev1` 😁 

```suggestion
	corev1 "k8s.io/api/core/v1"
```

pkg/realizer/healthcheck/healthcheck.go
```diff
@@ -31,31 +31,33 @@ func DetermineHealthCondition(rule *v1alpha1.HealthRule, realizedResource *v1alp
...
+			return conditions.AlwaysHealthyResourcesHealthyCondition()
 		}
-		if resultString, ok := result[0].(string); ok {
-			conditionStatus := metav1.ConditionStatus(resultString)
-			if conditionStatus == metav1.ConditionFalse || conditionStatus == metav1.ConditionTrue {
-				return conditions.SingleConditionMatchCondition(conditionStatus, rule.SingleConditionType)
-			} else {
-				return conditions.SingleConditionMatchCondition(metav1.ConditionUnknown, rule.SingleConditionType)
+		if rule.SingleConditionType != "" && stampedObject != nil {
+			jsonpathQuery := fmt.Sprintf("{.status.conditions[?(@.type==\"%s\")].status}", rule.SingleConditionType)
```
cirocosta, line 54: another option here would be to do something like

```go
type Conditionable struct {
        Status struct {
                Conditions []metav1.Condition `json:"conditions,omitempty"`
        } `json:"status,omitempty"`
}
```

which could be populated by unmarshalling into it; `Conditonable` could then have some helper methods to do what we're doing here

```go
func (c Conditionable) GetConditionOfType(ttype string) (metav1.Condition, bool) {
   // ...
}
```

wdyt? maybe that's a different way of getting towards the same evaluation but with less of that code in this scope

idoru: Yeah, I was on the fence, but I didn't want to choose between adding a dependency on `mapstructure` or marshalling the map just to unmarshal it back into the struct. I improve the re-use of the private `conditionList` in realizer in the follow-up PR for owner `ResourcesHealthy` and I would still like to remove the json path query here.

pkg/realizer/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/vmware-tanzu/cartographer/pkg/utils"
+)
+
+func DetermineHealthCondition(rule *v1alpha1.HealthRule, realizedResource *v1alpha1.RealizedResource, stampedObject *unstructured.Unstructured) metav1.Condition {
+	if rule == nil {
+		if realizedResource == nil {
+			return conditions.NoResourceResourcesHealthyCondition()
+		} else if len(realizedResource.Outputs) > 0 {
+			return conditions.OutputAvailableResourcesHealthyCondition()
+		} else if realizedResource.TemplateRef != nil && realizedResource.TemplateRef.Kind == "ClusterTemplate" && realizedResource.TemplateRef.APIVersion == "carto.run/v1alpha1" {
```
cirocosta: uuuuhh it seems a little fragile to have these strings right here 😬 

wdyt of leveraging v1alpha1.SchemaGroupVersion

https://github.com/vmware-tanzu/cartographer/blob/693036beb08f237ad45e3715c92a61a7e718cd81/pkg/apis/v1alpha1/groupversion_info.go#L23-L26

for APIVersion? I think the `.Identifier()` or `.String()` method of that struct would do it

given that various `&&`,  might be useful to turn the predicate into a small function that returns a bool, wdyt?

cirocosta (approved):
lgtm! left just a nit comment, but ack'ed! feel free to merge as is or tackle the comment - on you 😁   thanks!

PR #876, opened 2022-05-27
486 runnable output bug

squeedee:
Change the behaviour of runnable's outputs:

1. outputs come from the latest StampedObject with a succeeded:true status. objects without a suceeded status are ignored.
2. not specifying outputs is legal - supports terminal/throwaway tasks


pkg/templates/cluster_run_template.go
```diff
@@ -26,91 +26,68 @@ import (
...
 }
 
-func (t runTemplate) GetOutput(stampedObjects []*unstructured.Unstructured) (Outputs, *unstructured.Unstructured, error) {
-	var (
-		updateError                        error
-		everyObjectErrored                 bool
-		mostRecentlySubmittedSuccesfulTime *time.Time
-		evaluatedStampedObject             *unstructured.Unstructured
-	)
+const StatusPath = `status.conditions[?(@.type=="Succeeded")].status`
```
cirocosta: wdyt of making it more specialized?

```suggestion
const StatusSucceededPath = `status.conditions[?(@.type=="Succeeded")].status`
```

(not a strong opinion at all)

squeedee: did something similar.

idoru (approved):

PR #877, opened 2022-05-27
812 owner healthy status

idoru:

Owners now report `ResourcesHealthy` condition which summarizes the `Healthy` condition of its owned resources.

When any resource has a `Healthy` condition with `Unknown` status, the status of `ResourcesHealthy` is `Unknown`.
If not, when any resource has a `Healthy` condition with `False` status, the status of `ResourcesHealthy` is `False`.
Else, all resources have a `Healthy` condition of `True` and  the status of `ResourcesHealthy` is `True`.

Objects stamped with `ClusterTemplate` without a `healthRule` are now only considered to have a `Healthy` with a status of `True` when the object is successfully created.

builds on PR #875 

updates #812

## Release Note

Owners report `ResourcesHealthy` condition to reflect overall `Healthy` condition of owned resources

## PR Checklist

- [] Modified the docs to match changes

### DIFF
--- a/pkg/realizer/healthcheck/healthcheck.go
+++ b/pkg/realizer/healthcheck/healthcheck.go
@@ -36,6 +36,27 @@ func IsClusterTemplate(reference *corev1.ObjectReference) bool {
 	return false
 }
 
+func OwnerHealthCondition(resourceStatuses []v1alpha1.ResourceStatus, previousConditions []metav1.Condition) metav1.Condition {
+	var previousHealthCondition []metav1.Condition
+	condition := conditions.ConditionList(previousConditions).ConditionWithType(v1alpha1.ResourceHealthy)
+	if condition != nil {
+		previousHealthCondition = append(previousHealthCondition, *condition)
+	}
+	healthyConditionManager := conditions.NewConditionManager(v1alpha1.ResourcesHealthy, previousHealthCondition)
+
+	for _, resourceStatus := range resourceStatuses {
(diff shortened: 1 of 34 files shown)


pkg/realizer/healthcheck/healthcheck.go
```diff
@@ -25,14 +25,39 @@ import (
...
+
+	for _, resourceStatus := range resourceStatuses {
+		resourceHealthyCondition := conditions.ConditionList(resourceStatus.Conditions).ConditionWithType(v1alpha1.ResourceHealthy)
+		if resourceHealthyCondition != nil {
+			healthyConditionManager.AddPositive(*resourceHealthyCondition)
+		}
+	}
+
+	healthyConditionResult, _ := healthyConditionManager.Finalize()
+	healthyCondition := healthyConditionResult[len(healthyConditionResult)-1]
```
cirocosta, line 55: nit: I wonder if we should have a method for this instead of relying on grabbing the latest condition 🤔 

indeed, `Finalize()` will always produce a non-empty array, but, here we're hardcoding the assumption that the aggregated condition is the last one - wdyt of method like "AggregatedCondition()" or something like that that returns that?

cirocosta (approved):
nice! lgtm, left a couple nitpicking comments but that's all 😁

pkg/realizer/healthcheck/healthcheck.go
idoru: Perhaps, I'm also not sure ConditionManager was intended to be used this way. Will follow up in a future PR.

PR #891, opened 2022-06-09
Artifact Tracing with Health Rules

waciumawanjohi:

[Readable](https://github.com/vmware-tanzu/cartographer/blob/artifact-tracing-with-health-rules/rfc/rfc-0000-artifact-tracing-with-health-rules.md)

### DIFF
--- a/rfc/rfc-0000-artifact-tracing-with-health-rules.md
+++ b/rfc/rfc-0000-artifact-tracing-with-health-rules.md
@@ -13,14 +13,16 @@
 
 [summary]: #summary
 
-Tracing allows cartographer to determine which set of inputs (which update) of a stamped object led to a given output (
+Tracing allows Cartographer to determine which set of inputs (which update) of a stamped object led to a given output (
 status) of the object. Some resources may explicitly provide this information, but many do not. This RFC specifies how
 tracing can be accomplished when no such fields can be leveraged in the resource.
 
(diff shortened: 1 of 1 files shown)


squeedee (approved):
Oh yeah!

tbr11 (approved):

idoru (approved):
Looks good, but I wonder about forcing explicit duplication of top-level health rules when there rules for tracing aren't different. Was there a strong case for not having such cleverness?

waciumawanjohi:
> Looks good, but I wonder about forcing explicit duplication of top-level health rules when there rules for tracing aren't different. Was there a strong case for not having such cleverness?

@sclevine and @scothis I remember that you were in favor of this. I would say that the concern was that: health rules for the purpose of reporting health on the workload will most likely be the same as those used to determine if the object was "at rest" for the purpose of artifact tracing. But there is no guarantee that's the case. So we make separate fields to put users in control.

Is that a good summation?

scothis:
Have we implemented these ideas to see how they perform in a cluster? There are enough moving parts that to prove correct behavior in every situation will be difficult (the text of the RFC already indicates that we know there are cases that can deadlock a supply chain).

scothis:
If we step back and look at the larger picture of artifact tracing in the context of the Blueprints API, what would an artifact tracing native API look like?

PR #893, opened 2022-06-09
Artifact tracing with Generation Output Correlation

waciumawanjohi:
## Changes proposed by this PR
Introduce Generation-Output Correlation RFC
[Readable](https://github.com/vmware-tanzu/cartographer/blob/artifact-tracing-with-generation-output-correlation/rfc/rfc-0000-artifact-tracing-with-generation-output-correlation.md)

## PR Checklist

- ~[ ] Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`~
- [ ] Removed non-atomic or `wip` commits
- ~[ ] Filled in the [Release Note](#Release-Note) section above~
- ~[ ] Modified the docs to match changes~

PR #894, opened 2022-06-09
Artifact Tracing with Correlation Rules

waciumawanjohi:
[Readable](https://github.com/vmware-tanzu/cartographer/blob/artifact-tracing-with-correlation-rules/rfc/rfc-0000-artifact-tracing-with-correlation-rules.md)

### DIFF
--- a/rfc/rfc-0000-artifact-tracing-with-correlation-rules.md
+++ b/rfc/rfc-0000-artifact-tracing-with-correlation-rules.md
@@ -1,0 +1,210 @@
+# Meta
+
+[meta]: #meta
+
+- Name: Artifact Tracing with Correlation Rules
+- Start Date: 2022-06-09
+- Author(s): waciumawanjohi
+- Status: Draft <!-- Acceptable values: Draft, Approved, On Hold, Superseded -->
+- RFC Pull Request:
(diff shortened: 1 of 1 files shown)


waciumawanjohi:
Following feedback, the RFC no longer includes caching of any values.

squeedee (approved):
Love it!

tbr11 (approved):

idoru (approved):
LGTM!

scothis:
What resources choreographed by Cartographer follow this pattern today? Without knowing that we are picking up support for common patterns that exist in the wild, this proposal has an element of "if you build it, they will come" vibes.

waciumawanjohi:
> What resources choreographed by Cartographer follow this pattern today? Without knowing that we are picking up support for common patterns that exist in the wild, this proposal has an element of "if you build it, they will come" vibes.

The only resource I know of that would fit would be Tekton, where a Tekton task status includes the spec definition that led to the output.

I agree that this and [Artifact Tracing with Generation Output Correlation](https://github.com/vmware-tanzu/cartographer/pull/893) are Cartographer encouraging the k8s resources in our domain to adopt one of these patterns. If those other resources never do, tracing will always rely on [Artifact Tracing with Health Rules](https://github.com/vmware-tanzu/cartographer/pull/891), which is sub-optimal.

PR #897, opened 2022-06-15
Use Rosetta for env-test tools on M1 macs

squeedee:
The install instructions for the envtest binaries on an M1 Mac do not work. They try to download `darwin-arm64` binaries which do not exist. We use the x86 binaries instead.

@cirocosta Please check these still give you the right binaries on Linux please?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,7 +172,8 @@ and can download them with:
 
 ```
 K8S_VERSION=1.19.2
-curl -sSLo envtest-bins.tar.gz "https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-${K8S_VERSION}-$(go env GOOS)-$(go env GOARCH).tar.gz"
+if [[ "$(go env GOARCH)" == "arm64" ]] && [[ "$(go env GOOS)" == "darwin" ]]; then ENV_TEST_ARCH="amd64"; else ENV_TEST_ARCH="$(go env GOARCH)"; fi # use Rosetta on M1 macs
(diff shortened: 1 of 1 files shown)


scothis:
When setting up envtest, you can set `GOARCH=amd64`. Once the binaries are downloaded you should no longer need that var set (until something wants to download envtest binaries again).

cirocosta:
@squeedee got the right one 😁 

```console
```
$ if [[ "$(go env GOARCH)" == "arm64" ]] && [[ "$(go env GOOS)" == "darwin" ]]; then ENV_TEST_ARCH="amd64"; else ENV_TEST_ARCH="$(go env GOARCH)"; fi # use Rosetta on M1 macs
$ echo $ENV_TEST_ARCH

        amd64

$ curl -sSLo envtest-bins.tar.gz "https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-${K8S_VERSION}-$(go env GOOS)-${ENV_TEST_ARCH}.tar.gz"
$ ls

        envtest-bins.tar.gz

$ tar xvzf envtest-bins.tar.gz

        kubebuilder/
        kubebuilder/bin/
        kubebuilder/bin/etcd
        kubebuilder/bin/kubectl
        kubebuilder/bin/kube-apiserver
        tmp.XbfxeSnYky $
        tmp.XbfxeSnYky $ ls
        envtest-bins.tar.gz  kubebuilder



$ ./kubebuilder/bin/etcd --help

        Usage:
          etcd [flags]
            Start an etcd server.
          etcd --version
            Show the version of etcd.

```

squeedee:
> When setting up envtest, you can set GOARCH=amd64. Once the binaries are downloaded you should no longer need that var set (until something wants to download envtest binaries again).

@scothis that's what the if statement essentially does.. If on mac, and arm, use amd64 instead.

cirocosta (approved):

PR #898, opened 2022-06-15
Avoid bringing up a registry on a MACOS default port

squeedee:
Port 5000 is used by ControlCenter: https://developer.apple.com/forums/thread/682332
We now map 5001->5000 when creating the registry

Co-authored-by: Ciro da Silva da Costa <ciroscosta@vmware.com>


squeedee:
@idoru you may appreciate this.

cirocosta (approved):

PR #899, opened 2022-06-16
This deepcopy only gives us a copy of a garbage collected value anyway

squeedee:


cirocosta (approved):

squeedee:
Obviously the consistant test failures point to an issue with this change. We're not sure why it was introduced, but it feels valuable and would like to capture the reason in a comment.

Issue #486, opened 2021-12-15
Runnable Status is Ready even when stamped object has never been successful


waciumawanjohi:
### Bug description:
When a Runnable is submitted, it will not display output until the created object has a condition `Ready` with a status `True`. But while waiting for a healthy child object, the Runnable itself will still have a status `Ready` with status `True`.

### Steps to reproduce:
Submit these objects:

```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: super-service-account

---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: super-role-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: super-role
subjects:
  - kind: ServiceAccount
    name: super-service-account

---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: super-role
rules:
  - apiGroups: ["*"]
    resources: ["*"]
    verbs: ["*"]

---

apiVersion: carto.run/v1alpha1
kind: ClusterRunTemplate
metadata:
  name: cm
spec:
  outputs:
    val: data.val
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
     generateName: cm-
    data:
      val: $(runnable.spec.inputs.some)$

---
apiVersion: carto.run/v1alpha1
kind: Runnable
metadata:
  name: cm
spec:
  serviceAccountName: super-service-account

  runTemplateRef:
    name: cm

  inputs:
    some: string
```

### Expected behavior:
`k get -o yaml runnable cm`
returns a Runnable with condition `Ready` == False

### Actual behavior:
`k get -o yaml runnable cm`
returns a Runnable with condition `Ready` == True

### Additional context:
See branch `runnable-status-bug` for a test that exercises this.

# TODO
* [ ] fix the issue
* [ ] document the status behavior

martyspiewak:
Just to +1 this, this causes some unexpected behavior at the workload level too if a supplychain is stamping out a runnable because the workload will report that it is `Ready` before the runnable tasks have run because the runnable itself is reporting that it is `Ready`.

jwntrs:
Just to clarify, the desired behaviour for the `Ready` condition on `Runnable` would be as follows:
`Ready: True` -> Latest thing passed
`Ready: False` -> Latest thing failed
`Ready: Unknown` -> Something Running

squeedee:
Current state of the code:

[...]

PR #900, opened 2022-06-16
Runnable status should report the state of the latest stamped object

squeedee:
Introduce a status that tracks the runnable's stamped object

Currently we assume it's a tekton TaskRun so we only
concern ourselves with the "Succeeded" condition

[finishes #486]


pkg/apis/v1alpha1/conditions.go
```diff
@@ -164,4 +165,6 @@ const (
 	FailedToListCreatedObjectsReason                  = "FailedToListCreatedObjects"
 	UnknownErrorReason                                = "UnknownError"
 	ClientBuilderErrorResourcesSubmittedReason        = "ClientBuilderError"
+	SucceededStampedObjectConditionReason             = "SucceededCondition"
```
cirocosta, line 168: **nit** wdyt of just succeeded? e.g.

```diff
  status:
    conditions:
    - type: RunTemplateReady
      reason: Ready
      status: "True"
      message: ""
    - type: StampedObjectCondition
-     reason: SucceededCondition
+     reason: Succeeded
      status: "True"
      message: ""
    - type: Ready
      reason: Ready
      status: "True"
      message: ""
```

this way, similar to how with `RunTemplateReady` we report "ready: true" (rather than "readycondition: true"), we'd be reporting "succeeded: true" rather than "succeededcondition:     true")

cirocosta (approved):
lgtm! just a [small nit](https://github.com/vmware-tanzu/cartographer/pull/900/files#r903844281) around the reason

pkg/apis/v1alpha1/conditions.go
squeedee: So this one's always going to be imperfect, I feel. We (@idoru and myself) thought this would be the "chosen 'pass' field + 'condition'" so as to support other Resource Kinds. So when/if we broaden support outside of Tekton, the condition that is watched for "StampedObjectCondition" could be "Pass", "Ready", "Done", whatever, so the reason would be PassCondition, ReadyCondition, DoneCondition etc... Dropping the "Condition" does also work because the type implies the reason might point to the Stamped Objects Condition, but that would be implicit, not explicit.

squeedee: Also follows the [HealthStatus condition](https://github.com/vmware-tanzu/cartographer/pull/738/files#diff-b070f9379f1bd0cd98da4d4df230daad975709d751738a70f0f6e996425a3eb8R62)

PR #906, opened 2022-06-21
Runnable integration recrun

squeedee:

closes # 

## Release Note

## PR Checklist


hack/tools/go.mod
```diff
@@ -13,6 +13,7 @@ require (
 
 require (
 	4d63.com/gochecknoglobals v0.1.0 // indirect
```
cirocosta, line 15: check out https://4d64.com 😆 (yes, 64 rather than 63)

> _Jordan's Online Shopping Plaza_
> _This site is your one stop online shopping spot!_

totally not what I was expecting lol

pkg/controllers/runnable_reconciler_test.go
```diff
@@ -45,10 +53,252 @@ import (
...
+	_ = v1alpha1.AddToScheme(scheme)
+
+	rts := rtesting.ReconcilerTests{
+		"default service account missing": {
+			Request:           runnableRequest,
+			AdditionalConfigs: nil,
+			GivenObjects: []client.Object{
+				baseRunnable,
+			},
+			ExpectStatusUpdates: []client.Object{
```
cirocosta, line 146: so neat, right? excited to see where it's going! cc @scothis

pkg/controllers/runnable_reconciler_test.go
```diff
@@ -45,10 +53,252 @@ import (
...
+	secretName := "my-secret"
+
+	scheme := runtime.NewScheme()
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = v1alpha1.AddToScheme(scheme)
+
+	rts := rtesting.ReconcilerTests{
+		"default service account missing": {
+			Request:           runnableRequest,
+			AdditionalConfigs: nil,
```
scothis: **nit** rm unused lines

```suggestion
```

pkg/controllers/runnable_reconciler_test.go
```diff
@@ -45,10 +53,252 @@ import (
...
+			d.Name(serviceAccountName)
+			d.CreationTimestamp(now)
+		})
+
+	baseRunnable := dies.RunnableBlank.
+		MetadataDie(func(d *diesv1.ObjectMetaDie) {
+			d.
+				Name(runnableName).
+				Namespace(runnableNamespace).
+				AddLabel("some-val", "first")
```
scothis, line 113: **sugestion** within a callback, the die is mutable, I typically will flatten the mutations rather than chain them. The net effect is the identical

```suggestion
			d.Name(runnableName)
			d.Namespace(runnableNamespace)
			d.AddLabel("some-val", "first")
```

pkg/controllers/runnable_reconciler_test.go
```diff
@@ -17,24 +17,32 @@ package controllers_test
 import (
 	"context"
 	"errors"
+	"testing"
 
+	v1 "dies.dev/apis/core/v1"
+	diesv1 "dies.dev/apis/meta/v1"
```
scothis: **suggestion** I like to make it clear that the imported package is from dies and not the upstream version. This will have avoid conflicts as more packages are imported.
```suggestion
	diecorev1 "dies.dev/apis/core/v1"
	diemetav1 "dies.dev/apis/meta/v1"
```

pkg/controllers/runnable_reconciler_test.go
```diff
@@ -45,10 +53,252 @@ import (
...
+				baseRunnable.SpecDie(func(d *dies.RunnableSpecDie) {
+					d.ServiceAccountName("my-service-account")
+				}),
+			},
+			ExpectStatusUpdates: []client.Object{
+				baseRunnable.
+					StatusDie(func(d *dies.RunnableStatusDie) {
+						d.ConditionsDie(
+							diesv1.ConditionBlank.
+								Type("RunTemplateReady").
```
scothis, line 177: **suggestion** with dies, if you find yourself repeating the same pattern, it can be helpful to store the common bits. Common conditions while structurally generic, are semantically part of the type domain. Along side the other die types you can define something like.

```go
var RunTemplateReadyCondition := diesv1.ConditionBlank.Type("RunTemplateReady")
```

and then consume it

```suggestion
							dies.RunTemplateReadyCondition.
```

scothis (commented):
a few minor suggestions

pkg/controllers/runnable_reconciler_test.go
squeedee: Yup I saw that in other examples, and don't personally like it. Perhaps after driving the tests out a few more times I'd appreciate the encapsulation and abstraction, but when driving out code with tests, I don't like to hide meaning behind indirection, at least not early. I want to see at a glance what's noticeably different. I know plenty of folks skew the other way.

pkg/controllers/runnable_reconciler_test.go
squeedee: that's useful

pkg/controllers/runnable_reconciler_test.go
squeedee: Yeah I'm currently working out how to deal with the discovery client so this is a mental placeholder..

PR #932, opened 2022-07-15
Ensure cache is also keyed by labels for runnables [Backport]

idoru:
## Changes proposed by this PR

Backport #931 into `release/0.4.x` branch

Prevents resources stamped by a RunTemplate but different Runnables from being conflated.
This prevents Cartographer from mistakenly stamping the resources endlessly for each Runnable in turn

[Fixes #930]

## Release Note

Runnable stamped resources now also key by their labels in the cache, ensuring stamped resources from different Runnables sharing the same RunTemplate do not cause endless resources to be stamped [Fix #930]

## PR Checklist

Note: Please do not remove items. Mark items as done `[x]` or use ~strikethrough~ if you believe they are not relevant

- [x] Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`
- [x] Removed non-atomic or `wip` commits
- [x] Filled in the [Release Note](#Release-Note) section above 
- ~[ ] Modified the docs to match changes ~

PR #937, opened 2022-07-19
Add tests, tweak implementation

### LINKED ISSUE #930: Cartographer stamps RunTemplates continuously when referenced by multiple Runnables
When a RunTemplate is referenced by multiple Runnables, Cartographer will endlessly stamp the template alternately for each Runnable.

Reproduce with the following (courtesy of @waciumawanjohi):
```
---
kind: ClusterTask
metadata:
  name: passing-cluster-task
spec:
  description: |-
    A task that passes
  steps:
    - name: solo-step
      image: busybox
      script: |
        exit 0

---
kind: ClusterRunTemplate
metadata:
  name: passing-template
spec:
  template:
    kind: TaskRun
    metadata:
      generateName: passing-
    spec:
      taskRef:
        name: passing-cluster-task
        kind: ClusterTask


---
kind: Runnable
metadata:
  name: passing
spec:
  runTemplateRef:
    name: passing-template

---
kind: Runnable
metadata:
  name: passing-2
spec:
  runTemplateRef:
    name: passing-template
```

waciumawanjohi:
- Adds a regression test
- Implements discriminant creation using O(nlogn) algo rather than O(n^2)

## Release Note

Runnable stamped resources now also key by their labels in the cache, ensuring stamped resources from different Runnables sharing the same RunTemplate do not cause endless resources to be stamped [Fix #930]

## PR Checklist


pkg/repository/repository.go
```diff
@@ -501,16 +501,16 @@ func (r *repository) GetScheme() *runtime.Scheme {
 
 func buildOwnerDiscriminant(labels map[string]string) string {
 	var discriminantComponents []string
-	for key, value := range labels {
-		discriminantComponents = insertSorted(discriminantComponents, fmt.Sprintf("{%s:%s}", key, value))
+
```
idoru, line 504: Good spot. I'm also currently working on another PR which unifies cache key creation at a different layer (the addition of the discriminant has stretched the original design where Cache was aware of how to make cache keys) and would rather work this into there, if the final solution requires sorting at all. My view was either #930 is critical enough that we can merge #931 and #932 first or it could wait for the refactoring.

tests/integration/runnable/runnable_test.go
```diff
@@ -259,6 +260,55 @@ var _ = Describe("Stamping a resource on Runnable Creation", func() {
 					}))
 				})
 			})
+
+			Context("and a second Runnable matches the RunTemplateRef", func() {
```
idoru, line 264: I'm not a fan of this style of regression test, as it's key assertion is blindly waiting for something to not happen. As such, it either has to hope it has waited long enough (else you get false positives), or always has to wait longer than necessary, protracting test execution time.

To be clear, I'm not saying this test is worthless, but my calculus was: If I know we already have tests like this for a single object only being stamped once in here (per input change) which implies the Cache, then is such an integration test much akin to a corner case test for a `Repository -> Cache` bug? If so, what value does such an integration test add that makes it worth paying the time / uncertainty cost?

Interestingly, I think if we are destined to ultimately use ReconcilerTestSuite, then this and similarly structured tests would be converted to forms that reduce the cost (increased test execution time) and eliminate uncertainty.

pkg/repository/repository.go
waciumawanjohi: Assuming [this doc](https://confluence.eng.vmware.com/pages/viewpage.action?pageId=789919969) is still accurate, we want to get the fix into TAP by Friday.

tests/integration/runnable/runnable_test.go
waciumawanjohi: This test unquestionably adds 2 seconds to the integration tests.

> is such an integration test much akin to a corner case test for a Repository -> Cache bug

This is undoubtedly a regression test. There was a bug, it effected customers.

As you point out, this test is similar to the assertion block above it for stamping out a single object. We can reduce the 2 second delay by combining this test with the other (and thus have one 2 second wait rather than two).

pkg/repository/repository.go
idoru: That looks correct. I agree this PR improves on #931 / #932 and in essence I would approve, but since it's based on these PRs which I worked on and they've not been approved yet, it didn't quite feel legitimate for me to do so.

idoru (approved):

PR #951, opened 2022-07-27
Get ServiceAccount tokens with TokenRequest API

### LINKED ISSUE #923: Cartographer fails with ServiceAccountSecretError on k8s 1.24
You'll see an error like this:

```
Status:
  Conditions:
    Last Transition Time:  2022-06-23T07:27:06Z
    Message:
    Reason:                Ready
    Status:                True
    Type:                  SupplyChainReady
    Last Transition Time:  2022-06-23T07:27:06Z
    Message:               service account [my-apps/default] does not have any token secrets
    Reason:                ServiceAccountSecretError
    Status:                False
    Type:                  ResourcesSubmitted
    Last Transition Time:  2022-06-23T07:27:06Z
    Message:               service account [my-apps/default] does not have any token secrets
    Reason:                ServiceAccountSecretError
    Status:                False
    Type:                  Ready
  Observed Generation:     1
  Supply Chain Ref:
    Kind:  ClusterSupplyChain
    Name:  source-to-url
Events:    <none> 
```

This is because in k8s 1.24:

- Secrets are not automatically created for ServiceAccounts 
- ServiceAccounts no longer list their Secrets

We should instead obtain tokens as necessary using the Token API which is the recommended way from the deprecation.

A similar issue occurred with `carvel-kapp-controller`:
(Link to Issue)[https://github.com/vmware-tanzu/carvel-kapp-controller/issues/687]
(Link to Fix)[https://github.com/vmware-tanzu/carvel-kapp-controller/pull/695]

idoru:

Bump KIND on end-to-end tests to 1.24.2

Obtain bearer tokens using the TokenRequest API, which is enabled by default since k8s 1.12. [1]

This is a necessary change to support Kubernetes 1.24+. Since 1.24.0, our current method of enumerating the Secrets of a ServiceAccount to obtain a token is no longer possible because:
 - Secrets are no longer automatically created for ServiceAccounts;
 - ServiceAccounts no longer reference their secrets, making enumeration infeasible. [2]

These changes do not sit well with the KUTTL tests using the mocked control plane. To advance this work, this change switches those over to run with KIND, which is much slower. It is my hope that we can find a way to make this work with the mocked control plane and revert this particular aspect ASAP.

[1] https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
[2] https://itnext.io/big-change-in-k8s-1-24-about-serviceaccounts-and-their-secrets-4b909a4af4e0

Fixes #923

## Release Note

Support k8s 1.24+ by obtaining ServiceAccount tokens with the TokenRequest API.

## PR Checklist


cirocosta (approved):
lgtm! acceptance tests ran well! 😁

PR #970, opened 2022-08-11
Make a meaningful change for upgrade test

emmjohnson:
Previously, we were making a change to README.md. This is no longer triggering a new image to be built. This actually makes sense that it should only trigger on a meaningful change to the source code. 

Root cause for the change: https://github.com/paketo-buildpacks/go-build/pull/335

Co-authored-by: Rasheed Abdul-Aziz <rabdulaziz@vmware.com>


squeedee:
Needs a `uname` check for `Darwin` and switch sed implementations, or some cross platform edit instead.

cirocosta (approved):

idoru (approved):

PR #973, opened 2022-08-15
Add flags to enable metrics and pprof

### LINKED ISSUE #50: Make it possible to activate the metrics endpoint
_"I want to be able to ingest the metrics from the controller"_

```gherkin
Given that the controller is running
When I hit the metrics endpoint
Then I retrieve the metrics out of controller-runtime's instrumentation
```

## context

`controller-runtime` internally keeps track of some pretty useful metrics that an operator can make use of, like how long reconciles have been taking, etc etc

e.g.

```go
	// ReconcileErrors is a prometheus counter metrics which holds the total
	// number of errors from the Reconciler
	ReconcileErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
		Name: "controller_runtime_reconcile_errors_total",
		Help: "Total number of reconciliation errors per controller",
	}, []string{"controller"})


	// ReconcileTime is a prometheus metric which keeps track of the duration
	// of reconciliations
	ReconcileTime = prometheus.NewHistogramVec(prometheus.HistogramOpts{
		Name: "controller_runtime_reconcile_time_seconds",
		Help: "Length of time per reconciliation per controller",
		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
			1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60},
	}, []string{"controller"})
```

_(see https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/internal/controller/metrics/metrics.go#L34-L48)_

to be able to fetch them, the controller needs to start an http server, which them takes care
of exporting those metrics for us. that's done via controller-runtime's manager config: 

```diff
        mgr, err := manager.New(cfg, manager.Options{
                Scheme:             scheme,
-               MetricsBindAddress: "0",
+               MetricsBindAddress: "9000",
        })
```

having that "activated" (non-zero port specified in the BindAddress), anyone can hit that the `/metrics` endpoint on it and retrieve the information (something that a [prometheus](https://prometheus.io/) instance or similar would do).

## approach

recently, we introduced the `-dev` flag to `cmd/controller` in order to tweak the configuration of logging format . 

we could make use of a similar approach for that: perhaps a `-metrics-address`?

another approach is making use of the new `controller-runtime`'s mechanism for tweaking manager settings: [component-config](https://github.com/kubernetes-sigs/controller-runtime/blob/fbf50b04fe17b0f091430d22557245d2aedabfd4/designs/component-config.md).

with it, you're able to tweak a controller's configuration via ... kubernetes objects! e.g.:

```
apiVersion: examples.x-k8s.io/v1alpha1
kind: CustomControllerManagerConfiguration
clusterName: example-test
cacheNamespace: default
metrics:
  bindAddress: :8081
leaderElection:
  leaderElect: false
```

_(see https://github.com/kubernetes-sigs/controller-runtime/blob/fbf50b04fe17b0f091430d22557245d2aedabfd4/examples/configfile/custom/config.yaml)_

I _personally_ prefer just `-metrics-address` from a flag, but if most products that use kubebuilder/controller-runtime are going for component-config (something we should check out), I'd vote towards going with that.

thanks!

### LINKED ISSUE #952: allow enabling pprof
### Description of problem

Sometimes we need to collect cpu/mem/etc profiles, and without an easy way to do so, we currently resort to manually modifying source code to add entries for it, which means having to recompile the controller, which prevents us from quickly collecting those traces in running environments.

### Proposed solution

Perhaps an approach we could take is leveraging the extension to the manager's metrics endpoint and register the pprof handlers there (see [`AddMetricsExtraHandler`](https://github.com/kubernetes-sigs/controller-runtime/blob/196828e54e4210497438671b2b449522c004db5c/pkg/manager/manager.go#L68-L73)) which would then be activate in case a `--enable-pprof` flag is passed in conjunction with a flag for enabling metrics collection (see #50)

```gherkin
Given I initialize the controller with `--metrics-bind-addr` and `--enable-pprof`
When I hit pprof's endpoints
Then I collect the traces I care about
```

idoru:

Adds `--metrics-port` flag, which will bring up prometheus metrics if a non-zero value is provided

Adds `--pprof-port` flag, which will enable pprof on that port if a non-zero value is provided.

These are both enabled when using the Makefile target `run`.

Closes #50
Closes #952

## Release Note

Adds --metrics-port and --pprof-port to the controller startup flags.

## PR Checklist


idoru:
Try this with `make run` and try some metrics, e.g.:
```
curl localhost:9998/metrics | grep reconcile
```

Or do some memory profiling, e.g.:
```
curl localhost:9999/debug/pprof/heap > heapdump
go tool pprof -web heapdump
```

cirocosta:
@idoru thank you! have you considered having `./config/` updated?

my first thought was of making use of an overlay in `vmware-tanzu/package-for-cartographer` that would let you configure these via something like 

```
cartographer:
  metrics_port: <uint>
  pprof_port: <uint>
```

but then looking at the metrics, it seems like the cardinality is all good (thus, no worries about having it on all the time), however _pprof_ is for sure one of those that you might or might not want to have it on all the time ...

soo, wdyt of

- update `./config`'s `Deployment` to have under `args: [ .. ]` a value for `--metrics-port` so that it's always enabled by default (as well as create a `Service` object, probably with `prometheus` annotations ala https://github.com/vmware-tanzu/cartographer-conventions/blob/9634b26dfb5fe7bbdbfa4f0607180c719b589493/config/rbac/auth_proxy_service.yaml#L5-L7

- update `package-for-cartographer` to include an overlay for enabling pprof (I can take this one)

?

thank you!

cirocosta:
fwiw, for folks not super familiar with the metrics exported, from a plain run w/ no workloads/etc in the cluster:

| name | description |
| ---- | ----------- |
| certwatcher_read_certificate_errors_total | Total number of certificate read errors |
| certwatcher_read_certificate_total | Total number of certificate reads |
| controller_runtime_active_workers | Number of currently used workers per controller |
| controller_runtime_max_concurrent_reconciles | Maximum number of concurrent reconciles per controller |
| controller_runtime_reconcile_errors_total | Total number of reconciliation errors per controller |
| controller_runtime_reconcile_total | Total number of reconciliations per controller |
| process_cpu_seconds_total | Total user and system CPU time spent in seconds. |
| process_max_fds | Maximum number of open file descriptors. |
| process_open_fds | Number of open file descriptors. |
| process_resident_memory_bytes | Resident memory size in bytes. |
| process_start_time_seconds | Start time of the process since unix epoch in seconds. |
| process_virtual_memory_bytes | Virtual memory size in bytes. |
| process_virtual_memory_max_bytes | Maximum amount of virtual memory available in bytes. |
| rest_client_requests_total | Number of HTTP requests, partitioned by status code, method, and host. |
| workqueue_adds_total | Total number of adds handled by workqueue |
| workqueue_depth | Current depth of workqueue |
| workqueue_longest_running_processor_seconds | How many seconds has the longest running processor for workqueue been running. |
| workqueue_queue_duration_seconds | How long in seconds an item stays in workqueue before being requested |
| workqueue_retries_total | Total number of retries handled by workqueue |
| workqueue_unfinished_work_seconds | How many seconds of work has been done that is in progress and hasn't been observed by work_duration. Large values indicate stuck threads. One can deduce the number of stuck threads by observing the rate at which this increases. |
| workqueue_work_duration_seconds | How long in seconds processing an item from workqueue takes. |

to generate ^, run `curl -s localhost:9998/metrics | ./table-printer.awk  | grep -v go_`, `table-printer.awk` being:

```awk
#!/usr/bin/awk -f

BEGIN {
  print "| name | description |"
  print "| ---- | ----------- |"
}

/HELP/ {
  line="| " $3 " |"
  for (i = 4; i <= NF; i++) {
    line = line " "$i
  }

  line = line " |"
  print line
}
```

idoru:
> @idoru thank you! have you considered having `./config/` updated?
> 
> my first thought was of making use of an overlay in `vmware-tanzu/package-for-cartographer` that would let you configure these via something like
> 
> ```
> cartographer:
>   metrics_port: <uint>
>   pprof_port: <uint>
> ```
> 
> but then looking at the metrics, it seems like the cardinality is all good (thus, no worries about having it on all the time), however _pprof_ is for sure one of those that you might or might not want to have it on all the time ...
> 
> soo, wdyt of
> 
> * update `./config`'s `Deployment` to have under `args: [ .. ]` a value for `--metrics-port` so that it's always enabled by default (as well as create a `Service` object, probably with `prometheus` annotations ala https://github.com/vmware-tanzu/cartographer-conventions/blob/9634b26dfb5fe7bbdbfa4f0607180c719b589493/config/rbac/auth_proxy_service.yaml#L5-L7
> * update `package-for-cartographer` to include an overlay for enabling pprof (I can take this one)
> 
> ?
> 
> thank you!

@cirocosta Yes, I had, I wasn't sure if it was going to be okay having metrics always on when I first started this PR, but I'll go ahead and do that.

config/manager/deployment.yaml
```diff
@@ -67,3 +68,20 @@ spec:
...
               cpu: 500m
               memory: 512Mi
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/port: "9998"
+    prometheus.io/scheme: https
```
cirocosta: whopsy, it's actually `http` by default 😅  

```suggestion
    prometheus.io/scheme: http
```

cirocosta (approved):
lgtm! for acceptance:

1. created a local release

```
REGISTRY=ghcr.io/cirocosta ./hack/release.sh
```

2. deployed the release

```
kapp deploy -a cartographer -f ./release/cartographer.yaml
```

3. verified that we're able to hit the service

```
k -n cartographer-system  port-forward service/cartographer-controller-metrics-service --address 0.0.0.0 8080:9998

curl <ip>:8080 ..
```

PR #995, opened 2022-08-25
config: prefix generic objects (validatingwebhook and issuer)

### LINKED ISSUE #992: Cluster scoped resources need to be prefixed
### Bug description:

Cluster scoped resources are cluster scoped and share a naming scope with all other components on the cluster. All cluster scoped resources need to have a prefix to avoid collisions.

A ValidatingWebhookConfiguration named `validating-webhook-configuration` has a high risk of colliding with another package.

https://github.com/vmware-tanzu/cartographer/blob/2ee79e68041b06c62d2326673fb2376a0b22e380/config/webhook/manifests.yaml#L20

### Steps to reproduce:
1. Install the RabbitMQ cluster operator 1.3.0
2. Try to install cartographer using kapp

### Expected behavior:
Both packages should be installable and run side-by-side

### Actual behavior:
kapp refuses to install Cartographer because the `validating-webhook-configuration` resource already exists and isn't owned by Cartographer.

### Logs, k8s object dumps:

```
      - Resource 'validatingwebhookconfiguration/validating-webhook-configuration (admissionregistration.k8s.io/v1) cluster' is already associated with a different app 'rabbitmq-operator-ctrl' namespace: tap-install (label 'kapp.k14s.io/app=1661261380955806430')
```

cirocosta:

- config: prefix issuer w/ cartographer-

    despite the object being namespace-scoped, it's such a generic name that
    it makes sense to have it prefixed with `cartographer-` so that we're
    able to carry the component nomenclature.

- config: prefix cluster-scoped validatingwhook

    given that the ValidatingWebhookConfiguration we have (previously named
    `validating-webhook-configuration`) has a high risk of colliding with
    another package, prefix it with `cartographer-` to lower those chances.

closes #992

## Release Note

- prefix cluster-scoped validatingwebhook with `cartographer` to avoid conflicts with other validating webhook configurations
- prefix issues with `cartographer-` to avoid potential collisions in the cartographer-system namespace

## PR Checklist

### DIFF
--- a/hack/overlays/webhook-configuration.yaml
+++ b/hack/overlays/webhook-configuration.yaml
@@ -21,6 +21,7 @@ metadata:
   #@overlay/match missing_ok=True
   annotations:
     cert-manager.io/inject-ca-from: cartographer-system/cartographer-webhook
+  name: cartographer-validating-webhook-configuration
 
 webhooks:
   #@overlay/match by=overlay.all, expects="1+"
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -17,7 +17,7 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   creationTimestamp: null
-  name: cartographer-validating-webhook-configuration
+  name: validating-webhook-configuration
 webhooks:
 - admissionReviewVersions:
   - v1beta1
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -17,7 +17,7 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   creationTimestamp: null
-  name: validating-webhook-configuration
+  name: cartographer-validating-webhook-configuration
 webhooks:
 - admissionReviewVersions:
   - v1beta1
--- a/config/webhook/support.yaml
+++ b/config/webhook/support.yaml
@@ -25,14 +25,14 @@ spec:
     - cartographer-webhook.cartographer-system.svc.cluster.local
   issuerRef:
     kind: Issuer
-    name: selfsigned-issuer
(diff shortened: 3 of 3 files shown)


cirocosta:
image

a-ha, it's generated (makes sense) - will patch

hack/overlays/webhook-configuration.yaml
```diff
@@ -21,6 +21,7 @@ metadata:
   #@overlay/match missing_ok=True
   annotations:
     cert-manager.io/inject-ca-from: cartographer-system/cartographer-webhook
+  name: cartographer-validating-webhook-configuration
```
squeedee, line 24: This will do for now, I guess, but I'm sure there's a better way. A kubebuilder/controller-runtime marker perhaps?

squeedee (approved):
Good enough for now.

idoru (approved):

hack/overlays/webhook-configuration.yaml
cirocosta: didn't find a marker / way of getting around the name (it's [hardcoded](https://github.com/kubernetes-sigs/controller-tools/blob/ebdf5177ad92f89da82686f4100c06d428ab3ce5/pkg/webhook/parser.go#L384-L405)) and, _yes_, using what kubebuilder gives us (kustomize-based config gen) there is a pattern for dealing with it (e.g., see https://github.com/vmware-tanzu/cartographer-conventions/blob/35a8c0b98b2933fcfc0aeb1e97343fd78b8a9cfe/config/default/kustomization.yaml#L4-L9)

scothis: ~how kubebuilder projects handle this problem by default:~ (ciro beat me to posting 😄)

https://github.com/kubernetes-sigs/kubebuilder/blob/9711dc84d0215d2ef06b8cac5046e84ae810eba2/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml#L4-L9

Issue #850, opened 2022-05-13
Add basic events for stamping and observing changes to resources


jwntrs:
### Description of problem

There are times when a user needs to know more detail about the machinations of the Cartographer controller, especially in debugging failures. For these situations, Kubernetes provides [events](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events) to expose the user to these temporal triggers/actions from inside controller processes.

### Proposed solution

As per #711, implement the following events on the owner object:

`StampedObjectApplied`
`StampedObjectRemoved` (StampedObjectGarbageCollected in the RFC)
`ResourceOutputChanged` **
`ResourceHealthyStatusChanged ` **

**: stretch, if they require a lot of time, better to deliver the StampedObject events.

cirocosta:
note from a discussion @ zoom:

- for `StampedObjectApplied`,  it might be interesting to be able to distinguish between update vs create - debugging wise, it can be helpful to figure that out
- given that we're covering both `create` and `update`, might be interesting to include `delete` in here too? (`StampedObjectGarbageCollected`)

PR #1001, opened 2022-08-31
Introduce events

idoru:

Introduces event recording. I've started with the purportedly most desired event, `StampedObjectApplied` first, to gather feedback on the implementation before working on other events requested in #850.

The StampedObjectApplied event is emitted when creating/patching objects.
The Related object for the event is the stamped object itself.
Message indicates whether objects are being created or patched.

Updates #850

A trivial test can be had by running end-to-end tests while watching for events with:
```
kubectl get ev --watch
```

## Release Note

Cartographer now emits the StampedObjectApplied event when creating or patching resources.

## PR Checklist


tests/integration/runnable/runnable_suite_test.go
```diff
@@ -17,6 +17,7 @@ package runnable_test
 import (
 	"context"
 	"io"
+	eventsv1 "k8s.io/api/events/v1"
```
cirocosta: nit: move to non-stdlib section

idoru: :+1

cirocosta:
trying it out here, I wanted to check how error reporting looks like from the perspective of the controller, and unfortunately the library itself logs error messages from underneath us, breaking the nice stream of json-encoded logs :(

```
{
  "level": "info",
  "ts": 1662063165.8117745,
  "msg": "patching object",
  "controller": "workload",
  ...
  "object": {
    "apiVersion": "v1",
    "kind": "ConfigMap",
    "namespace": "ootb-supply-chain-basic-docker-context",
    "name": "app"
  }
}

E0901 20:12:45.817179       1 event.go:267] Server rejected event
'&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""},
ObjectMeta:v1.ObjectMeta{Name:"app.1710d5a154b6eff9", GenerateName:"",
Namespace:"ootb-supply-chain-basic-docker-context", SelfLink:"", UID:"", ResourceVersion:"",
Generation:0, CreationTimestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC),
DeletionTimestamp:<nil>, DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil),
...
in API group "" in the namespace "ootb-supply-chain-basic-docker-context"' (will not retry!)


{
  "level": "info",
  "ts": 1662063165.8399498,
  "logger": "workload-stamping-repo-cache",
  "msg": "checking for changes since cached",
  "key": "ootb-supply-chain-basic-docker-context:ConfigMap:app-with-claims:"
}
```

it makes sense as the recorder is very much "fire and forget", with the broadcasting happening in the background (there's more to it than just an API call  (see [recordEvent](https://github.com/kubernetes/client-go/blob/7ccf7b05af286664a658af15c8502a6066ae3288/tools/record/event.go#L234-L238)), but, something to keep in mind.

_ps.: this is similar to those warnings `client-go`-based libs see when throttling_

metrics-wise, it's also quite opaque, e.g., we can see that the client did face some trouble, but you can't reeeally tell that's related to event emission at all

```
# HELP rest_client_requests_total Number of HTTP requests, partitioned by status code, method, and host.
# TYPE rest_client_requests_total counter
rest_client_requests_total{code="200",host="10.96.0.1:443",method="GET"} 10442
rest_client_requests_total{code="200",host="10.96.0.1:443",method="PATCH"} 100
rest_client_requests_total{code="200",host="10.96.0.1:443",method="PUT"} 77
rest_client_requests_total{code="201",host="10.96.0.1:443",method="POST"} 160
rest_client_requests_total{code="403",host="10.96.0.1:443",method="POST"} 9
rest_client_requests_total{code="404",host="10.96.0.1:443",method="GET"} 51
rest_client_requests_total{code="409",host="10.96.0.1:443",method="PUT"} 18
```

(something else to keep in mind, I don't think there's much we could do in this PR around this as, as you mentioned, this one is an introduction to better supporting events)

idoru:
Yeah, it's kinda yuck - but doesn't seem like something we can change without controller-runtime changes or completely re-implementing event broadcasting ourselves. I got the impression that controller-runtime is (at least for now) passing the buck here when I saw many projects using it were just using `Manager.GetEventRecorderFor()`.

Metrics-wise, if we decide we care about breaking down API calls numbers by event-propagation vs. business-as-usual, would it be safe to assume the primary concern would be that a disproportionate number of event calls are being made (to B-A-U calls)? If so, we could measure fairly easily by decorating event recorders to collect metrics. This may not show us the method breakdown (PATCH/PUT/etc) but the actual quantity of emitted events *ought* to be enough to make an informed decision on whether event creation has gotten out of control.

cirocosta (approved):
lgtm!

squeedee:
The events are on the Stamped Object, not the Owner Object - we're working on a fix.

idoru:
Given what @squeedee discovered, and what has been changed, I think it may be advisable when manually verifying:
- watch the `Workload` to see that `Events` properly materialize upon it.
- check that the `<resource>.<group>/<name>` emitted in the event `message` field relates to the correct applied stamped object.

cirocosta (approved):
lgtm!

one thing to note for folks looking at this PR is that it _doesn't_ cover gc dynamics, so, e.g., if you update a template such that the children object has a new name (i.e., we expect to end up with the previous object deleted and a new one created), one would only see the _creation_ event and not the deletion (as the implementation here is solely for the reason `StampedObjectApplied`):

```
Events:
  Type    Reason                Age    From      Message
  ----    ------                ----   ----      -------
  Normal  StampedObjectApplied  3m42s  Workload  Created object [tests.test.run/test-git-template]
           - initial creation, cool!
  
  Normal  StampedObjectApplied  2m13s  Workload  Patched object [tests.test.run/test-git-template]
           - I induced a little change to the spec, and we see the patch! cool!
  
  Normal  StampedObjectApplied  11s    Workload  Created object [tests.test.run/fooo-test-git-template]
           - this one came from changing `metadata.name` of the child in the
           - template, thus, we see the creation of the new one here but lack a 
           - delete (something to be done in a later PR imo)
```

PR #1009, opened 2022-09-09
Use inclusive terms

emmjohnson:


waciumawanjohi:
The changes here look appropriate. Should we add an automatic linter to our process
1. To ensure that we've caught all current issues.
2. To ensure that we prevent later reversions.
?

waciumawanjohi (approved):

PR #1034, opened 2022-09-22
Fix ytt failing test with version 0.42 of ytt

squeedee:
Fix ytt failing test with version 0.42 of ytt
- we currently ship with 0.42
- we need to work out what's actually wrong with 0.43
- includes version display step for debug in the future

### DIFF
--- a/.github/workflows/validation.yaml
+++ b/.github/workflows/validation.yaml
@@ -40,6 +40,11 @@ jobs:
 
       - name: setup carvel tooling binaries
         uses: vmware-tanzu/carvel-setup-action@v1
+        with:
+          ytt: v0.42.0 # Todo unlock as soon as 0.43 is working.
+
+      - run: |
+             ytt version
 
       - name: run source checks
         uses: ./.github/actions/source-checks


.github/workflows/validation.yaml
```diff
@@ -40,6 +40,12 @@ jobs:
 
       - name: setup carvel tooling binaries
         uses: vmware-tanzu/carvel-setup-action@v1
+        with:
+          only: ytt
+          ytt: v0.42.0 # Todo unlock as soon as 0.43 is working.
+
+      - run: |
```
emmjohnson, line 46: do you want to leave the version command in?

emmjohnson (approved):

.github/workflows/validation.yaml
squeedee: I'll take it out as soon as 0.43 is working.

squeedee: (in a future pr)

squeedee:
Ok well this is not passing all the time either. Looks like it's not the version.

PR #1035, opened 2022-09-22
Fix flaky CI for template/ytt tests

squeedee:
Fix flaky CI for template/ytt tests
- also reduce the chances of the same problem occurring in the wild.


idoru (approved):
LGTM :+1:

PR #1038, opened 2022-09-26
Switch to VMWare provided Inclusive Language sonarqube instance

squeedee:
We also run our internal scanner to ensure we miss nothing.
Currently we're not gating on it because certain exclusions need to be fixed.


idoru (approved):
LFTM

PR #1042, opened 2022-09-28
Add woke

squeedee:
Add Woke 

- remove inclusive script
- remove sonarqube its scan


cirocosta:
@squeedee have you checked https://github.com/carto-run/fake-jenkins/pull/19? seems like we could make use of the action and have the rules etc pulled as such 🤔  wdyt?

squeedee:
@cirocosta i think i still want woke to run when i do a `make lint` - so then I'd have to contextually omit woke from `make lint` on ci

Makefile
```diff
@@ -136,8 +136,12 @@ coverage:
 	go tool cover -html=coverage.out -o coverage.html
 	open coverage.html
 
+.PHONY: woke
+woke:
+	woke -c https://via.vmw.com/its-woke-rules
```
cirocosta: wdyt of using that same pattern as we're going for the other tools?

- add to [hack/tools/go.mod](https://github.com/vmware-tanzu/cartographer/blob/375fd52a60184125ada30a77d40a137199ae9a90/hack/tools/go.mod)
- `go run` it [ala golangci-lint](https://github.com/vmware-tanzu/cartographer/blob/375fd52a60184125ada30a77d40a137199ae9a90/Makefile#L1)

this way `dependabot` would also be bumping it and we all run the same version etc

squeedee: Oh yeah i forgot it was written in go. Original project was javascript

emmjohnson (approved):

PR #1044, opened 2022-09-30
Emit ResourceHealthyStatusChanged events

idoru:
Transitioning to Status Unknown will not emit this event if this Condition type was not previously seen to exist on the resource status

[#850]

Emit events with the ResourceHealthyStatusChanged reason when  the ResourceHealthy status of a resource is observed to change between Unknown, True and False

## Release Note

When the ResourceHealthy status of a resource is observed to change between Unknown, True and False, Cartographer now emits an event with the ResourceHealthyStatusChanged reason.

## PR Checklist

~- [ ] Modified the docs to match changes ~


pkg/realizer/realizer.go
```diff
@@ -149,6 +150,9 @@ func (r *realizer) Realize(ctx context.Context, resourceRealizer ResourceRealize
 			}
 		}
 		resourceStatuses.Add(realizedResource, err, additionalConditions...)
+		if slices.Contains(resourceStatuses.ChangedConditionTypes(realizedResource.Name), v1alpha1.ResourceHealthy) {
+			events.FromContextOrDie(ctx).ResourceEventf(events.NormalType, events.ResourceHealthyStatusChangedReason, "[%s] found a new healthy status in [%Q]", stampedObject, realizedResource.Name)
```
cirocosta: per the rfc, it does indeed implement what we've agreed back then

```
Events:
  Type    Reason                        Age                    From      Message
  ----    ------                        ----                   ----      -------
  Normal  StampedObjectApplied          12m                    Workload  Created object [configmaps/workload-provider]
  Normal  ResourceOutputChanged         12m                    Workload  [config-provider] found a new output in [configmaps/workload-provider]
  Normal  ResourceHealthyStatusChanged  12m                    Workload  [config-provider] found a new healthy status in [configmaps/workload-provider]
  Normal  StampedObjectApplied          12m                    Workload  Created object [configmaps/workload-consumer]
  Normal  ResourceHealthyStatusChanged  12m                    Workload  [config-consumer] found a new healthy status in [configmaps/workload-consumer]
...
```

but I'm wondering if we should instead go one step further and be closer to what `pod` does for things like `readinessprobe`, where e.g., having a pod that's not properly passing its readinessprobe, we end up with the following sequence of events:

```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  2m53s                 default-scheduler  Successfully assigned default/foo to cartographer-control-plane
  Normal   Pulling    2m52s                 kubelet            Pulling image "registry.k8s.io/busybox"
  Normal   Pulled     2m52s                 kubelet            Successfully pulled image "registry.k8s.io/busybox" in 527.71046ms
  Normal   Created    2m52s                 kubelet            Created container foo
  Normal   Started    2m52s                 kubelet            Started container foo
  Warning  Unhealthy  43s (x21 over 2m18s)  kubelet            Readiness probe failed: cat: can't open '/tmp/healthy': No such file or directory
```

had that event been rewritten in a way similar to what we implemented here, it'd be something in the lines of 

```
  Normal  ResourceHealthyStatusChanged  12m                    Pod  Readiness probe found a new readiness state
```

which wouldn't be very useful 🤔  wdyt of going for trying to pinpoint:

- ~the resource that the healthiness check observed the change for~ oh this does show!
- the healthiness state that's been observed (i.e., indicate whether we transitioned to `ready: true|false|unknown`)

wdyt?

cirocosta: yaay after 8803ba5, trying a simple transition `healthy -> unknown -> healthy`, we can clearly see the events reflecting such state transitions:

```
  Normal  ResourceHealthyStatusChanged  2m41s               Workload  [config-consumer] found healthy status in [configmaps/workload-consumer] changed to [True]
  Normal  ResourceHealthyStatusChanged  33s (x2 over 33s)   Workload  [config-provider] found healthy status in [configmaps/workload-provider] changed to [Unknown]
  Normal  ResourceHealthyStatusChanged  4s (x2 over 2m41s)  Workload  [config-provider] found healthy status in [configmaps/workload-provider] changed to [True]
```

lgtm!

cirocosta (approved):

PR #1059, opened 2022-10-07
Refactor Params handling

squeedee:
A small effort to absorb some Go practices:

1. Accept interfaces
2. Return structs
3. Describe the interface where you ingest it

This is part of a larger effort to disentangle realizer and component. We picked params as a simple target to start with.

There's still more interfaces that Params could imbibe, for each of the Owner, Blueprint and Resource fields


pkg/realizer/params_test.go
```diff
@@ -56,21 +68,17 @@ var _ = Describe("Params", func() {
 	}
 
 	DescribeTable("ParamsBuilder",
```
idoru:
```suggestion
	DescribeTable("ParamGenerator",
```

idoru (approved):
LGTM I left a suggestion to update the Gingko prose to match your changes.

PR #1060, opened 2022-10-10
Refactor context

emmjohnson:

closes # 

## Release Note

## PR Checklist


pkg/realizer/component.go
```diff
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/vmware-tanzu/cartographer/pkg/stamp"
```
cirocosta: nit: move to the local imports section

pkg/templates/reader.go
```diff
@@ -17,39 +17,33 @@ package templates
...
 
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/vmware-tanzu/cartographer/pkg/apis/v1alpha1"
-	"github.com/vmware-tanzu/cartographer/pkg/eval"
 )
 
-const PREVIEW_CHARACTER_LIMIT = 1024
+type Reader interface {
```
cirocosta, line 25: It's quite a "go"-ism, but, usually single-method interfaces are named after the method + `er`, e.g., [`io.Reader`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/io/io.go;l=83) provides the `Read` method

```go
type Reader interface {
	Read(p [][byte](https://pkg.go.dev/builtin#byte)) (n [int](https://pkg.go.dev/builtin#int), err [error](https://pkg.go.dev/builtin#error))
}
```

and _usually_ rather than getters being named "GetFoo", they're _usually_ plain "`Foo()`" which in this case could look like

```go
// uf, `outputter` looks like a rough name? not sure exactly 
//
type Outputter interface {
        Output(stampedObject *unstructured.Unstructured) (*templates.Output, error)
}
```

wdyt?

cirocosta: ps.: i'm _personally_ not that strongly against explicit `Get<Foo>()` methods in interfaces

https://go.dev/doc/effective_go#Getters

cirocosta (approved):
overall lgtm! I've left a comment about the naming choice around the Reader interface, but that's really it - good refactor!

pkg/testing/templates.go
```diff
@@ -313,15 +320,15 @@ func (i *TemplateTestGivens) getActualObject() (*unstructured.Unstructured, erro
...
 
 	blueprintParams, err := i.getBlueprintParams()
 	if err != nil {
 		return nil, fmt.Errorf("get blueprint params failed: %w", err)
 	}
 
-	paramGenerator := realizer.NewParamGenerator([]v1alpha1.BlueprintParam{}, blueprintParams, workload.Spec.Params)
-	params := paramGenerator.GetParams(template)
+	paramGenerator := realizer.NewParamMerger([]v1alpha1.BlueprintParam{}, blueprintParams, workload.Spec.Params)
+	params := paramGenerator.Merge(template)
```
waciumawanjohi:
```suggestion
paramMerger := realizer.NewParamMerger([]v1alpha1.BlueprintParam{}, blueprintParams, workload.Spec.Params)
	params := paramMerger.Merge(template)
```

pkg/templates/reader_test.go
```diff
@@ -27,7 +27,7 @@ var _ = Describe("Templates", func() {
 	var (
 		err           error
 		apiTemplate   client.Object
-		templateModel templates.Template
+		templateModel templates.Reader
```
waciumawanjohi: Should we rename the variable to templateReader?

pkg/templates/reader.go
```diff
@@ -17,39 +17,33 @@ package templates
...
+	GetDefaultParams() v1alpha1.TemplateParams
 
-type Template interface {
+	// GetResourceTemplate returns the actual representation of a resource to stamp, and how to handle it
+	// TODO: we should be expecting something with a [ytt|template] interface, the health rules and params should
```
waciumawanjohi, line 29: The changes in pkg/templates/cluster_template.go seem to cut the opposite direction. Before we returned a struct with only the ytt and template field. Now a struct with those and the healthRules and params is returned.

pkg/realizer/context.go
```diff
@@ -0,0 +1,75 @@
...
+	inputGenerator := NewInputGenerator(resource, outputs)
+	merger := NewParamMerger(resource.Params, c.blueprintParams, c.ownerParams)
+
+	configs := inputGenerator.GetConfigs()
+	sources := inputGenerator.GetSources()
+	images := inputGenerator.GetImages()
+	result := map[string]interface{}{
+		"workload":    c.owner,
+		"deliverable": c.owner,
+		"params":      merger.Merge(templateParams),
```
waciumawanjohi, line 49: Should we continue to create the merger object? It is not long lasting, we create it and then call a method on it. The parameter for that method exists when we create the merger object. Should this just be a function?

pkg/realizer/realizer_test.go
```diff
@@ -145,12 +154,12 @@ var _ = Describe("Realize", func() {
...
 			outputFromFirstResource := &templates.Output{Image: "whatever"}
 
-			resourceRealizer.DoCalls(func(ctx context.Context, resource realizer.OwnerResource, blueprintName string, outputs realizer.Outputs, mapper meta.RESTMapper) (templates.Template, *unstructured.Unstructured, *templates.Output, error) {
+			resourceRealizer.DoCalls(func(ctx context.Context, resource realizer.OwnerResource, blueprintName string, outputs realizer.Outputs, mapper meta.RESTMapper) (templates.Reader, *unstructured.Unstructured, *templates.Output, error) {
 				executedResourceOrder = append(executedResourceOrder, resource.Name)
 				Expect(blueprintName).To(Equal("greatest-supply-chain"))
 				if resource.Name == "resource1" {
 					Expect(outputs).To(Equal(realizer.NewOutputs()))
-					template, err := templates.NewModelFromAPI(template1)
+					template, err := templates.NewReaderFromAPI(template1)
```
waciumawanjohi: Matching the variable names to the function output can clarify both what's happening and what's the difference between what is now `template1`  and `template`.

```suggestion
					templateReader, err := templates.NewReaderFromAPI(template1)
```

pkg/templates/reader.go
emmjohnson: im fine with outputter

pkg/templates/reader.go
emmjohnson: I'm not sure what you mean, we have removed a bunch of methods from this interface

pkg/realizer/context.go
squeedee: We're definitely descending into a refactor rabbit-hole at the moment. This is quite possibly a responsibility (construction) that we can lift outside of the context builder or the context builder is the imperfect abstraction. Would just like to leave this open until we've considered the interfaces as a whole. 

If we never end up passing this by interface then i think a local public function (public for the `testing` package) is absolutely fine.

pkg/templates/reader.go
waciumawanjohi: GetResourceTemplate used to be:
```go
func (t *clusterTemplate) GetResourceTemplate() v1alpha1.TemplateSpec {
	return v1alpha1.TemplateSpec{
		Template: t.template.Spec.Template,
		Ytt:      t.template.Spec.Ytt,
	}
}
```

and is now

```go
func (t *clusterTemplate) GetResourceTemplate() v1alpha1.TemplateSpec {
	return t.template.Spec
}
```

e.g. GetResourceTemplate is now returning a struct with healthRules and params included, which seems to run counter to your goals.

emmjohnson: No, we actually were making it consistent with the other templates and following the interface.

PR #1079, opened 2022-10-20
Runnable behavior in all templates

waciumawanjohi:

Implements [the interface](https://github.com/vmware-tanzu/cartographer/blob/rfc-simplify-runnable-usage/rfc/rfc-0000-simplify-runnable-use.md#proposed-example) from #705 
While the interface remains the same, the implementation differs from the RFC as no runnable or clusterRunTemplate objects are created.
Runnable is hardcoded to tekton's success criteria. This implementation leverages uses the healthRule on the template to determine success. This allows immutable templates to stamp out more than Tekton taskruns.

Future work that is necessary to fully deprecate Runnable: [allow templates to read objects that weren't created by cartographer](https://docs.google.com/document/d/1-FiuD6iLl9tT9nsLY4FDb_cMKMvA6SkjFo-oNzXI9Js/edit#heading=h.xeoad5e7oji8)

## Release Note

Lifecycle field added to all templates. Enables 'runnable' behavior for any template.

## PR Checklist


config/crd/bases/carto.run_clustertemplates.yaml
```diff
@@ -199,6 +199,17 @@ spec:
                       unhealthy. Otherwise, healthiness is Unknown.
                     type: string
                 type: object
+              lifecycle:
+                default: default
+                description: Lifecycle specifies whether template modifications should
```
squeedee, line 204: Nit: I know it's still dependent on @zrob, but I would like this to be clearly marked `Experimental`

pkg/apis/v1alpha1/cluster_template.go
```diff
@@ -68,6 +68,20 @@ type TemplateSpec struct {
 	// stamped by this template
 	// +optional
 	HealthRule *HealthRule `json:"healthRule,omitempty"`
+
+	// Lifecycle specifies whether template modifications should result in originally
+	// created objects being updated (`default`) or in new objects created alongside
+	// original objects (`immutable` or `tekton`).
+	// +kubebuilder:validation:Enum=default;immutable;tekton
+	// +kubebuilder:default="default"
```
squeedee: "Default" lifecycle is only meaningful if you know what things used to look like. I think "mutable" is a better, specific term.

squeedee (approved):

pkg/apis/v1alpha1/cluster_template.go
waciumawanjohi: I'd say this is also a design question for @zrob

zrob: agree default doesn't help much. let's go with "mutable" as the default behavior.

also propose we remove immutable. it is confusing to have three options and only two behaviors. we're being explicit with tekton right now. we can add immutable and give it meaning if we have a specific use case to deliver.

lifecycle: mutable (default), tekton

waciumawanjohi: lifecycle `default` has been changed to `mutable`.

waciumawanjohi: `Immutable` and `Tekton` have different behavior when no healthRule is specified by the user.

The new behavior stamping immutable resources relies on the `healthRules` field of the template. This is not the case with Runnable, which is hardcoded to consider a resource with a true `Succeeded` condition to be healthy and everything else to be unhealthy. That ties Runnable to Tekton. 

For the new behavior, when no healthRule is specified by the user:
- `tekton`: defaults to the healthRule of the succeeded condition == true means healthy (similar to Runnable)
- `immutable`: defaults to the healthRule of AlwaysHealthy (similar to mutable templates)

For both a template with lifecycle `tekton` and `immutable` users can specify healthRules. When healthRules are specified, there is no difference between these two lifecycles.

_After a conversation with Zach, we'll keep the current behavior and clarify it in documentation._

PR #1128, opened 2022-12-15
Add labels to template context for workload & deliverable

idoru:

This change makes the [labels that Cartographer will apply](https://cartographer.sh/docs/v0.6.0/stamping/#labels) to a stamped object available in the templating context for Workloads and Deliverables. This is useful if you want make templates which render Cartographer objects which you wish to make inert for later use (by placing their content into a ConfigMap or Secret, for example) and still want to preserve Cartographer's labelling.

## Release Note
`labels` for the currently templated resource are now available in the templating context for Workloads and Deliverables.

## PR Checklist

### DIFF
--- a/pkg/realizer/component.go
+++ b/pkg/realizer/component.go
@@ -39,7 +39,7 @@ import (
 //go:generate go run -modfile ../../hack/tools/go.mod github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 type ContextGenerator interface {
-	Generate(templateParams TemplateParams, resource OwnerResource, outputs OutputsGetter) map[string]interface{}
+	Generate(templateParams TemplateParams, resource OwnerResource, outputs OutputsGetter, labels templates.Labels) map[string]interface{}
 }
 
 type resourceRealizer struct {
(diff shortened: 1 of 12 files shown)


waciumawanjohi:
Looks good!

Some questions I had when I was looking at the test:

- What would happen if someone tried to access the whole labels object at once?
- Does this syntax align with user expectations of how they access other items in the templating context? I had a vague expectation that the value would be accessed at `data.values.labels.label-name`. (Of course there's the issue that our label names have a `/` character in them. So maybe the `data.values.labels['label-name']`) syntax is a benefit.
- Is the syntax the same in the non-ytt templates?

Not blocking concerns; food for thought.

idoru:
You could access the whole labels object to simply copy them all forward into the template as a map. That is the use case I had in mind with this, but the solution does not preclude also doing template logic with discrete labels if the template author desires.

I think in this situation it (I hope?) shouldn't be surprising that labels are a map, and both templating realms have support for quoting "troublesome keys".

waciumawanjohi (approved):

PR #1280, opened 2023-06-21
Remove libgit2 docs

tbr11:

Remove references to libgit2 in documentation, as the most likely usecase is fluxcd source-controller which no longer supports libgit2.


waciumawanjohi:
If we're bumping the docs and examples to match FluxCD +v0.33, we should do that in our tests too: https://github.com/vmware-tanzu/cartographer/blob/9d76098051c7e66027b1ce5e4fdfb7a81a03c719/hack/setup.sh#L41

PR #1267, opened 2023-06-14
Pin go version for GitHub actions

waciumawanjohi:
[PR requirements]: https://github.com/vmware-tanzu/cartographer/blob/main/CONTRIBUTING.md#commit-message-and-pr-guidelines

## Changes proposed by this PR

Github actions are currently broken by go 1.19.10. We delete files and then run make generate to recreate them. Track the bug in counterfeiter here: https://github.com/maxbrunsfeld/counterfeiter/issues/246

## Release Note

[imperative mood]: https://chris.beams.io/posts/git-commit/#imperative

## Cherry-pick branches
_All_

## PR Checklist

Note: Please do not remove items. Mark items as done `[x]` or use ~strikethrough~ if you believe they are not relevant

- [x] Linked to a relevant issue. Eg: `Fixes #123` or `Updates #123`
- [x] Removed non-atomic or `wip` commits
- ~[ ] Filled in the [Release Note](#Release-Note) section above~
- [x] Added any relevant branches to cherry-pick
- ~[ ] Modified the docs to match changes~

PR #1285, opened 2023-07-10
Bump go version to 1.20

waciumawanjohi:

Stop pinning go version in github action. See https://github.com/vmware-tanzu/cartographer/pull/1267

## Release Note

## Cherry-pick branches
0.6, 0.5

## PR Checklist

- [x] Added any relevant branches to cherry-pick

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
       - name: set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '^1.19.9'
+          go-version-file: 'go.mod'
 
       - name: setup carvel tooling binaries
         uses: vmware-tanzu/carvel-setup-action@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: '^1.19.9'
+          go-version-file: 'go.mod'
```
squeedee, line 33: This has caused issues in the past with surprise non-repro. That is, we cant seem to build/test something that was working and can't figure it out. That's why we used a pinned minor. Would rather we use `^1.20.6`

squeedee (approved):

.github/workflows/release.yaml
squeedee: I had misread the change. After discussion, agree this is a great solution for unpinning but tracing a minor version.

waciumawanjohi: To document: we chatted, this change uses the minor written into `go.mod`. While we could achieve patch pinning and auto-updating with UpdateCLI, we're satisfied with the minor pinning that we achieve with `go-version-file`.

waciumawanjohi: Jinx

